PR #21, opened 2021-08-17
4 initial frontend commit

davidsmejia:
Adds the initial commit there isn't much to review

* used `yarn create next-app`
*  moved the default gitignore to top level of the repo
* cleaned up boilerplate code

PR #22, opened 2021-08-17
8 setup storybook

davidsmejia:
This is branched off of #21 because the structure is required. I think reviewing this will be easier after that is merged.

* Adds storybook
* cleans out boilerplate examples


kurtwheeler (approved):
You're missing newlines but other than that this LGTM

Issue #14, opened 2021-08-17
Add presigned `download_url` to computed files


kurtwheeler:
Use a AWS presigned URL to offer computed files to anyone with an API token.

API tokens will also need to be added to the API.

kurtwheeler:
closed by #35

PR #35, opened 2021-08-27
Add APIToken and check for it in ComputedFile

kurtwheeler:
## Issue Number

#14 

This implementation is pretty similar to refine.bio's. The only notable difference is that the endpoint is named `tokens` instead of `token` for consistency with all other plural endpoints.

## Types of changes

## Functional tests

I've already tested that I can't just download the S3 object without these.

I also tested the curl commands I added to the README and made sure I was able to download the file using the `download_url` from the response.

## Checklist


davidsmejia (approved):
LGTM

PR #47, opened 2021-09-02
Remove trailing _sce postfix from file names.

kurtwheeler:
## Issue Number

N/A Things broke when the input format changed

These used to have this in them, now they don't.

I also was getting `null value in column "s3_bucket" violates not-null constraint` errors when trying to purge a project because we were setting `s3_bucket` and `s3_key` to null when we deleted them from s3. I'd rather have the not null constraint than be able to set them to None right before we delete them though.

## Types of changes

## Functional tests

I've run

```
sportal manage-api load_data --reload-existing --upload True
```

To update the local data bucket.

## Checklist


davidsmejia (approved):
LGTM

PR #56, opened 2021-09-10
Make sure the exit code of sportal matches the command it calls.

kurtwheeler:
## Issue Number

#55 

#55 should have failed, but it didn't because the exit code of `sportal` wasn't correct. This should prevent that from happening in the future.

## Types of changes

## Functional tests

I've made sure `sportal test-api` with a failing test has an exit code of 1.


davidsmejia (approved):
nice catch

Issue #64, opened 2021-09-20
Misc API Tasks


davidsmejia:
1. `project.diagnoses_counts` has an extra brace around the diagnosis ex:
```"diagnoses_counts": "{'Neuroblastoma'} (1)",```
1. project.additional_metadata isnt a thing so could we add a `project.additional_metadata_keys` or something like that so i can present them in the search results?

kurtwheeler:
Closed by #66

Issue #50, opened 2021-09-03
Switch to using SCPCP0001 id instead of pi_name


kurtwheeler:
### Context

We're currently using `pi_name` as a unique id, but a single PI could have more than one project.

### Problem or idea

The science team is going to create a project accession code for us

### Solution or next step

Switch the pk for projects to be the new accession code. (That will be in the URL as well.)

kurtwheeler:
Closed by #66

PR #66, opened 2021-09-20
Use accession codes and other tweaks

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/50
https://github.com/AlexsLemonade/scpca-portal/issues/64

This changes `Sample.scpca_sample_id` to `Sample.scpca_id` to reduce redundancy, along with being consistent with with the new field added to Project.

These fields are used as the pk field for their models and the API exposes them. We still have an auto incrementing id field just to track differences between records when deleted and recreated. Also Django doesn't like non-integer PKs: https://stackoverflow.com/a/3944247/6095378

This also adds `human_readable_pi_name` and `additional_metadata_keys` to Project.

This includes each Samples' ComputedFile on the samples list endpoint.

It also fixes `Project.diagnoses_counts` and adds `Project.additional_metadata_keys`.

## Types of changes

## Functional tests

I've run `sportal load-data` locally and with `sportal load-data --reload-existing --upload True` to update the local bucket.

I've updated unit tests as well.

## Checklist


api/scpca_portal/views/project.py
```diff
@@ -13,10 +13,12 @@ class ProjectDetailSerializer(ProjectSerializer):
...
     queryset = Project.objects.all().order_by("-created_at")
     ordering_fields = "__all__"
-    lookup_field = "pi_name"
+    lookup_field = "scpca_id"
+
     filterset_fields = {
-        "id": ["exact"],
+        "scpca_id": ["exact"],
         "pi_name": ["exact"],
+        "human_readable_pi_name": ["contains"],
```
davidsmejia, line 21: i think icontains would be better here

davidsmejia (approved):
LGTM one comment tho

PR #67, opened 2021-09-20
39 setup generic components

davidsmejia:
## Issue Number

#39 
#10 

* refactors the storybook to use the updated story format
  * lets us do more and fixes some linting issues
* adds generic components ...and a couple not very generic components
* adds a bunch of images for icons and badges
* adds helpers and hooks to support the components

## Types of changes

## Functional tests

The stories are the tests

## Checklist

- [X] I have added necessary documentation (if appropriate)

## Screenshots
Screen Shot 2021-09-20 at 6 24 03 PM


davidsmejia:
I still need to figure out how to publish the storybook so we can have @dvenprasad review them

client/.storybook/data/project.js
```diff
@@ -0,0 +1,66 @@
...
+    {
+      diagnosis: 'Neuroblastoma',
+      seq_unit: 'cell',
+      technology: '10Xv3',
+      sample_count: 1,
+      updated_at: '2021-09-10T20:02:07.360550Z'
+    }
+  ],
+  id: 3,
+  pi_name: 'dyer_chen',
```
kurtwheeler, line 49: There's `human_readable_pi_name` and `additional_metadata_keys` now too. This looks like it basically just came from `https://api.staging.scpca.alexslemonade.org/v1/projects/SCPCP00003/`. Maybe just refresh it now that #66 is out and the data is updated?

client/.storybook/data/samples.js
```diff
@@ -0,0 +1,26 @@
...
+    additional_metadata: {
+      'COG Risk': 'High',
+      Treatment: 'treated',
+      'IGSS Stage': '4',
+      'Primary Site': 'Adrenal'
+    },
+    created_at: '2021-09-10T20:02:07.315349Z',
+    updated_at: '2021-09-10T20:02:07.315377Z'
+  }
+]
```
kurtwheeler, line 26: Same comment about refreshing this as above and also for `projects.js`

client/.storybook/stories/DetailsTable.stories.js
```diff
@@ -0,0 +1,19 @@
...
+  args: { order, data: project }
+}
+
+export const Default = (args) => (
+  <DetailsTable data={args.data} order={args.order} />
+)
+
+export const UsingOrder = (args) => (
+  <DetailsTable data={args.data} order={args.order} />
+)
```
kurtwheeler, line 19: Is this not identical to `Default`?

client/.storybook/stories/Table.stories.js
```diff
@@ -0,0 +1,21 @@
...
+
+const columns = [
+  { Header: 'Technologies', accessor: 'technologies' },
+  { Header: 'Diagnosis', accessor: 'diagnosis' },
+  { Header: 'Disease Timing', accessor: 'disease_timing' }
+]
+
+export default {
+  title: 'Components/Table',
+  args: { columns, data: samples }
```
kurtwheeler, line 14: Rather than repeating the same sample 10x in `samples.js`, couldn't you just do something like:

```suggestion
  args: { columns, data: Array(10).fill(sample) }
```

Although actually it'd be nice to do >10 times to see a second page on the table. I think there might be an off by one error in it somewhere because there's two pages shown for only the ten items and they don't change back and forth:

pagination

client/src/contexts/ScPCAPortalContext.js
```diff
@@ -0,0 +1,69 @@
...
+
+  const getToken = () => ({
+    email,
+    is_activated: acceptsTerms
+  })
+
+  // const signUpForEmails = async () => {
+  //  // probably shouldnt use the api for this
+  //  // maybe helpers?
+  // }
```
kurtwheeler, line 27: Are this and the other comment out block above TODOs or dead code?

kurtwheeler (changes requested):
I only had a few comments but I think the table is showing an extra page. Other than that this looks great!

client/.storybook/stories/Table.stories.js
davidsmejia: Im going to just unique the samples data more.. i wanted something that would let me show off the filtering anyway

davidsmejia: Not an off by one error but I did change the getLastOffset helper I added to more accurately handle that

kurtwheeler (approved):
LGTM!

PR #68, opened 2021-09-21
Bad tokens cause a 403 rather than just not providing a download URL.

kurtwheeler:
## Issue Number

N/A @davidsmejia thought something broke but it was just a bad token

Silent failures bad! If there's a token and it's not valid, return a 403.

## Types of changes

## Functional tests

I've added a unit test.

## Checklist

### DIFF
--- a/api/scpca_portal/views/computed_file.py
+++ b/api/scpca_portal/views/computed_file.py
@@ -1,5 +1,6 @@
 from django.core.exceptions import ValidationError
 from rest_framework import serializers, viewsets
+from rest_framework.exceptions import PermissionDenied
 
 from rest_framework_extensions.mixins import NestedViewSetMixin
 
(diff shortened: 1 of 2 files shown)


api/scpca_portal/views/computed_file.py
```diff
@@ -64,8 +65,13 @@ def get_serializer_context(self):
...
-            token = APIToken.objects.get(id=token_id, is_activated=True)
-            return {**serializer_context, "token": token}
-        except (APIToken.DoesNotExist, ValidationError):
+        if token_id:
+            try:
+                token = APIToken.objects.get(id=token_id, is_activated=True)
+                return {**serializer_context, "token": token}
+            except (APIToken.DoesNotExist, ValidationError):
+                raise PermissionDenied(
+                    {"message": "Your token is not valid or activated.", "token_id": token_id}
```
davidsmejia, line 74:
```suggestion
                    {"message": "Your token is not valid or not activated.", "token_id": token_id}
```

PR #73, opened 2021-09-27
add donate to header

davidsmejia:
## Issue Number

N/A

* Adds a Donate button for the header
* Adds an option to show it on the header
* Adds the option when user is on about page to show the donate button on the header

## Types of changes

## Functional tests

n/a can test when we make the about page

## Checklist

## Screenshots

n/a


kurtwheeler (approved):
LGTM! It seems like maybe this could be reused for all our sites... I wonder if a ccdl-react-components library would be worth making at some point :thinking:

PR #74, opened 2021-09-27
58 projects browse page

davidsmejia:
## Issue Number

#58 

* Adds the browse projects page and the filtering to support that page

## Types of changes

## Functional tests

n/a works but would work better with more data

## Checklist

## Screenshots
Screen Shot 2021-09-27 at 3 46 27 PM

### DIFF
--- a/client/src/components/ProjectSearchFilterPills.js
+++ b/client/src/components/ProjectSearchFilterPills.js
@@ -52,7 +52,7 @@ export const ProjectSearchFilterPills = ({
 
   const removeFilter = (filter, option) => {
     const newFilters = { ...filters }
-    newFilters[filter].splice(newFilters[filter].indexOf(option), 2)
+    newFilters[filter].splice(newFilters[filter].indexOf(option), 1)
     setFilters(newFilters)
     onFilterChange(newFilters)
   }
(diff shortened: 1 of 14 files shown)


kurtwheeler:
No storybooks? Do the components just not lend themselves to it?

client/src/components/ProjectSearchFilterPills.js
```diff
@@ -0,0 +1,82 @@
...
+        <Box margin={{ right: 'medium', bottom: 'medium' }}>
+          <Text>Applied Filters: </Text>
+        </Box>
+      )}
+      {appliedFilters.map((f) =>
+        filters[f].map((o) => (
+          <FilterPill
+            key={`${f}-${o}`}
+            option={o}
+            onRemove={() => removeFilter(f, o)}
```
kurtwheeler, line 74: What happens if there's enough/long enough filters that they reach the edge of the page? Do they wrap or do they run into `Number of Projects:`?

kurtwheeler (approved):
This LGTM aside from that one question!

client/src/components/ProjectSearchFilterPills.js
kurtwheeler: Sweet. I'm def less worried about an individual one being too long but that's a good thought too.

Issue #79, opened 2021-09-28
Add filtered cell counts to sample model


kurtwheeler:
### Context

https://github.com/AlexsLemonade/scpca-nf/pull/37

### Problem or idea

We're going to be getting an additional metadata field. It'll need to be added to the model.

### Solution or next step

Add that field to the model.

kurtwheeler:
Closed by #96

PR #80, opened 2021-09-28
Add stats and filter-options to api-root routes and add cell_count to Sample

kurtwheeler:
## Issue Number

#79 

Is this a good field name for it? Should it be `filtered_cell_count`? I think we've been saying cell counts, but that's because we're talking about counts for multiple samples right?

## Types of changes

## Functional tests

I've added unit tests and made sure they work. I also checked the api-root to make sure the new routes are there.

## Checklist


davidsmejia (approved):
LGTM i think if we get other cell_counts we can change it to be more specific

PR #85, opened 2021-10-15
mock data command

davidsmejia:
## Issue Number

n/a

* adds a command that creates "realistic" mock data for user testing
* fixes a change on the api for the client `options/projects` -> `project-options`
* client change to make 2 api calls happen async

## Types of changes

## Functional tests

was able to load data locally

## Checklist

## Screenshots

n/a

### DIFF
--- a/api/scpca_portal/management/commands/load_mock_data.py
+++ b/api/scpca_portal/management/commands/load_mock_data.py
@@ -1,0 +1,401 @@
+import random
+from typing import Dict
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from scpca_portal.config.logging import get_and_configure_logger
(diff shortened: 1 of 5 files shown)


api/scpca_portal/management/commands/load_mock_data.py
```diff
@@ -0,0 +1,412 @@
...
+
+def get_mock_option(key: str, k=False):
+    if k:
+        return random.choices(mock_sample_options[key], k=k)
+    return random.choice(mock_sample_options[key])
+
+
+def purge_all_projects():
+    for project in Project.objects.all():
+        purge_project(project.scpca_id, False)
```
kurtwheeler, line 249: This should be True so we purge the old projects data so it doesn't need to be done manually.

PR #109, opened 2021-11-19
Save ssl certs between deploys

kurtwheeler:
## Issue Number

I think there is one somewhere but I can't find it.

This will store our SSL certs into S3 when they're created with an expiration of 30 days. If they haven't expired yet, then they'll be used instead of making new ones.

This way we won't hit certbot limits.

## Types of changes

## Functional tests

I've tested these things as best as I can by looking at the prod API instance and testing on a dev one, but I can't test it for real until at least staging.

## Checklist

## Screenshots

This is from a test object I uploaded to the bucket
image


davidsmejia (approved):
LGTM!

PR #111, opened 2021-11-23
Fix the nginx config when reusing certbot certs.

kurtwheeler:
## Issue Number

#109 

The certs were installed but nginx wasn't using them! This fixes that.

## Types of changes

## Functional tests

I made these changes to staging manually and it worked!


davidsmejia (approved):
👍

PR #112, opened 2021-11-23
Restart nginx after installing new config with certbot enabled.

kurtwheeler:
## Issue Number

#111 

Close but I forgot I had to run this to get it working. I ran just this step on staging and it worked!


davidsmejia (approved):
👍

PR #121, opened 2021-11-30
Add load_data.sh to readme.

kurtwheeler:
## Issue Number

N/A I noticed these were omitted.

I added a script without mentioning it in the readme. That was bad but I fixed it. I also added the top level readme since we haven't had one.
y screenshots that illustrate these changes.


davidsmejia (approved):
👍

Issue #138, opened 2022-01-26
Upgrade Postgres


davidsmejia:
We are currently using `engine_version = "9.6.11"` of the postgres engine and that version will no longer be supported. 
**We should do this update this during the sprint of Feb. 1st**

* create a PR with new parameter group with the appropriate target version and deploy to staging
* wait 24 hours? this might not be required as refinebio was much larger
* create a new PR to remove the old parameter group

### Last time we did something like this:
https://github.com/AlexsLemonade/refinebio/pull/2638

### Where to find the upgrade target:
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html#USER_UpgradeDBInstance.PostgreSQL.MajorVersion.Process

>We strongly encourage customers to upgrade their PostgreSQL 9.6 DB instances to version 12 or higher as soon as possible, ahead of this end-of-life schedule. The PostgreSQL community discontinued support for PostgreSQL 9.6 on November 11, 2021 [2], and the community will no longer produce bug fixes or security patches for this version. PostgreSQL 12 contains major innovations including JSON path queries per the SQL/JSON specifications and pluggable table storage interface [3]. PostgreSQL 13 introduced parallel processing of indexes with the VACUUM command and improved duplicate data handling by B-tree indexes [4]. Both versions contain numerous fixes to various software bugs in earlier versions of the database.

>We recommend that you first upgrade to minor version 9.6.20 or higher and then upgrade directly to PostgreSQL 12 or higher to skip intermediate major versions. You can initiate an upgrade of your database instance, either immediately or during your next maintenance window, to a newer version of PostgreSQL using the AWS Management Console or the AWS Command Line Interface. The upgrade process will shut down the database instance, perform the upgrade, and restart the database instance. The database instance may be restarted multiple times during the upgrade process. While major version upgrades typically complete within the standard maintenance window, the duration of the upgrade depends on the number of objects within the database. To avoid any unplanned unavailability outside your maintenance window, we recommend that you first take a snapshot of your database and test the upgrade to get an estimate of the upgrade duration. To learn more about upgrading PostgreSQL major versions in Amazon RDS, review the Upgrading Database Versions page [5].
```

PR #149, opened 2022-03-03
Upgrade to Postgres 12

davidsmejia:
## Issue Number

#138 

* Modifies the parameter group to use `family = postgres12`
* adds lifecycle block that prevents the parameter group from being deleted
  * this should prevent us from having to create a new parameter group and then committing again to remove it
* sets the database version to `12.10` (current minor version)

## Types of changes

- Dependency Update

## Functional tests

N/A

## Checklist

## Screenshots

n/a


dvenprasad (approved):
🐠

PR #150, opened 2022-03-09
Add description of bulk files to README template

### LINKED ISSUE #144: Add bulk files to readme template
### Context

Currently the `README` copy does not reflect the contents of a download if there is bulk RNA sequencing since there now may be two extra files, `bulk_quants.tsv` and `bulk_metadata.tsv`. 

### Solution or next step
We should update the `README` copy to include information that if a project contains bulk, they will have those two extra files. We can follow a similar format to how we included it in the [`download_files` section of the docs](https://github.com/AlexsLemonade/scpca-docs/blob/main/docs/download_files.md#download-folder-structure-for-project-downloads).

allyhawkins:
## Issue Number
Closes #144 

Here I'm adding the description of the files that are included in a download if a project has bulk RNA-seq. I just took the sentences that describe what the files are from the download files section of the docs, since here we are only describing the contents of the download. I didn't think there was anything else to expand on other than listing the two files, what they contain, and then we have the link to the download section of the docs. 

I also removed the `metadata.json` from the list of files included in the single cell download since we do not include that file anymore and updated the name for the `single_cell_metadata.tsv` file.


jashapiro (approved):
LGTM!

davidsmejia (approved):
LGTM!

PR #155, opened 2022-03-11
add new parameter group

davidsmejia:
## Issue Number

#138 

Adds a new parameter group because the lifecycle create_before_destroy does not seem to create before destroying.

## Types of changes

## Functional tests

n/a

## Checklist

## Screenshots

n/a

### DIFF
--- a/infrastructure/database.tf
+++ b/infrastructure/database.tf
@@ -1,8 +1,33 @@
 # This terraform file hosts the resources directly related to the
 # postgres RDS instance.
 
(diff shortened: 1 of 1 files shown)


davidsmejia:
The vercel deploy will fail because the statically generated sites cannot be generated while the API is down. They will fail again after merge, I will manually redeploy them after terraform has terraformed.

jaclyn-taroni (approved):
🦖

PR #168, opened 2022-03-17
sort samples if not all downloadable

davidsmejia:
## Issue Number

N/A

* Adds a check to only sort initial downloads when there are more samples than downloadable samples

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

n/a


nozomione (approved):
LGTM

Issue #194, opened 2022-04-12
How do we want to handle multiple computed files per experiment / sample?


davidsmejia:
### Context

Spatial data zipped files will be generated separately along with "single cell" zipped file for experiments and samples that have that data.

### Solution or next step

Implement multiple computed files per resource

PR #203, opened 2022-04-27
Implement changes for supporting spatial data

arkid15r:
Issue #194 

Spatial data support.

- New feature (non-breaking change which adds functionality).

Unit tests are green.


api/scpca_portal/models/computed_file.py
```diff
@@ -16,35 +19,83 @@ class Meta:
         get_latest_by = "updated_at"
         ordering = ["updated_at", "id"]
 
+    class FileNames:
```
davidsmejia, line 22: can we change this to MetadataFilename or something similar for clarity

api/scpca_portal/models/sample.py
```diff
@@ -17,109 +20,164 @@ class Meta:
...
+
+    @property
+    def output_spatial_data_file_name(self):
+        return f"{self.scpca_id}_spatial.zip"
+
+    @property
+    def output_spatial_data_file_path(self):
+        return os.path.join(common.OUTPUT_DATA_DIR, self.output_spatial_data_file_name)
+
+    def create_single_cell_data_file(self, libraries_metadata: List[Dict], workflow_version: str):
```
davidsmejia, line 113: do you think this makes more sense in the ComputedFile instead of Sample? same with project?

davidsmejia (approved):
LGTM great job!

Issue #225, opened 2022-05-23
Add favicon and ensure page titles render on broswer tabs


dvenprasad:
### Context

<img width="243" alt="Screen Shot 2022-05-23 at 9 58 00 AM" src="https://user-images.githubusercontent.com/15315514/169836224-4e641ef5-ace6-485b-b15e-511ca03700e5.png">

### Problem or idea

Add a favicon (the blue lemon from resources portal) and show human readable and relevant page titles 
For example: The below image should say : About- ScPCA Portal 

<img width="314" alt="Screen Shot 2022-05-23 at 9 58 06 AM" src="https://user-images.githubusercontent.com/15315514/169836841-b81a2119-fdf2-46da-b805-8f6c403d0a2b.png">

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

nozomione:
Related issue: https://github.com/AlexsLemonade/scpca-portal/pull/179
**Note:** The favicon has been added to the codebase. Once we deploy the change to Production, it will load correctly, thanks!

PR #232, opened 2022-05-24
225 Add favicon and ensure page titles render on broswer tabs

nozomione:
## Issue Number
#225 

Show human readable and relevant page titles

## Types of changes
Added a custom title for each page

(reference) https://nextjs.org/docs/advanced-features/custom-document#caveats

## Checklist

## Screenshots


dvenprasad:
For the individual project pages, could we please display the title of the project? 
<Project title> - ScPCA Portal like we do for refine.bio

davidsmejia:
We should do this programatically. You can make a component called `PageTitle` for example that we would include in `_app.js`. That component would use the hook `useRouter` to get the router object and then we would map the `router.pathnames` to the titles that we would like. 

The only possible exception would be the project page which would be rendered in the Page and then append the data that is fetched (the title for a project for example) to show that information in there. 

There might be a way to access that data from the parent component but I am not sure right now. You can check to see if anything is passed in `_app.js` perhaps `pageProps` will have that information that we can pass to the `PageTitle` component.

davidsmejia:
We should do this programatically. You can make a component called `PageTitle` for example that we would include in `_app.js`. That component would use the hook `useRouter` to get the router object and then we would map the `router.pathnames` to the titles that we would like. 

The only possible exception would be the project page which would be rendered in the Page and then append the data that is fetched (the title for a project for example) to show that information in there. 

There might be a way to access that data from the parent component but I am not sure right now. You can check to see if anything is passed in `_app.js` perhaps `pageProps` will have that information that we can pass to the `PageTitle` component.

nozomione:
Hello, I've refactor the code based on your feedback. Thank you @davidsmejia !
1. Created the `PageTitle` component using the `useRouter` hook. 
2. Render the page titles programmatically in `_app` 
3. For each project page title, it includes its corresponding project ID

The page titles are rendered as follow:
   -  **Landing:** ScPCA Portal
   -   **About:** About - ScPCA Portal
   -  **Browse:** Browse Projects -  ScPCA Portal
   -  **Per Project:** Project ID SCPCPXXXXXX - ScPCA Portal

Let me know if you have any preferences/suggestions @dvenprasad !

dvenprasad:
Landing Page, About, Browse looks good to me! 

For Project, I think we should display the titles (even if they are long) rather than the IDs, since the titles convey more meaning and as users open multiple tabs it's less likely that they'll be able to differentiate between the IDs whose first ten characters are the same (even if we remove Project ID).  

Screen Shot 2022-06-22 at 10 16 50 AM

dvenprasad (approved):
Looking good! 🐯 🐯 🐯

nozomione:
> Landing Page, About, Browse looks good to me!
> 
> For Project, I think we should display the titles (even if they are long) rather than the IDs, since the titles convey more meaning and as users open multiple tabs it's less likely that they'll be able to differentiate between the IDs whose first ten characters are the same (even if we remove the words 'Project ID').
> 
> Screen Shot 2022-06-22 at 10 16 50 AM

Hello, I updated the code and now it renders a title instead of an ID. Thank you for the feedback, @dvenprasad !

@davidsmejia currently a project title is fetched via HTTP call with `getServerSideProps`. Thus, each project page <title> is locally rendered in `[scpca_id]`, thank you!

client/src/pages/projects/[scpca_id].js
```diff
@@ -8,67 +8,72 @@ import { ProjectSamplesSummaryTable } from 'components/ProjectSamplesSummaryTabl
...
-            >
-              <ProjectSamplesTable
-                project={project}
-                stickies={responsive(0, 3)}
-              />
-            </Box>
-          </Tab>
-        </Tabs>
+    <>
+      <Head>
```
Comment (davidsmejia) [inline client/src/pages/projects/[scpca_id].js:22]: Instead of this can we do something like: `<PageTitle title={`project.title`} />` and just have the component render out that value with `- ScPCA Portal` as a suffix? 
So the logic would stay the same for the homepage etc (we should probably add a title for the about page).
Also, I believe you can remove the changes from `pages/projects/index.js` as the component in `_app.js` will result in the same title.

Comment (nozomione) [reply client/src/pages/projects/[scpca_id].js]: Updated according to the feedback, thank you David!
(for the about page, currently its title is rendered as **About - ScPCA Portal,** pls let me know if you have any suggestions)

Comment (nozomione) [reply client/src/pages/projects/[scpca_id].js]: All the changes have been reflected. Thank you for the feedback, David!

Issue #223, opened 2022-05-20
ComputedFiles workflow_version is empty


davidsmejia:
### Context

Workflow version is missing on `ComputedFiles`
https://api.staging.scpca.alexslemonade.org/v1/computed-files

This affects projects now.

### Problem or idea

We should ensure that it is added when loading data.

### Solution or next step

Determine why it is missing and update to pass workflow_version to `ComputedFile` model creation.

PR #233, opened 2022-05-25
Populate project's computed files workflow version. 

arkid15r:
Display all listed versions for project/sample computed files.

#223 

Fixes missed workflow versions for project level computed files

Bugfix (non-breaking change which fixes an issue)

[x] Lint and unit tests pass locally with my changes
[x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/api/scpca_portal/models/computed_file.py
+++ b/api/scpca_portal/models/computed_file.py
@@ -8,7 +8,7 @@ from django.db import models
 import boto3
 from botocore.client import Config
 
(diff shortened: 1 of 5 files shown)


davidsmejia (approved):
LGTM!

One note: I think it might be better to omit duplicates in `utils.join_workflow_versions`

arkid15r:
Looks cleaner now, thank you!

PR #234, opened 2022-05-25
Add `skip-input-bucket-sync` flag.

arkid15r:
Add a flag to skip input bucket data sync when it's not needed.


arkid15r:
> Could we rename this option to `--skip-sync`?
> 
> I think this is a great idea!

Renamed, thank for the suggestion!

PR #237, opened 2022-05-26
Skip sample data files creation if sample directory is not present.

arkid15r:
The samples will be always created. The computed files will be generated when sample directory exists only.


davidsmejia (approved):
LGTM

Issue #228, opened 2022-05-24
Add has_bulk to the sample model + populate during load-data


davidsmejia:
We need to add `has_bulk` to the individual samples and not just at the project level.

PR #238, opened 2022-05-31
Add `has_bulk_rna_seq` to Sample model.

arkid15r:
#228 

New feature (non-breaking change which adds functionality)

Add the new field and populate it during data loading based on sample_id presence in the bulk metadata file.
Set project's `has_bulk_rna_seq` setting based on project's metadata file value.


api/scpca_portal/models/sample.py
```diff
@@ -60,11 +61,12 @@ def create_from_dict(cls, data, project):
         sample = cls(
             additional_metadata=additional_metadata,
             age_at_diagnosis=data["age"],
-            cell_count=data["cell_count"],
+            cell_count=data.get("cell_count", 0),
```
davidsmejia, line 64: I think we need to confirm that this should be either 0 or null where we would display 'N/A' on the client.

arkid15r: We certainly do.
Would you prefer to address it in this PR or a separate one?

davidsmejia: Ill make an issue for this

davidsmejia (approved):
LGTM

PR #240, opened 2022-06-02
Fix gunicorn dependency.

arkid15r:
Remove gunicorn from the Django config.
Remove new-relic from local env.


davidsmejia (approved):
LGTM!

PR #245, opened 2022-06-08
miscellaneous fixes

davidsmejia:
## Issue Number

n/a

* fixes the modalities on the samples table
* dynamically figures out which data is inside of the computed file

## Types of changes

## Functional tests

n/a

## Checklist

## Screenshots

n/a


api/scpca_portal/models/sample.py
```diff
@@ -38,6 +38,25 @@ class Meta:
     def __str__(self):
         return f"Sample {self.scpca_id} of {self.project}"
 
+    @property
+    def modalities(self):
+        print(self.scpca_id)
```
arkid15r, line 43: Remove print.

client/src/config/downloadOptions.js
```diff
@@ -1,28 +1,32 @@
...
+    included: {
+      has_bulk_rna_data: 'Bulk RNA-seq data',
+      has_cite_seq_data: 'CITE-seq data'
+    },
+    metadata: 'Project and Sample Metadata'
   },
   SAMPLE_SPATIAL_ZIP: {
-    items: ['Spatial data', 'Project and Sample Metadata'],
-    header: ' Download Spatial Data'
+    header: ' Download Spatial Data',
```
arkid15r, line 27:
```suggestion
    header: 'Download Spatial Data',
```

api/scpca_portal/models/sample.py
```diff
@@ -38,6 +38,25 @@ class Meta:
     def __str__(self):
         return f"Sample {self.scpca_id} of {self.project}"
 
+    @property
```
arkid15r, line 41: I'd prefer to have this in properties section under static methods

client/src/components/Download.js
```diff
@@ -110,8 +110,9 @@ export const Download = ({ icon, resource }) => {
             <DownloadToken />
           ) : !publicComputedFile ? (
             <DownloadOptions
-              handleSelectFile={handleSelectFile}
+              resource={resource}
               computedFiles={resource.computed_files}
```
nozomione, line 114: **Suggestion:** since we're now passing the `resource`, can we remove the line `computedFiles={resource.computed_files}` and destructure the `computed_files` value in `DownloadOptions` instead? 
(Or we'll keep it for the sake of readability?)

api/scpca_portal/models/sample.py
```diff
@@ -38,6 +38,25 @@ class Meta:
...
+        if self.has_cite_seq_data:
+            modalities.append("CITE-seq")
+
+        if self.has_spatial_data:
+            modalities.append("Spatial Transcriptomics")
+
+        if len(modalities) != 0:
+            return ", ".join(modalities)
+
+        return "N/A"
```
arkid15r, line 58: Not sure if that's the approach we normally use. Would setting it to N/A on the front-end for null modalities values look better?

nozomione (changes requested):
The update changes look great. Please see my suggestion before merging. Thank you, David!

api/scpca_portal/models/sample.py
davidsmejia: its kinda a mix, we have some values as N/A and others are null. I think youre right though. Null might make more sense here

client/src/components/Download.js
davidsmejia: yeah thats a good call, done!

api/scpca_portal/models/sample.py
```diff
@@ -88,6 +88,24 @@ def get_output_single_cell_metadata_file_path(scpca_sample_id):
...
+        if self.has_cite_seq_data:
+            modalities.append("CITE-seq")
+
+        if self.has_spatial_data:
+            modalities.append("Spatial Transcriptomics")
+
+        if modalities:
+            return ", ".join(modalities)
+
+        return None
```
arkid15r, line 107:
```suggestion

```

arkid15r: No return == return None

arkid15r (approved):
LGTM

nozomione (approved):
LGTM

api/scpca_portal/models/sample.py
davidsmejia: yeah I tend to prefer explicit returns in languages that are less strict about that, but I'm happy to defer to your preference

arkid15r: I see. For me it just looks cleaner this way.
Thank you!

Issue #230, opened 2022-05-24
Import multiplexed data with load-data script


davidsmejia:
### Context

We now have enough data to start building out the multiplexed portion of the load-data script.

### Problem or idea

After addressing:
* #227
* #229

PR #246, opened 2022-06-08
Add multiplexed samples metadata support.

arkid15r:
#230 

 - Add multiplexed computed file types
 - Add multiplexed data import business logic
 - Refactor existing code: split `def load_data`, refactor test_load_data.py

## Types of changes

- Existing code refactoring

## Checklist


davidsmejia (approved):
This looks good to me. 

Before we merge could you pull in dev I believe the migrations are already in that branch.

I just wanted to suggest that maybe we would be better served by pulling some of the load data logic and properties into another module or two. As some methods only make sense when the files are locally available. That is to say if you accessed `project.output_multiplexed_computed_file_name`, I don't believe we should expect that to exist outside of the time the file is being generated as a property of the project. I think we should give it some thought at least.

For the functions that generate computed files and their zips. It looks like some of the overall patterns are very similar and we could probably reduce some repetition there. 

Great job getting this to look much more standardized though!

arkid15r:
Thank you, David!

I appreciate the constructive review! I've extracted your comments to a separate [issue](https://github.com/AlexsLemonade/scpca-portal/issues/2640).

The load data script (and friends) is still a tough cookie. I'll make sure to address the details you're concerned about as soon as I have a chance.

PR #263, opened 2022-06-28
Update actions in run linters job(minor fix)

nozomione:
## Issue Number
N/A

Updated the actions(minor revisions)

The following changes were made in `Run linters` job:
* In the `Set up Node.js` action, the node version is now specified(the same version as the official Node Docker image we use for the client) 
   * updated `setup-node` action version to `v3`(latest) 
   *  `node-version` parameter is optional but recommended
   (resource: https://github.com/actions/setup-node#usage)
* In the `Install Node.js dependencies` action, `yarn install` command now includes `--frozen-lockfile` flag
  * It's the equivalent of [`npm ci`](https://docs.npmjs.com/cli/v7/commands/npm-ci) and is recommended to pass for better performance in CI process
  (resource: https://classic.yarnpkg.com/en/docs/cli/install) 

## Types of changes


davidsmejia (approved):
LGTM

PR #280, opened 2022-07-12
Add bulk object creation support

arkid15r:

Make `bulk_create` operation available for ComputedFile and Sample models.

## Types of changes

## Checklist


davidsmejia (commented):
I think we could have a short discussion about this...

After thinking about it for a bit, I think it makes more sense to change the names of the methods and change them to `get_project_multiplexed_files` etc, and omit saving the object inside the method. 

I also think the `bulk_create` might be unnecessary since we can just call that directly where we handle the bulk creation. I'm not completely opposed to it just feels a bit extra. 

I feel more strongly about the first comment than the second but open to discussion about either.

davidsmejia (approved):
LGTM!

Issue #266, opened 2022-07-01
Remove extraneous metadata keys


davidsmejia:
### Context

I noticed this when inspecting the API and we might need to rework how these values are written both to the db and also the metadata files. Having these attributes be present might be confusing for end users.

The following values should only be present when the project contains at least one sample that has that modality:

Known metadata keys that we do not expect to appear in every project/sample
* `has_bulk_rna_seq`
* `has_cite_seq_data`
* `has_spatial_data`
* `multiplexed_with`

### Solution or next step

We will need to add these conditionally.

davidsmejia:
Actually I don't think it makes sense for projects to have `multiplexed_with` at all. We should probably remove that from the `additional_metadata` for all projects explicitly if the above changes don't already take care of that.

PR #282, opened 2022-07-15
Remove extraneous metadata keys.

arkid15r:
## Issue Number
A good start to complete https://github.com/AlexsLemonade/scpca-portal/issues/266

  - Set `has_cite_seq_data` and `has_multiplexed_data` from project_metadaa.csv
  - Make library/sample field dropping code more consistent
  - Get ignored library/sample metadata keys based on sample modalities
  - Include `has_bulk_rna_seq`, `has_cite_seq_data`, `has_spatial_data` in relevant additional metadata only
  - Don't show `multiplexed_with` in project's additional metadata
  - Introduce modalities/names in `Sample` model
  - Update tests

## Types of changes

## Checklist


api/scpca_portal/models/project.py
```diff
@@ -55,6 +55,26 @@ def get_input_project_metadata_file_path():
     def computed_files(self):
         return self.project_computed_files.order_by("created_at")
 
+    @property
+    def ignored_additional_metadata_keys(self):
```
davidsmejia, line 59: two comments here:
* I don't think we need to have the individual mappings yet since we should be keeping everything in `additional_metadata` that isn't an attribute on the model itself
* I think we can remove the negation and just filter out the keys that we don't want to put in `additional_metadata` and just return the keys that we want to keep

api/scpca_portal/models/project.py
```diff
@@ -290,13 +241,15 @@ def combine_multiplexed_metadata(
...
-            - set(self.output_multiplexed_metadata_ignored_fields["single_cell"])
+        modalities = (Sample.MULTIPLEXED_MODALITY,)
+        ignored_library_metadata_keys = self.get_ignored_library_metadata_keys(
+            modalities=modalities
         )
+        ignored_sample_metadata_keys = self.get_ignored_sample_metadata_keys(modalities=modalities)
+
+        # Get all the field names to pass to the csv.DictWriter
+        all_fields = set(multiplexed_libraries_metadata[0].keys()) - ignored_library_metadata_keys
+        all_fields.update(set(samples_metadata[0].keys()) - ignored_sample_metadata_keys)
```
davidsmejia, line 252: I can't comment below this line but around 255 I think we should iterate over the keys in the metadata that exist and sort them based on the list of `output_multiplexed_metadata_field_order`.

arkid15r: Reworked based on your comment.

api/scpca_portal/models/project.py
arkid15r: These ideas were great! Thank you for the suggestions, David!

davidsmejia (approved):
LGTM

Issue #258, opened 2022-06-22
Support Multiplex on Portal


davidsmejia:
We need to incorporate the following changes to support mutliplexed downloads.

Designs:
https://www.figma.com/file/S3cC0Wb0nHPBEMVF0mgGxi/Multiplexed-Samples?node-id=0%3A1

PR #284, opened 2022-07-15
258 -  Support Multiplexed on Portal 

nozomione:
## Issue Number
#258  

Added support for Multiplexed downloads.

## Types of changes

## Checklist

## Screenshots

N/A


client/src/components/Download.js
```diff
@@ -45,10 +51,41 @@ export const Download = ({ icon, resource }) => {
...
+    setToggleFile(true)
+    setInital(false)
+    setDownload(false)
+  }
+
+  useEffect(() => {
+    if (initial || (!initial && !toggleFile)) setResource(initialResource)
+
+    setPublicComputedFile(
+      mutltipleComputedFiles ? null : resource.computed_files[0]
```
davidsmejia, line 67: lets write a helper her to get `getDefaultComputedFile` which returns resource.computed_files[0]

client/src/components/Download.js
```diff
@@ -105,10 +142,18 @@ export const Download = ({ icon, resource }) => {
               resource={resource}
               computedFile={download}
               handleSelectFile={handleSelectFile}
+              hasToggleFile={
```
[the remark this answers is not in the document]
nozomione: Done, thank you!

davidsmejia:
This is looking good. One recommendation would be to instead of assigning `isVisible` on every column. We should just loop over the columns and if that property is not defined, set it to true, since that is the default behavior that we want.

nozomione:
> assigning `isVisible` on every column

Completed, thank you for the feedback!

client/src/components/Download.js
```diff
@@ -1,31 +1,41 @@
...
+  const [togglePublicComputedFile, setTogglePublicComputedFile] =
+    useState(false)
   const [showing, setShowing] = useState(false)
   const [download, setDownload] = useState(false)
   const label = isProjectID(resource.scpca_id)
     ? 'Download Project'
     : 'Download Sample'
 
+  const defaultComputedFile = getDefaultComputedFile(resource)
+  const mutltipleComputedFiles = hasMultiple(resource.computed_files)
```
davidsmejia, line 37:
```suggestion
  const multipleComputedFiles = hasMultiple(resource.computed_files)
```

client/src/components/WarningText.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
+import { Box, Paragraph } from 'grommet'
+import { Icon } from 'components/Icon'
+import { Link } from 'components/Link'
+
+export const WarningText = ({
```
davidsmejia, line 6: Can we use the same interface for passing the margins and pads as grommet? I don't see any benefit of passing in an array and mapping that to the sides.

davidsmejia (approved):
LGTM just a few suggestions!

client/src/components/Download.js
nozomione: Done, thank you!

client/src/components/WarningText.js
nozomione: Done, thank you!

PR #291, opened 2022-07-20
Add `has_single_cell_data` to `Project` and `Sample` models.

arkid15r:
## Issue Number

Resolves https://github.com/AlexsLemonade/scpca-portal/issues/288

## Purpose/Implementation Notes

  - Add migrations.
  - Set `has_single_cell_data` flag values during data loading.
  - Add `unavailable_samples_count` to `Project` model.
  - Populate it with a number of samples without any available data (multiplexed/single-cell/spatial).

## Types of changes
- New feature (non-breaking change which adds functionality)

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #296, opened 2022-07-27
289 - Prevent undownloadable sample to be rendered as downloadable in the UI

nozomione:
## Issue Number

Related BE PR:  #291

Prevented the unavailable sample to be rendered as the downloadable sample in the UI.
(Adjusted the codebase based on the BE update)

## Types of changes

## Checklist

## Screenshots

N/A


client/src/components/ProjectHeader.js
```diff
@@ -5,16 +5,20 @@ import { Box, Grid, Text } from 'grommet'
...
 import { accumulateValue } from 'helpers/accumulateValue'
 import { capitalize } from 'helpers/capitalize'
 import { formatBytes } from 'helpers/formatBytes'
 import { getReadable } from 'helpers/getReadable'
-import { InfoText } from './InfoText'
-import { WarningText } from './WarningText'
 
 export const ProjectHeader = ({ project, linked = false }) => {
   const { responsive } = useResponsive()
+  const isSampleCountEqualToDownloadableCount =
```
davidsmejia, line 18: I think here we should just go off of the fact that `unavailable_sample_count` is non-zero. The point of this attribute is to say "these are the samples that should have computed files but dont yet"

nozomione: Completed, thank you!

davidsmejia (approved):
LGTM!

Issue #328, opened 2022-11-28
Display of multiple modalities in sample table needs separator


jashapiro:
### Context

Some projects  (one?) have multiple"Other modalities", and the web view does not currently display those correctly; they are 

### Problem or idea

Multiple modalities on in the sample table should be presented as a comma-separated list, or similarly separated, but currently display as a single string, mashed together (see screenshot from https://scpca.alexslemonade.org/projects/SCPCP000009):
<img width="163" alt="Screenshot 2022-11-28 at 9 34 27 AM" src="https://user-images.githubusercontent.com/20483/204304032-265fc4a2-1d53-493e-af8e-204f540da645.png">

### Solution or next step

I am guessing this a front-end issue where we are getting an array and not properly joining with separators, but if the back end is returning a string, the fix would need to be done there!

PR #329, opened 2022-11-28
328 - Display of multiple modalities in sample table needs separator

nozomione:
## Issue Number
#328 

**API:** https://api.scpca.alexslemonade.org/v1/projects/SCPCP000009

Added a separator for multiple modalities in the sample table.

## Types of changes

## Functional tests

## Checklist

### DIFF
--- a/client/src/components/ProjectSamplesTable.js
+++ b/client/src/components/ProjectSamplesTable.js
@@ -78,12 +78,17 @@ export const ProjectSamplesTable = ({
     { Header: 'Technology', accessor: 'technologies' },
     {
       Header: 'Other Modalities',
-      accessor: ({ modalities }) =>
-        modalities.map((modality, i) => (
-          <>
-            {i ? ' ,' : ''} {modality}
-          </>
-        )) || 'N/A'
+      accessor: ({ modalities }) => (
+        <>
+          {modalities.length
+            ? modalities.map((modality, i) => (
+                <>
+                  {i ? ', ' : ''} {modality}
+                </>
+              ))
+            : 'N/A'}
+        </>
+      )
     },
     { Header: 'Disease Timing', accessor: 'disease_timing' },
     { Header: 'Tissue Location', accessor: 'tissue_location' },
--- a/client/src/components/ProjectSamplesTable.js
+++ b/client/src/components/ProjectSamplesTable.js
@@ -79,15 +79,7 @@ export const ProjectSamplesTable = ({
     {
       Header: 'Other Modalities',
       accessor: ({ modalities }) => (
-        <>
-          {modalities.length
-            ? modalities.map((modality, i) => (
-                <>
-                  {i ? ', ' : ''} {modality}
-                </>
-              ))
-            : 'N/A'}
-        </>
+        <>{modalities.length ? <>{modalities.join(', ')}</> : 'N/A'}</>
       )
     },
     { Header: 'Disease Timing', accessor: 'disease_timing' },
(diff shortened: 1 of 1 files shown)


client/src/components/ProjectSamplesTable.js
```diff
@@ -78,7 +78,12 @@ export const ProjectSamplesTable = ({
     { Header: 'Technology', accessor: 'technologies' },
     {
       Header: 'Other Modalities',
-      accessor: ({ modalities }) => modalities || 'N/A'
+      accessor: ({ modalities }) =>
+        modalities.map((modality, i) => (
+          <>
+            {i ? ' ,' : ''} {modality}
+          </>
+        )) || 'N/A'
```
jashapiro, line 86: Thanks for doing this so fast!

As I'm not really sure of the data structures here, I'm not sure whether we can just use `modalities.join(', ')` to do this more simply as we do for the additional metadata? Are the `<>` & `</>` required?

Either way, we should reverse the comma space order, as suggested below:
```suggestion
      accessor: ({ modalities }) =>
        modalities.map((modality, i) => (
          <>
            {i ? ',' : ''} {modality}
          </>
        )) || 'N/A'
```

nozomione: Revised the spacing (oops), thank you Josh!

**Note:** The `<></>`(short form of `<React.Fragment>`) is called [fragments](https://reactjs.org/docs/fragments.html) and is the way to properly render [JSX](https://reactjs.org/docs/introducing-jsx.html) in React UI :)

jashapiro (approved):
Thanks again for getting to this so fast! 

I have one more question, which is not about a change here, as this seems to preserve existing behavior, but about what might come after this in code. When I look at the portal now, if there are no "other modalities" for a sample, that cell is blank, which seems correct. But here there is an `N/A` being filled in: does that get replaced later with a blank? Or is this a change that hasn't yet percolated to the live version of the portal, and we will display "N/A" in the future?

dvenprasad:
Based on the designs, I would expect blanks to be populated as `NA`. 
@jashapiro  Is a blank cell preferred? We can also use `None` which might work better for this field. 

Screen Shot 2022-11-28 at 2 33 26 PM

davidsmejia (approved):
LGTM

client/src/components/ProjectSamplesTable.js
```diff
@@ -78,7 +78,9 @@ export const ProjectSamplesTable = ({
     { Header: 'Technology', accessor: 'technologies' },
     {
       Header: 'Other Modalities',
-      accessor: ({ modalities }) => modalities || 'N/A'
+      accessor: ({ modalities }) => (
+        <>{modalities.length ? <>{modalities.join(', ')}</> : 'N/A'}</>
```
davidsmejia, line 82:
```suggestion
        <>{modalities.length ? modalities.join(', '): 'N/A'}</>
```

doesn't this work all the same?

jashapiro:
> Based on the designs, I would expect blanks to be populated as `NA`. @jashapiro Is a blank cell preferred? We can also use `None` which might work better for this field.
> 
> Screen Shot 2022-11-28 at 2 33 26 PM

I think "None" is better for this field (and maybe also for "Additional Metadata Fields", though I expect that is never null).

client/src/components/ProjectSamplesTable.js
nozomione: Complented, thank you!

PR #345, opened 2022-12-22
Optimize data loading process disk usage

arkid15r:
## Issue Number
NA

Clean up output data for each project once it's processed.

## Types of changes

## Functional tests
NA

## Checklist

## Screenshots
NA

### DIFF
--- a/api/scpca_portal/management/commands/load_data.py
+++ b/api/scpca_portal/management/commands/load_data.py
@@ -82,6 +82,7 @@ class Command(BaseCommand):
 def cleanup_output_data_dir():
     cleanup_items = (
         ComputedFile.README_FILE_NAME,
+        ComputedFile.README_MULTIPLEXED_FILE_NAME,
         ComputedFile.README_SPATIAL_FILE_NAME,
     )
     for item in cleanup_items:
@@ -211,7 +212,7 @@ def load_data_from_s3(
         if cleanup_output_data:
             logger.info(f"Cleaning up '{project}' output data")
             for computed_file in computed_files:
-                os.unlink(computed_file.zip_file_path)
+                Path(computed_file.zip_file_path).unlink(missing_ok=True)
 
             for path in Path(common.OUTPUT_DATA_DIR).glob("*.tsv"):
-                path.unlink()
+                path.unlink(missing_ok=True)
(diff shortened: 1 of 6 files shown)


api/scpca_portal/management/commands/load_data.py
```diff
@@ -80,14 +80,19 @@ def handle(self, *args, **options):
...
+        ComputedFile.README_SPATIAL_FILE_NAME,
+    )
     for item in cleanup_items:
         for path in Path(common.OUTPUT_DATA_DIR).glob(item):
             path.unlink()
 
 
 def load_data_from_s3(
     allowed_submitters: set = ALLOWED_SUBMITTERS,
+    cleanup_output_data=not (settings.DEBUG or settings.TEST),
```
davidsmejia, line 95:
```suggestion
    default_cleanup_output_data: bool = IS_PRODUCTION),
```

davidsmejia: Add to the top of the file, under `ALLOWED_SUBMITTERS`:
```python
IS_PRODUCTION = not (settings.DEBUG or settings.TEST)
```

davidsmejia: Also add another parameter that overrides the option `default_cleanup_output_data` so if we want to persist the data for debugging we can without changing the settings in production.

api/scpca_portal/management/commands/load_data.py
```diff
@@ -203,3 +208,11 @@ def load_data_from_s3(
                     settings.AWS_S3_BUCKET_NAME,
                     computed_file.s3_key,
                 )
+
+        if cleanup_output_data:
```
davidsmejia, line 212: if you add the parameter to allow for overriding:
```suggestion
        if cleanup_output_data or default_cleanup_output_data:
```

davidsmejia (changes requested):
Also you can scale the volume size to 100GB and test running everything available on staging. https://github.com/AlexsLemonade/scpca-portal/blob/dev/infrastructure/api.tf#L89

I would create a new ticket to make sure this gets tested.

api/scpca_portal/management/commands/load_data.py
arkid15r: I've added the implementation based on your comments.

arkid15r:
PTAL

davidsmejia (approved):
LGTM

PR #349, opened 2022-12-23
Fix - Downgrade the eslint version to 7.32.0 to prevent error

nozomione:
## Issue Number
N/A

Include an issue number if applicable.

## Purpose/Implementation Notes

Downgraded the eslint version to `7.32.0` to prevent an error during deploy(which was observed locally as well) 

(resource) https://github.com/vercel/next.js/issues/30182

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

## Functional tests

List out the functional tests you've completed to verify your changes work locally.

## Checklist

- [x] Lint and unit tests pass locally with my changes

## Screenshots

Please attach any screenshots that illustrate these changes.

PR #350, opened 2022-12-23
Fix Vercel deploy

arkid15r:
(fix) downgrade the eslint version to 7.32.0(recommended: https://github.com/vercel/next.js/issues/30182)to prevent an error (#349)


nozomione (approved):
LGTM

PR #388, opened 2023-05-05
Add md and pdf versions of call for contributions guidelines

jaclyn-taroni:
## Issue Number

N/A

Here I'm adding two copies of the guidelines for the call for contributions:

* Markdown to `client/src/config`
* PDF to `client/public`

Soon, these will be available at `scpca.alexslemonade.org/contribute` with the option to download the PDF version.

Note to reviews: Please make sure these are in the right place!

## Types of changes

## Functional tests

N/A

## Checklist


davidsmejia:
This looks good to me mostly. There are two changes that I am happy to take care of:
- trim whitespace to satisfy linter
- escape backticks (ex: \` -> \\` ) so the loader will work.

I already pulled and made these changes to test locally and we get the `<code>` tag that we can style however we want for this page.

The default rendering looks back but we just need to use the overrides to tell grommet which components we want for parsed markdown.

Screen Shot 2023-05-05 at 5 55 31 PM

jaclyn-taroni:
Thanks, @davidsmejia! I went ahead and made those changes.

> * trim whitespace to satisfy linter

Now feels like a good time to tell you that not every editor designed for Markdown will trim trailing whitespace upon save 😄

PR #416, opened 2023-07-19
Update the staging/prod to fully support available commands

### LINKED ISSUE #414: Update the staging / prod command to fully support available flags.
### Context

We currently have a templated script for loading data. This script when generated only allows the caller to load data into the portal. Script: https://github.com/AlexsLemonade/scpca-portal/blob/dev/infrastructure/api-configuration/load_data.tpl.sh

### Problem or idea

The problem is that sometimes we need to purge a project from the prod or staging env which either requires duplicating the script and changing some arguments or editing it, running it, and then editing it again. We should either create a new templated script for removing projects / samples or replace it with something that allows us to  purge and load.

### Solution or next step

- Add the script or update the existing one

arkid15r:
## Issue Number

Resolves #414 

Replace load data script with run command script. 
Update requirements in order to fix the stack build process.

  - Add run_command.sh to support any management
      command in production
  - Update versions for several base.txt dependencies
  - Update user data script
  - Update the related documentation

## Types of changes

## Checklist


davidsmejia (approved):
LGTM 🚀

PR #417, opened 2023-07-25
Attach a persistent data volume to the data loading container

### LINKED ISSUE #415: Create volume for loading data to persist downloaded files from s3
### Context

We currently keep a fairly large attached volume to the scpca ec2 instances. One change we could make that would have no additional cost to processing times would be to create a volume for the data loading docker images so when running the command we do not have to redownload the data from s3.

### Solution or next step

- Create a volume to attach to the container

arkid15r:
## Issue Number

Resolves #415 

Use part of constantly available disk space as a container data volume.

## Types of changes

## Checklist


davidsmejia (approved):
LGTM

Issue #429, opened 2023-09-11
Upgrade to Nextjs 13 / Yarn 3 / Storybook


davidsmejia:
### Context

Ref: https://github.com/AlexsLemonade/refinebio-web/issues/143

Currently, `node:14.17` is used for the client docker image. However, due to the [deprecation of Node.js 16](https://nodejs.org/en/blog/announcements/nodejs16-eol) and Vercel will [discontinue](https://vercel.com/changelog/node-js-14-and-16-are-being-deprecated) the support as a result, we'll need to upgrade the docker image to v18.

Because of this change, the following upgrades will be also made to our repository to eliminate the dependency conflict:

- Migrate the storybook v6 to v7
- Migrate the Next.js v12 to v13 
- Upgrade Yarn to v3 

### Solution or next step
- Migrate Next.js and storybook
- Upgrade the package manager yarn to the latest stable version v3

PR #432, opened 2023-09-13
429 - Upgrade to Nextjs 13 / Yarn 3 / Storybook

nozomione:
## Issue Number

Closing #429

The following changes were made in this PR: 
- Upgraded the node docker image `v14.17` to `v18.17.1`
- Migrated the storybook `v6` to `v7`
- Migrated the Next.js `v11` to `v13`
- Upgrade Yarn `v1` to `v3` in project
- Resolved the compile warnings/errors (please see commits for each warning/error)
- Excluded `./yarn` auto-generated directory from the pre-commit workflow (otherwise test will fail due to the trailing whitespace e3d361f09df3d242ef16431e38892aa95adcbe22)
- Other adjustments due to the migrations 

## Types of changes

## Checklist


docker-compose.env
```diff
@@ -11,6 +11,7 @@ DOCS_PORT=8001
 # Client
 STORYBOOK_PORT=7001
 CLIENT_PORT=7000
+NODE_IMAGE=node:18.17.1
```
davidsmejia, line 14: I don't think this is necessary. I could kinda see an argument for it if we used it in multiple places but I would expect to find the version of the container I am running in `docker-compose.yml`. If it is all the same could we keep it in that file?

davidsmejia (approved):
This looks great just one comment

docker-compose.env
nozomione: Got it, let's remove it then!

Issue #442, opened 2023-09-18
Tweaks to contribution page banner


dvenprasad:
### Context

Minor tweaks to make the banner on contributions page look pleasing

### Problem or idea

1. Warning Icon - The `!` on the warning icons should be white, not black
![warning-red](https://github.com/AlexsLemonade/scpca-portal/assets/15315514/ba5d98c5-ab3c-434f-b3d1-4b0cc55b99c7)

2. Increase padding between the navbar and the card so that there is at least 32px between the end of the ribbon logo and the card. The ribbon logo shouldn't overlap with the card.

3. Add more padding (maybe 16px) between the `We are not currently accepting contributions` and the ribbon on top of the card.

![Screenshot 2023-09-18 at 15 43 40](https://github.com/AlexsLemonade/scpca-portal/assets/15315514/253787bc-5c90-4625-95cf-2e219b587d6e)

### Solution or next step

Tagging @nozomione for next steps.

PR #443, opened 2023-09-18
442 - Tweaks to contribution page banner

nozomione:
## Issue Number
Closing #442

Made minor tweaks to contribution page banner. 

@dvenprasad please review the latest UI [here](https://scpca-portal-7h7glvvo1-ccdl.vercel.app).

## Types of changes

## Checklist


client/src/components/WarningText.js
```diff
@@ -21,7 +22,11 @@ export const WarningText = ({
       margin={{ top: 'small', bottom: 'medium' }}
     >
       <Box margin={iconMargin} pad={iconPad}>
-        <Icon color={iconColor} size={iconSize} name="Warning" />
+        <Icon
+          color={iconColor}
+          size={iconSize}
+          name={`Warning${iconNoFill ? 'NoFill' : ''}`}
```
davidsmejia, line 28: Can we just have a prop called `iconName` and default it to `WarningNoFill`. Would provide more flexibility and I'm pretty sure we will always use the "no fill" option.

davidsmejia (changes requested):
Small change requested looks good though

client/src/components/WarningText.js
davidsmejia: Ah ok, So in that case we should just switch to `iconName` and default it to `Warning` and then for this one use case we pass in `WarningNoFilll`. We won't have to update the other uses of this component and we can now expand it as we need for new iconography.

client/src/images/warning-red.svg
davidsmejia, line 1: Can we keep the icon file name and icon name matching? So, either rename this to `warning-no-fill` or change the name to be `WarningRed` in `components/icon` to match this file name.

nozomione: The icon's color is set dynamically using the `iconColor` prop, so I updated the name of the image file to be `warning-no-fill`. Please review,. thank you David.

davidsmejia (approved):
LGTM

PR #454, opened 2023-10-05
Use distinct samples for samples stats endpoint

### LINKED ISSUE #450: Samples counts do not seem correct
### Context

I was verifying some samples had been made available to the portal when I discovered a mismatch of sample counts that the API response with. The discrepancy can be found on the `count` attribute of the [samples list endpoint](https://api.scpca.alexslemonade.org/v1/samples) and in the `samples_count` attribute of [the stats endpoint](https://api.scpca.alexslemonade.org/v1/stats).

Samples endpoint: 503 samples
Stats endpoint: 541 samples

### Problem or idea

The samples count for the stats endpoint should be lower because samples without computed files (non-downloadable) are filtered out of the count.

In the samples list view we use:
```python
queryset = Sample.objects.all().order_by("-created_at")
```

And in the stats endpoint we use:
```python
Sample.objects.filter(sample_computed_files__isnull=False).count()
```

### Solution or next step

Determine if we are including duplicate samples in the stats endpoint count and remove them.

arkid15r:
## Issue Number

Resolves #450 

Update samples_count query to include distinct samples only.

## Types of changes

## Functional tests

## Checklist

### DIFF
--- a/api/scpca_portal/views/stats.py
+++ b/api/scpca_portal/views/stats.py
@@ -25,6 +25,8 @@ class StatsViewSet(viewsets.ViewSet):
                 "cancer_types_count": cancer_types_queryset.count(),
                 "labs_count": Project.objects.values("pi_name").distinct().count(),
                 "projects_count": Project.objects.count(),
-                "samples_count": Sample.objects.filter(sample_computed_files__isnull=False).count(),
+                "samples_count": Sample.objects.filter(sample_computed_files__isnull=False)
+                .distinct()
+                .count(),
             }
         )


davidsmejia (approved):
LGTM

api/scpca_portal/views/stats.py
```diff
@@ -25,6 +25,8 @@ def list(self, request):
                 "cancer_types_count": cancer_types_queryset.count(),
                 "labs_count": Project.objects.values("pi_name").distinct().count(),
                 "projects_count": Project.objects.count(),
-                "samples_count": Sample.objects.filter(sample_computed_files__isnull=False).count(),
+                "samples_count": Sample.objects.filter(sample_computed_files__isnull=False)
+                .distinct()
+                .count(),
```
arkid15r, line 30: This looks a bit weird, however black is happy so I'm keeping it as is.

Issue #455, opened 2023-10-12
 Display external accession on the portal


dvenprasad:
### Context

We would like to show other repositories where the data is also deposited (external accessions)

### Problem or idea
Add a field `Also deposited under` on:

- Project Card - show after Publications, values are comma separated. 
![Screenshot 2023-10-12 at 11 17 50](https://github.com/AlexsLemonade/scpca-portal/assets/15315514/0a06ffea-52d5-4c13-8993-9d520c983aff)

- Project Details Page - show after DOI, values are separated by a new line.
![Screenshot 2023-10-12 at 11 19 36](https://github.com/AlexsLemonade/scpca-portal/assets/15315514/ebafbff4-0d92-48f5-a21d-d3062a035ad3)

- If external accessions aren't available populate the field with `Not Available` on both project card and project details page. 
![Screenshot 2023-10-12 at 11 24 05](https://github.com/AlexsLemonade/scpca-portal/assets/15315514/fc130da1-b4db-4948-b589-adb1844d81ac)

### Solution or next step

Tagging @davidsmejia for next steps

davidsmejia:
closed with #477

PR #477, opened 2023-11-10
add support for displaying external accessions

davidsmejia:
## Issue Number

#455 

Adds support for displaying external accessions with the other project details.
We support two flavors of external accessions; comma separated list for search results, and one per line for project detail page.

Additionally adds a default array for publications for consistency and aligns table keys with values at the top like the comps.

Note: we need to do some cleanup on the API before this is ok to merge. (done 11/13)

## Types of changes

## Functional tests

Tested locally

## Checklist

## Screenshots
Search results with:
Screen Shot 2023-11-10 at 4 38 54 PM

Search results w/o:
Screen Shot 2023-11-10 at 4 39 36 PM

Project Detail Page with:
Screen Shot 2023-11-10 at 4 40 04 PM

Project Detail Page w/o:
Screen Shot 2023-11-10 at 4 41 00 PM

### DIFF
--- a/client/src/components/ProjectExternalAccessionsDetail.js
+++ b/client/src/components/ProjectExternalAccessionsDetail.js
@@ -8,7 +8,7 @@ export const ProjectExternalAccessionsDetail = ({ inline = false, externalAccess
   if (inline) {
     return externalAccessions.map(({ accession, url }, i) => (
       <React.Fragment key={accession}>
-        {i != 0 && ', '}
+        {i !== 0 && ', '}
         <Link label={accession} href={url} />
       </React.Fragment>
     ))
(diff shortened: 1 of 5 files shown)


davidsmejia:
Ok, I have manually stripped the `< >` from the external accession urls on staging and prod. Also deleted the external accession with the accession `N/A`.

dvenprasad (changes requested):
Displaying accessions looks good! 

I noticed that we say `Not Available` on the search card and `Not Specified` on the experiment page. We should make it consistent  and use `Not Specified` for both accession (and publication) fields

Not related to external accessions but publications are enclosed in `< >`, we should remove those. I can file another ticket for changes for publication field (including the above consistency issue), if that makes it easier.

nozomione:
I was going to add this suggestion but it was already merged. Perhaps in the future we can update. 

In the [`ProjecExternalAccessionDetail`](https://github.com/AlexsLemonade/scpca-portal/pull/477/files#diff-facf91b470aa906f0096bb8c0d9d726e4e18dab1601266c77aca58885e472410) at line 21, alternatively, we can use the shorthand instead of the ternary operator:
```js
 margin={{ top: i && 'small' }}
```

PR #481, opened 2023-11-13
Add AnnData support

### LINKED ISSUE #471: API support for PROJECT_ANNDATA_ZIP and SAMPLE_ANNDATA_ZIP
### Context

We are adding two types of new computed files to be created during the `load_data` command.

Adding support for this will be very similar to steps taken to support other computed_files.
The `project_metadata.tsv` will contain a new column `includes_anndata` which will indicate whether or not that project should contain the relevant files. These new files will be available in `hdf5` format.

The `computed_file.type` should be set to `PROJECT_ANNDATA_ZIP` and `SAMPLE_ANNDATA_ZIP`

Documentation for the expected downloadable zips' contents can be found here:
https://github.com/AlexsLemonade/scpca-docs/blob/development/docs/download_files.md#anndata-downloads

### Solution or next step

- Add support for creating Anndata project and sample zips to the `load_data` command.

arkid15r:
## Issue Number

Resolves #471 

Add AnnData support:
  - Add project/sample ann data zip file types
  - Add migrations
  - Update tests
  - Update date loading command

## Types of changes

## Checklist


davidsmejia (commented):
Ok, I think this is looking pretty good. I think we will want to make the following changes to support the client and other external users.

Let's address the following:

*Updated 11/16*

#### Remove AnnData from modalities
We can remove this as it is a file type that is adequately captured by the `computed_file.type` attribute.

#### Adding `file_type` to `computed_file`
This would be a new field that allows us to see which file type is contained in the zip. This is `SingleCellExperiment` by default and `AnnData` when the computed file contains AnnData.

#### Note about `includes_anndata` attribute
I think for now this is fine as is. At least I haven't formed a strong enough opinion around what would be best now vs the future. 

Here are my two competing beliefs/ concerns.

1. We probably want to rename `includes_anndata` more specific to the modality (ie `single_cell_anndata`), because hypothetically if we were to include multiplexed AnnData we would want to be able to distinguish these files. This probably means we should have a database rule so that each computed file must be unique across `project/sample`, `modality`, 'type'.
1. We aren't going to be pre-loading data forever. Eventually we will reach a size where we will want to have à la carte downloads. The changes described above most likely wont survive those changes as we will be moving away from this type of `computed_file`. Which means thinking about long term for this attribute seems unwarranted.

Let me know if you want me to clarify anything or if you have an alternative idea for addressing this.

In another PR we should make this consistent.

#### Change modality field from string to list.
Currently when we update counts on a project we concatenate modalities into a comma separated string. In order to facilitate introspection on a client we should make this an array type attribute and assign it as such.

davidsmejia (approved):
Could you add the `includes_anndata` to the project and sample serializers after that this LGTM

PR #515, opened 2024-01-16
Add AWS configuration flags

### LINKED ISSUE #509: Speed up data loading process, part 2
### Context

In order to address #501 epic we need to figure out next steps to tackle the problem.

### Problem or idea

As a possible data loading speed up solution we could benefit from better S3 download/upload speeds.

### Solution or next step

Implement configurable AWS CLI parameters to be able to control the [configuration values](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#configuration-values).

arkid15r:
## Issue Number

Resolves #509 

This PR adds ability to control AWS S3 related configuration flags.

## Types of changes

## Checklist

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -164,6 +164,27 @@ separate thread pool executors -- for project and sample processing.
 sportal load-data --max-workers 10 --reload-all --update-s3
 ```
 
+AWS S3 configuration options:
+
+The `--s3-max-bandwidth` flag controls the maximum bandwidth (in MB/s) that the S3 commands will
+utilize for S3 transfers. Default - None.
+```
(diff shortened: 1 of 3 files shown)


davidsmejia (approved):
This LGTM as an initial implementation.

My main concern is that this will be mutating the host system's aws config that is mapped to the container.

Could you create a ticket to find out if it would be possible to use envvars for these configurations. That way we could keep these setting changes in the container exclusively.

Some added context: There will be changes to how we configure the aws cli. We will be moving away from mapping the user config to the container for credentials. Instead we will use envvar credentials since that will allow us to use STS tokens generated on the host (via an op plugin).

PR #523, opened 2024-01-27
Add new ComputedFile attributes

### LINKED ISSUE #521: Add "contents" to ComputedFile
### Context

Determining the contents of a computed file aka "downloadable zip" used to be 1 to 1 association of the `computed_file`'s type to the pre-derminted contents. This is no longer the case. We should move creating this list to an attribute on `computed_file`. I would expect to name it `content_descriptions` and the value would be a list of strings.

### Problem or idea

Below are the expected outputs for various modality and file format combinations.

NOTE: Project and Sample Metadata should be included on all downloads.

#### Project
```
Single-cell data as SingleCellExperiement(R)  
Single-cell multiplexed data as SingleCellExperiment (R)  
Project and Sample Metadata
```
#### Sample
```
Single-cell multiplexed data as SingleCellExperiment (R)  
Project and Sample Metadata
```

#### Project
```
Single-cell data as SingleCellExperiement(R)  
Single-cell multiplexed data as SingleCellExperiment (R)  
Project and Sample Metadata
```

#### Project
```
Single-cell data, CITE-seq data as SingleCellExperiment (R)  
Bulk RNA-seq data  
Project and Sample Metadata
```

#### Project
```
Single-cell data as AnnData (Python)
CITE-seq data as AnnData (Python)
Bulk RNA-seq data  
Project and Sample Metadata
```

### Solution or next step

- Create a new attribute on `computed_file` called `content_descriptions` that is a List type
- Populate the list as we process projects  and sample computed files.
- Tests: serialization and extend test for data loading

arkid15r:
## Issue Number

Closes #521 

Adds the `has_bulk_rna_seq` and `has_cite_seq_data` fields to `ComputedFile` model.

## Types of changes

## Checklist


davidsmejia (commented):
I think this is a good first pass.

A couple things that don't look quite right to me:
1. Since Cell Experiment computed files that contain CITE-seq data should have one line indicating that it is in the same folder would be good to implement. (Ann Data should be two separate line items)
2. This implementation checks against if the project has the various datas, we really only want to report the contents. i.e. Spatial is unaccounted for here.

I think we should revisit adding new attributes to the `computed_file` during creation time to make this easier to handle. If we did that we could move the copy to the front end since it would have enough information to generate it. Moving this to the backend was suggested because the information about what is being put into the computed file is known at generation time.

davidsmejia:
Updated the parent issue here: https://github.com/AlexsLemonade/scpca-portal/issues/521#issuecomment-1917765495

arkid15r (commented):
I've updated PR with the changes based on the new requirements.

api/scpca_portal/models/computed_file.py
```diff
@@ -447,3 +459,29 @@ def delete_s3_file(self, force=False):
             return False
 
         return True
+
+    def generate_content_descriptions(self):
```
davidsmejia, line 463: Let's get rid of this for now, since the required properties will be on the computed file.

davidsmejia (commented):
Let's remove `ComputedFile::generate_content_descriptions` and then this looks good to me.

api/scpca_portal/models/computed_file.py
arkid15r: Yeah, I should've removed it -- it's unused.

davidsmejia (approved):
LGTM!

PR #545, opened 2024-02-08
Add celltype report to downloads

### LINKED ISSUE #496: Cell Type Annotation
### Context

This issue is to collect issues around adding Cell Type Annotation support on the ScPCA Portal.

This epic is "blocked" by the Anndata Support Epic only because we want to start this work after that is deployed to production.

Test data is available:
https://github.com/AlexsLemonade/ScPCA-admin/pull/717

arkid15r:
## Issue Number

Resolves #496 

Adds celltype report file for single-cell/multiplexed modalities when project has no cell lines.

## Types of changes

## Checklist


api/scpca_portal/models/computed_file.py
```diff
@@ -246,8 +246,17 @@ def get_sample_multiplexed_file(
...
             library_id = library["scpca_library_id"]
-            for file_suffix in ("_filtered.rds", "_processed.rds", "_qc.html", "_unfiltered.rds"):
-                file_name = f"{library_id}{file_suffix}"
+            file_suffixes = [
+                "filtered.rds",
+                "processed.rds",
+                "qc.html",
+                "unfiltered.rds",
+            ]
+            if not sample.project.includes_cell_lines:
```
davidsmejia, line 255: This should be based off of the sample/library and not the project.

`project.includes_cell_lines` just means that at least one sample/library where `sample.is_cell_line` is true.

The value that we want to base this off of can be found here:
https://github.com/AlexsLemonade/ScPCA-admin/blob/main/scpca-test-data/sample-metadata.tsv

api/scpca_portal/models/computed_file.py
```diff
@@ -280,20 +289,23 @@ def get_sample_single_cell_file(cls, sample, libraries, workflow_versions, file_
...
             file_name = sample.output_single_cell_computed_file_name
-            common_file_suffixes = (
+            common_file_suffixes = [
                 "filtered.rds",
                 "processed.rds",
                 "qc.html",
                 "unfiltered.rds",
-            )
+            ]
+        if not sample.project.includes_cell_lines:
```
davidsmejia, line 306: same comment as above

davidsmejia (commented):
Looking good, just need to make sure we check off of the sample and not the project.

arkid15r (commented):
I've changed the processing logic based on your comment. PTAL

api/scpca_portal/models/computed_file.py
arkid15r: Done.

api/scpca_portal/models/computed_file.py
arkid15r: Done.

PR #563, opened 2024-02-22
Sample.additional_metadata.multiplexed_with has moved to Sample.multiplexed_with

davidsmejia:
## Issue Number

N/A

There were some things that needed to be cleaned up to have the AnnData changes laid on top. Notably, the `sample.multiplexed_with` was undefined, and this PR resolves that by adding to the serializer and updating the accessors on the client.

Additionally, as part of unbreaking staging before AnnData support is available, this PR will filter out AnnData computed files from presentation on the client side as they currently exist on staging. This change wont affect production and is safe to ship allowing us to support two different looking backends temporarily.

Also, this was a choice, I added a second parameter to the `hasMultiple` helper. This is optional and allows us to pass a filter function to be applied to the array before checking the length. I'm not concerned with this helper as it will most likely be removed due to lack of usage post AnnData.
____
The actual changes are summarized below:

1. Corrects the accessor of `multiplexed_with` on `DownloadStarted` component. 
2. `multiplexed_with`, was added to the sample serializer.
3. Updates tests to expect to see `multiplexed_with` on the serializer (tests now pass)
4. Modified the Header of the Download Modal to reflect that the data is multiplexed. (they looked the same)
5. Adds multiplexed variants of `computed_file.type` to the computed files sorter for completeness.
6. Added back temporary filter on AnnData so that staging will be unbroken until AnnData client changes land.
7. Added optional filter to `hasMultiple` helper to help expose that we are filtering out AnnData

## Types of changes

## Functional tests

Running locally against staging data.

## Checklist

## Screenshots

N/A

### DIFF
--- a/client/src/components/Download.js
+++ b/client/src/components/Download.js
@@ -34,7 +34,12 @@ export const Download = ({ icon, resource: initialResource }) => {
     : 'Download Sample'
 
   const defaultComputedFile = getDefaultComputedFile(resource)
-  const multipleComputedFiles = hasMultiple(resource.computed_files)
+
+  // TODO:: Remove filter when ANN_DATA lands
+  const multipleComputedFiles = hasMultiple(
+    resource.computed_files,
+    f => f.format !== 'ANN_DATA'
+  )
 
   const handleClick = () => {
     setShowing(true)
--- a/client/src/components/Download.js
+++ b/client/src/components/Download.js
@@ -35,7 +35,7 @@ export const Download = ({ icon, resource: initialResource }) => {
 
   const defaultComputedFile = getDefaultComputedFile(resource)
 
-  // TODO:: Remove filter when ANN_DATA lands
+  // TODO: Remove filter when ANN_DATA lands
   const multipleComputedFiles = hasMultiple(
     resource.computed_files,
     (f) => f.format !== 'ANN_DATA'
--- a/client/src/components/Download.js
+++ b/client/src/components/Download.js
@@ -38,7 +38,7 @@ export const Download = ({ icon, resource: initialResource }) => {
   // TODO:: Remove filter when ANN_DATA lands
   const multipleComputedFiles = hasMultiple(
     resource.computed_files,
-    f => f.format !== 'ANN_DATA'
+    (f) => f.format !== 'ANN_DATA'
   )
 
   const handleClick = () => {
--- a/client/src/components/DownloadOptions.js
+++ b/client/src/components/DownloadOptions.js
@@ -5,6 +5,8 @@ import { sortArrayByKey } from 'helpers/sortArrayByKey'
 
 export const DownloadOptions = ({ resource, handleSelectFile }) => {
   const { computed_files: computedFiles } = resource
+
+  // TODO: Remove filter when ANN_DATA lands
   const sortedComputedFiles = sortArrayByKey('type', computedFiles, [
     'PROJECT_ZIP',
     'SAMPLE_ZIP',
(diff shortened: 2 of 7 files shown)


client/src/components/Download.js
```diff
@@ -34,7 +34,12 @@ export const Download = ({ icon, resource: initialResource }) => {
     : 'Download Sample'
 
   const defaultComputedFile = getDefaultComputedFile(resource)
-  const multipleComputedFiles = hasMultiple(resource.computed_files)
+
+  // TODO:: Remove filter when ANN_DATA lands
```
arkid15r, line 38:
```suggestion
  // TODO: Remove filter when ANN_DATA lands
```

arkid15r (commented):
LGTM (Python part)

client/src/components/DownloadOptions.js
```diff
@@ -5,12 +5,16 @@ import { sortArrayByKey } from 'helpers/sortArrayByKey'
 
 export const DownloadOptions = ({ resource, handleSelectFile }) => {
   const { computed_files: computedFiles } = resource
+
+  // TODO: Remove filter when ANN_DATA lands
```
nozomione, line 9: I won't comment on the code blocks marked as `// TODO: Remove filter when ANN_DATA lands`, since they'll be cleaned up in later AnnData PRs (e.g., the no longer used `type` field  value `PROJECT_MULTIPLEXED_ZIP`).

client/src/helpers/hasMultiple.js
```diff
@@ -2,9 +2,11 @@
 @name hasMultiple
 @description returns true if an array contains more than one element, otherwise false
 @param {any[]} a - an array to be checked
+@param {function} filter - optional filter function to be applied to the array
 */
-export const hasMultiple = (a) => {
-  return a.length > 1
+export const hasMultiple = (a, filter) => {
+  const arr = filter ? a.filter(filter) : a
```
nozomione, line 8: Perhaps it would be useful to add the filtered item to a config file (which you may have already done in later PRs!)

nozomione (approved):
LGTM

client/src/components/DownloadOptions.js
davidsmejia: Yeah this will be removed, temporarily added so that staging can be deployed to staging where AnnData exists on the backend but not the client. This will be removed when the AnnData content on the backend is supported on the client. Basically, the existing client doesn't know how to present AnnData computed files, so filtering them out restores the client side functionality.

client/src/helpers/hasMultiple.js
davidsmejia: Well this helper was only used to handle presenting two computed files side by side. This wont be a feature on the portal moving forward as we will be hiding all computed files behind two dropdowns. The filtering will be done with a hook that handles user selected options to present the best fit instead of trying to display multiple computed files simultaneously.

client/src/config/downloadOptions.js
```diff
@@ -33,7 +33,7 @@ export const downloadOptions = {
     }
   },
   PROJECT_MULTIPLEXED_ZIP: {
-    header: 'Download Single-cell Data',
+    header: 'Download Multiplexed Data',
```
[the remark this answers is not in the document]
davidsmejia: Yeah this is a change that was to help disambiguate this scenario. Since `computed_files.type` will be removed in the future the `header` attribute will be removed soon.

Screen Shot 2024-02-23 at 10 00 03 AM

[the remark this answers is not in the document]
nozomione: Oh I see! Thank you for letting me know!

PR #565, opened 2024-02-27
Implement merged objects support

### LINKED ISSUE #549: Add Merged Objects Support
### Context

We want to support merged data on the ScPCA-Portal. This will create new computed files and a new attribute `ComputedFile::includes_merged`. This will only affect computed files for projects. There will be 3 potential new outputs. The user will be able to select an `is_merged` computed file variant on the portal by selecting a checkbox.

### Problem or idea

This is another computed file output for `load_data`. This should be the last one we add before solving for moving our processing to AWS batch. I expect most of this process to resemble how we have added new attributes to `ComputedFile` in the past.

Test Data version can be determined by looking at the changelog:
  - https://github.com/AlexsLemonade/ScPCA-admin/blob/main/scpca-test-data/CHANGELOG.md

### Solution or next step

Expected Approach:

#### Load Data
- There will be 2 new columns in `project_metadata.csv`
  - The new columns called `includes_merged_sce`, `includes_merged_anndata`.
  - Expected outputs can be found in [Merged Objects Docs](https://github.com/AlexsLemonade/scpca-docs/pull/256/files)
  - Three outputs conditions:
    - `includes_merged_anndata` and `has_CITE`
    - `includes_merged_anndata` and NOT `has_CITE`
    - `includes_merged_sce`

#### Computed Files
- We should be able filter / check if a `ComputedFile` contains merged data
- We should also add a new attribute on `ComputedFile` that indicates if the contents is merged.
  - We do not need to map the existing format as that is already accounted for in `ComputedFile::format`.
  - We can call this `ComputedFile::includes_merged` and make sure that it is added to the serializer.

##### Tests
- point to correct test data version
- ensure serialization is present
- ensure loading of data works with test data

arkid15r:
## Issue Number

Resolves #549 

In addition to the merged data support this PR addresses the recent test data structural changes. 
I also added mocks to `test_data_clean_up` test.

## Types of changes

## Checklist


nozomione:
@davidsmejia I've run pre-commit at 6e09756ce0a0b55854b646035706e45152adb182 (previously also ran the merge migration command for the test, but since getting the duplicate column error and the deploy is still in progress, I reverted it back), thank you!

PR #570, opened 2024-03-04
new and modified ann data helpers

davidsmejia:
## Issue Number

N/A related to ann data changes about to land

## Purpose/Implementation Notes

- adds version of capitalize that only capitalizes the first letter in the string
- adds a picker which is a shorthand for example `a.map(b => b[c])`
- adds helpers for creating drop downs that map to snake case attributes to make them pretty
- adds support for dynamically creating the download contents for computed files
- removes old recommended resource helper based on computedfile.type

Note this is a change that will break tests until the implementations are also updated.
We will be collecting all changes in feature/ann-data before merging to dev.

## Types of changes

- New feature (non-breaking change which adds functionality)

## Functional tests

List out the functional tests you've completed to verify your changes work locally.

## Checklist

- [X] Lint and unit tests pass locally with my changes

## Screenshots

n/a

PR #591, opened 2024-03-13
[AnnData Code Review] Add support for the `undefined` case to the `pick` helper  

nozomione:
## Issue Number
N/A

Merged PR: #570 at commit https://github.com/AlexsLemonade/scpca-portal/pull/570/commits/7d6de88a63246d0becbe08df6c82ee9eda9b489e

I've tweaked the [`/helpers/pick.js`](https://github.com/AlexsLemonade/scpca-portal/blob/dev/client/src/helpers/pick.js) file as follows:

-  Replaced `map` with  `reduce` so that it returns an empty array instead of an array of `undefined` values when the given key is not found in the given array (so that we can easily check for an empty array using `length` etc).
- Added a named export to match the export pattern of the existing helpers. 

Let's create a new issue to bulk update the helper named export with `default export` 👍 

Please see the examples below:

```js
// Before
pick([{ a: 'a', b: 'b'}, {b: 'b', c: 'c'},{c: 'c', d: 'd'}], 'e' ) // [undefined, undefined, undefined], length 3
pick([
    { has_bulk_rna_seq: false, format: "SINGLE_CELL_EXPERIMENT" },
    { has_bulk_rna_seq: false }],
    'format'
) // ['SINGLE_CELL_EXPERIMENT', undefined], length 2

// After
pick([{ a: 'a', b: 'b'}, {b: 'b', c: 'c'},{c: 'c', d: 'd'}], 'e' ) // [], length 0
pick([
    { has_bulk_rna_seq: false, format: "SINGLE_CELL_EXPERIMENT" },
    { has_bulk_rna_seq: false }],
    'format'
) // ['SINGLE_CELL_EXPERIMENT'], length 1
```

## Types of changes

## Checklist

### DIFF
--- a/client/src/contexts/DownloadOptionsContext.js
+++ b/client/src/contexts/DownloadOptionsContext.js
@@ -1,6 +1,6 @@
 import React, { createContext, useState, useEffect, useContext } from 'react'
 import { ScPCAPortalContext } from 'contexts/ScPCAPortalContext'
-import pick from 'helpers/pick'
+import { pick } from 'helpers/pick'
 
 export const DownloadOptionsContext = createContext({})
 
--- a/client/src/helpers/pick.js
+++ b/client/src/helpers/pick.js
@@ -1,1 +1,23 @@
-export default (arr = [], key) => arr.map((i) => i[key])
+/**
+ *  @name pick
+ *  @param {any[]} arr - an array of objects
+ *  @param {string} key - a key to find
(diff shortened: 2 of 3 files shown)


nozomione:
🗒️  Based on 1:1 chat, we don't exclude `undefined` in the returned array at this time. For example, in the future, `undefined` might be used for error handling. For this reason, I am closing this PR.

Issue #561, opened 2024-02-21
Complete Sample::is_xenograft and Sample::is_cell_line handling


davidsmejia:
### Context

We recently added `Project::includes_cell_lines` and `Project::includes_xenografts`. We should add `Sample::is_cell_line`  and `Sample::is_xenograft` to the Sample model. We should round out the implementation by including the other "is_[model]" attribute that we are tracking: `Sample::is_xenograft`. 

### Problem or idea

We should ensure that these attributes are treated the same and presented via the API. 

### Solution or next step

- Add `is_cell_line` boolean attribute to Sample model w/ migrations
- Add `is_xenograft` boolean attribute to Sample model w/ migrations
- Set the value when loading data
- Add `is_xenograft` and `is_cell_line` to sample serializers
- Update tests to check for these values

davidsmejia:
closed with #600

PR #600, opened 2024-03-21
561 - Add is_cell_line and is_xenograft properties to `Sample` model

avrohomgottlieb:
## Issue Number

[#561](https://github.com/AlexsLemonade/scpca-portal/issues/561)

- Added `is_cell_line` and `is_xenograft` as `BooleanFields` to the `Sample` model
- Implemented value setting during loading of data (in `get_from_dict` method)
- Implemented updating of corresponding values in `Project::includes_cell_lines` and `Project::includes_xenograft`
- Added `is_cell_line` and `is_xenograft` to `SampleLeafSerializer`
- Added new test to check for these values, and updated existing tests accordingly

## Types of changes

## Functional tests

- Added a new test inside `test/serializers/test_sample.py` called `test_existence_of_is_cell_line_and_is_xenograft_properties` which confirms that the new keys are reflected in `SampleSerializer`
- In the `test/management/commands/test_load_data.py` module, removed the `is_cell_line` and `is_xenograft` keys from the `expected_additional_metadata_keys` dictionary in the following tests:
  - `test_multiplexed_metadata`
  - `test_single_cell_metadata`
  - `test_spatial_metadata`

## Checklist

## Screenshots

N/A


avrohomgottlieb:
Another approach here would be to delete the `test_existence_of_is_cell_line_and_is_xenograft_properties` test method (which is kind of out of nowhere, no?) and cover it in `test/serializers/test_samples.py` similar to the way that individual fields are tested inside the `test/serializers/test_computed_file.py` test module[^1]. A possible way of going about this would be to add a utility assertContainsFields method (detailed below) and have `test_serializer_with_valid_data` and the `test_detail_serializer_with_valid_data` call it.

```python
    def assertContainsFields(self, serializer):
        for field in ("is_cell_line", "is_xenograft"):
            self.assertIn(field, serializer.data)
```

Another option could be to not test the existence of these individual properties inside `test_samples.py` if there's enough coverage in the `test/management/commands/test_load_data` test module (I've mentioned my edits to that test module as it pertains to this PR in the original PR message).

[^1]: See lines 12-14, 27 and 32 there

davidsmejia (commented):
This is looking good, I need to double check where we would expect to find that value and I will comment here.

davidsmejia (approved):
LGTM

Issue #367, opened 2023-02-09
Improve load data script flags handling


arkid15r:
### Problem or idea

`BooleanOptionalAction` provides better options for args/flags handling

### Solution or next step

Use `argparse.BooleanOptionalAction` for existing boolean flags of `load_data` command

avrohomgottlieb:
### Context
**Relevant arguments of type `Boolean`**:
```
--clean-up-input-data
--clean-up-output-data
--reload-all
--reload-existing
--skip-sync
--update-s3
```

### Discussion 
From a *debug* perspective, it seems like best practice would be to make sure that all arguments of type `Boolean` receive a valid value, as opposed to defaulting to `None`. 

The three arguments whose default values change between **production** and **local** environments,
```
--clean-up-input-data
--clean-up-output-data
--update-s3
```
benefit from using `argparse.BooleanOptionAction` because it allows for explicit option passing when the default values change.

However, it seems to me that the other three arguments
```
--reload-all
--reload-existing
--skip-sync
```
should default to `False` and keep the `store_true` action in order to prevent any unintended consequences of defaulting to `None`.

Happy to discuss further or to open a PR with my proposed changes, this is just my line of thinking.

PR #601, opened 2024-03-21
367 - Improve load data script flags handling

avrohomgottlieb:
## Issue Number

[#367](https://github.com/AlexsLemonade/scpca-portal/issues/367)

- Added default value of `False` to `--reload-all` and `--reload-existing` flags, in order that they should default to the anticipated `False` value, and not to type `None`. 
> Aside: This is also meant to clarify the convention for the `load_data` script flags of type `Boolean`, where all non environment dependent[^1] default values fall back on `False`, while the caller is given the ability to explicitly override this functionality when desired by passing the flag to the command. With those flags that do have environment dependent default values, we use the `argparse.BooleanOptionalAction` action, which creates a unified interface for both `--arg` and `--no-arg` to be handled by the command, irrespective of the environment's default value.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

[^1]: *Environment dependent* default values are flags who's default value change between development and production environments

### DIFF
--- a/api/scpca_portal/management/commands/load_data.py
+++ b/api/scpca_portal/management/commands/load_data.py
@@ -123,8 +123,8 @@ class Command(BaseCommand):
             "--clean-up-output-data", action=BooleanOptionalAction, default=settings.PRODUCTION
         )
         parser.add_argument("--max-workers", type=int, default=10)
-        parser.add_argument("--reload-all", action="store_true")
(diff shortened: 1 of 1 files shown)


davidsmejia (approved):
LGTM!

avrohomgottlieb:
@davidsmejia anything left to do before merging this in?

davidsmejia:
This is good to merge!

PR #609, opened 2024-04-04
Use ANNDATA readme for single-cell sample when format is ANNDATA

davidsmejia:
## Issue Number

N/A

Per slack:
> It also looks like the readme being packaged with AnnData downloads is still the one for SingleCellExperiment objects.

This is a small change that sets the source template for the readme to be the `ComputedFile.README_ANNDATA_FILE_PATH` when a single-cell sample's computed file is being zipped up. This was correctly set for project but not for sample.

## Types of changes

## Functional tests

n/a

## Checklist

## Screenshots

N/A


avrohomgottlieb (commented):
Yes, this is spot on. It must have not been updated when AnnData was added in for Single Cell. This is good to go.

avrohomgottlieb (approved):
Yes, this is spot on. It must have not been updated when AnnData was added in for Single Cell. This is good to go.

Issue #612, opened 2024-04-05
Escape backticks in markdown files


nozomione:
### Context
For Gromment's built-in `Markdown` component, we use the template literal loader to generate template literals from markdown files during build time. Backticks in those markdown files may results in a "SyntaxError" error which need to be prevented.

### Problem or idea
In the loader, we should parse the contents of markdown and escape backticks before evaluating them to prevent build errors.

### Solution or next step
- Add backslashes to backticks present in the markdown files before the eval in the template literal loader.
- Test the solution locally to make sure there are no build errors.

PR #633, opened 2024-04-08
612 - Escape backticks in markdown files

nozomione:
## Issue Number

Closing #612

I've add backslashes to backticks present in the markdown files before the eval in the template literal loader.

## Types of changes

## Functional tests

I tested the solution locally to make sure there are no build errors.

## Checklist

## Screenshots

### DIFF
--- a/client/loaders/template-literal-loader.js
+++ b/client/loaders/template-literal-loader.js
@@ -2,6 +2,9 @@
 // markdown to be consumed by grommet/Markdown
 
 module.exports = (source) => {
+  // escape backticks to prevent build errors
+  const formatted = source.replaceAll('`', '\\`')
+
   // eslint-disable-next-line
-  return eval(`\`${source}\``)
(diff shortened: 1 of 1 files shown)


client/loaders/template-literal-loader.js
```diff
@@ -2,6 +2,9 @@
 // markdown to be consumed by grommet/Markdown
 
 module.exports = (source) => {
+  // escape backticks to prevent build errors
```
davidsmejia, line 5:
```suggestion
  // escape backticks to prevent eval errors
```

davidsmejia (changes requested):
This regex is matching on the preceding character on the closing tick.
We want to be able to have non-escaped backticks and escaped backticks in `source` and only match on unescaped.

Example of the current issue:
```
this is `an example`
```
will produce
```
this is \`an exampl\`
```

PR #646, opened 2024-04-11
Fix multiplexed sample computed file generation when using multiple threads

davidsmejia:
## Issue Number

N/A

Main changes:
- Prevent deleting a multiplexed sample after one thread had created it so that it can be used when specifying to clean up output data, we clean up zips after created and every other file after the load_data script has completed. This is a problem for multiplexed because we re-use those files when processing the `multiplexed_with` samples.
- Make the checking of if a computed file exists thread safe. When creating multiplexed samples we reuse a computed file if it is already there. I added a change that creates a lock for each group of samples that are multiplexed together. This prevents a subsequent thread from trying to interpret the state of the generated zip before it is finished being written to.

Other notes:
- Our current implementation of concurrent processing of samples doesn't really scale. Writing to the disk is blocking regardless of how many threads you throw at it. I think what we should be handling concurrently is the generation of metadata as that would all be in memory. The writing of files can still be multithreaded but there is little benefit to that.
- I noticed that writing large files to a zip with `zipfile.ZipFile.write` is preposterously slow. This is particularly true when we create project zips. We should consider using a different module or call a sub-process directly.

ride along changes:
- Formatting

## Types of changes

## Functional tests

N/A Processed multiplexed project locally

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/models/project.py
+++ b/api/scpca_portal/models/project.py
@@ -295,7 +295,11 @@ class Project(CommonDataAttributes, TimestampedModel):
         if update_s3:
             logger.info(f"Uploading {computed_file}")
             computed_file.create_s3_file()
-        if clean_up_output_data:
+
+        # Don't clean up multiplexed sample zips until the project is done
+        is_multiplexed_sample = computed_file.sample and computed_file.sample.has_multiplexed_data
+
+        if clean_up_output_data and not is_multiplexed_sample:
             computed_file.zip_file_path.unlink(missing_ok=True)
 
         # Close DB connection for each thread.
--- a/api/scpca_portal/models/project.py
+++ b/api/scpca_portal/models/project.py
@@ -4,6 +4,7 @@ import logging
 from collections import Counter
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
+from threading import Lock
 from typing import Dict, List
 
 from django.contrib.postgres.fields import ArrayField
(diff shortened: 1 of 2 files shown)


api/scpca_portal/models/project.py
```diff
@@ -1039,6 +1044,14 @@ def update_spatial_data(future):
             f"Processing {samples_count} sample{pluralize(samples_count)} using "
             f"{max_workers} worker{pluralize(max_workers)}"
         )
+
+        # Prepare a threading.Lock for each multiplexed sample that shares a zip file.
+        # The keys are the sample.multiplexed_ids since that will be unique across shared zip files.
+        multiplexed_ids = set(
+            ["_".join(s.multiplexed_ids) for s in samples if s.has_multiplexed_data]
```
avrohomgottlieb, line 1051: Before performing the join, we should probably sort all of `s.multiplexed_ids` to guarantee the uniqueness of the set. It could be that the ids are sorted elsewhere, but we could remove the dependency by sorting here again (it's only a few ids in any case).

avrohomgottlieb:
In `project.py` the lock is imported from the `threading` library, while in `computed_file.py` its import from the `multiprocessing` library. Shouldn't they both be from `threading`?

api/scpca_portal/models/project.py
davidsmejia: These are already sorted so we should be good. The set is just do we don't create multiple keys.

avrohomgottlieb (approved):
LGTM

Issue #645, opened 2024-04-10
Move single cell and spatial library metadata handling in `Project::load_data` to their own methods


avrohomgottlieb:
### Context

The beginning of `Project::load_data` deals with metadata handling (parsing, extracting, collecting, combining, etc). At first, the general `samples_metadata.csv` file is parsed, followed by the parsing and collecting of library metadata in a `<modality>.json` file (modalities being single_cell, spatial, and multiplexed).

### Problem or idea

From a modularity perspective, there is room to say that having the handling of single cell and spatial library metadata inside of `Project::load_data` contributes to code bloat within the method.

### Solution or next step

The solution is to move the handling of single cell and spatial library metadata out of `Project::load_data` and into a generic method that can handle them both.

Additionally, they both update a variety of similar data structures within their execution, which should be passed by reference and handled within the method.

The stub of the method would look as follows:
```python
def parse_library_metadata(
    self,
    modality: str,
    modality_libraries_metadata: List,
    has_modality_data: Dict,
    sample_dir: Path,
    sample_seq_units: Set,
    sample_technologies: Set,
    sample_cell_count_estimates: Optional[List] = None
) -> None:
```

avrohomgottlieb:
Among the data structures touched within this method is an accumulator `int` variable, called `sample_cell_count_estimates`, which is currently passed by value. In order to pass this state through to the new method and get around the inability to pass a variable by reference in python, this  variable should be turned into a list which can be appended to during the new method's execution and summed up later on. The fact that this list will only hold a few values makes the performance price negligible and will contribute to cleaner code.

davidsmejia:
Closed with #652

PR #647, opened 2024-04-11
Add generic parse_library_metadata method which handles parsing of single cell and spatial library metadata

avrohomgottlieb:
## Issue Number

Resolves https://github.com/AlexsLemonade/scpca-portal/issues/645

Created a generic function which handles parsing single cell and spatial json library metadata files. This was a simple refactoring of a piece of code that could be moved elsewhere as we work to thin out `Project::load_data`.

## Types of changes

  - Non-breaker refactor of two methods

## Checklist

### DIFF
--- a/api/scpca_portal/models/project.py
+++ b/api/scpca_portal/models/project.py
@@ -4,7 +4,7 @@ import logging
 from collections import Counter
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
-from typing import Dict, List
+from typing import Dict, List, Set
 
 from django.contrib.postgres.fields import ArrayField
 from django.db import connection, models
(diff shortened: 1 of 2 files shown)


avrohomgottlieb:
Instead of using a single generic function which updates data structures passed by reference, the desired approach is to break this up into pure functions which take in values and then return something of substance. Commit `5319e04` better represents this approach. More to come.

avrohomgottlieb:
This PR was tossed after a better approach to refactoring was decided upon. The first PR in the better approach can be found at here: [Handle Single Cell and Spatial Metadata Parsing, and Sample-Library Combining #652](https://github.com/AlexsLemonade/scpca-portal/pull/652)

Issue #635, opened 2024-04-09
Move Metadata Extraction Logic From `load_data` Method in `Project` Model to its Own Method


avrohomgottlieb:
### Context

The `load_data` method inside the `Project` model accomplishes a variety of different tasks. Among those tasks includes generating readmes, combining metadata, and creating computed files.

Metadata is received in the form of three files: `project_metadata.csv`, `samples_metadata.csv` and within each sample (/ library) a `SCPCL<library_number>_metadata.json` file. Inside the `load_data` method, each of these files is parsed, the metadata is collected and combined at the sample level, and an accumulated `single_cell_metadata.tsv` file is created and outputted (or `spatial_metadata` if the project/sample is of the *spatial* modality).

### Problem or idea

The metadata extraction logic is too tightly coupled with the other tasks within the `load_data` method.

### Solution or next step

The solution is to bundle together all of the logic associated with metadata extraction within the`load_data` method, and move it to its own method within the `Project` model.

davidsmejia:
This sounds right. I think we should limit this issue to handling these lines. And just extracting the nested logic into multiple functions.
- https://github.com/AlexsLemonade/scpca-portal/blob/dev/api/scpca_portal/models/project.py#L859-L992


I think we should just write another issue for handling moving the metadata processing into another function.
- https://github.com/AlexsLemonade/scpca-portal/blob/dev/api/scpca_portal/models/project.py#L807-L848

avrohomgottlieb:
I agree with you on this. It makes sense to leave `get_samples` for another issue.

I think the following lines should also be included in the current issue:
- https://github.com/AlexsLemonade/scpca-portal/blob/dev/api/scpca_portal/models/project.py#L1116-L1138

avrohomgottlieb:
@davidsmejia With single-cell data we receive the three metadata files listed above and output a `single_cell_metadata.tsv` file, but with spatial data we output both the combined `spatial_metadata.tsv` file and the input library json file (unmodified). Is that by design or a bug?

davidsmejia:
This was addressed in #652 

Will move last question about spatial to another issue.

PR #652, opened 2024-04-16
Handle Single Cell and Spatial Metadata Parsing, and Sample-Library Combining

avrohomgottlieb:
## Issue Number

#645 (main one)
#635 (secondary one)

This PR is the first step in the refactoring of the metadata handling logic in the `Project::load_data` method. For the most part the logic itself was untouched but rather moved around and abstracted into generic sub-methods with a single `handle_samples_metadata` now being callable from within `load_data` itself. The next steps will be to tuck multiplex logic into `handle_samples_metadata`, while removing all of the file writing functionality from `handle_samples_metadata` to its own method.

## Types of changes

- Refactor (non-breaking change which does not add functionality but improves code organization and design)

## Functional tests

No tests added or modified (inputs and outputs are the same, just the implementation was refactored, so tests remained the same).

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/models/project.py
+++ b/api/scpca_portal/models/project.py
@@ -1176,13 +1176,10 @@ class Project(CommonDataAttributes, TimestampedModel):
                 sample_seq_units.add(library_json["seq_unit"].strip())
                 sample_technologies.add(library_json["technology"].strip())
 
-                libraries_metadata[Sample.Modalities.SINGLE_CELL].append(
-                    library_json
-                ) if "spatial" not in str(filename_path) else libraries_metadata[
-                    Sample.Modalities.SPATIAL
-                ].append(
-                    library_json
-                )
+                if "spatial" in str(filename_path):
+                    libraries_metadata[Sample.Modalities.SPATIAL].append(library_json)
+                else:
+                    libraries_metadata[Sample.Modalities.SINGLE_CELL].append(library_json)
 
             updated_sample_metadata["sample_cell_count_estimate"] = sample_cell_count_estimate
             updated_sample_metadata["seq_units"] = ", ".join(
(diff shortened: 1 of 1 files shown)


api/scpca_portal/models/project.py
```diff
@@ -1155,6 +1101,171 @@ def update_spatial_data(future):
...
+                sample_seq_units.add(library_json["seq_unit"].strip())
+                sample_technologies.add(library_json["technology"].strip())
+
+                libraries_metadata[Sample.Modalities.SINGLE_CELL].append(
+                    library_json
+                ) if "spatial" not in str(filename_path) else libraries_metadata[
+                    Sample.Modalities.SPATIAL
+                ].append(
+                    library_json
+                )
```
davidsmejia, line 1185: Could we rewrite this for clarity?

api/scpca_portal/models/project.py
```diff
@@ -1155,6 +1101,171 @@ def update_spatial_data(future):
...
+            updated_sample_metadata["technologies"] = ", ".join(
+                sorted(sample_technologies, key=str.lower)
+            )
+
+        return (updated_samples_metadata, libraries_metadata)
+
+    def combine_metadata(self, updated_samples_metadata, libraries_metadata, sample_id):
+        combined_metadata = {Sample.Modalities.SINGLE_CELL: [], Sample.Modalities.SPATIAL: []}
+
+        for modality in [Sample.Modalities.SINGLE_CELL, Sample.Modalities.SPATIAL]:
```
davidsmejia, line 1200:
```suggestion
        for modality in combined_metadata.keys():
```

davidsmejia (approved):
This looks good, made a couple suggestions.

api/scpca_portal/models/project.py
avrohomgottlieb: Good call. I think I might have been auto-formatted here. In any case I think a normal `if else` makes more sense here then the python ternary.

PR #653, opened 2024-04-16
Add merged data project downloads support

davidsmejia:
## Issue Number

#549 

Includes cherry-picked commits from first implementation in `ark/549-add-merged-objects-support`

- Adds support for generating merged project computed files
- updates `test_load_data` to use version in `2024-03-08`
- adds some breakdowns for explaining why certain number of samples or computed files are expected. I am going to create an issue for handling this because I think this should be located somewhere else.

Additional Changes since initial PR:
- Merged projects now use correct readme templates
- Moved merged readme templates to `templates/readme` where they should be
- Removed unused additional restrictions template while i was cleaning that up

## Types of changes

## Functional tests

N/A ran tests locally

## Checklist

## Screenshots

N/A


api/scpca_portal/models/computed_file.py
```diff
@@ -110,6 +111,83 @@ def __str__(self):
...
+            project.input_merged_summary_report_file_path: (
+                f"{project.scpca_id}_merged-summary-report.html"
+            ),
+        }
+
+        if file_format == cls.OutputFileFormats.ANN_DATA:
+            if not project.includes_merged_anndata:
+                return None
+            computed_file_name = project.output_merged_anndata_computed_file_name
+            readme_file_path = ComputedFile.README_ANNDATA_FILE_PATH
```
davidsmejia, line 130: Does this need to be the merged path? Same for SCE

api/scpca_portal/models/project.py
```diff
@@ -131,6 +149,18 @@ def single_cell_computed_file(self):
         try:
             return self.project_computed_files.get(
                 format=ComputedFile.OutputFileFormats.SINGLE_CELL_EXPERIMENT,
+                includes_merged=False,
+                type=ComputedFile.OutputFileTypes.PROJECT_ZIP,
```
davidsmejia, line 153: Remove type and change to check for `project_id`

davidsmejia, line 153:
```suggestion
                project_id__isnull=False,
```

api/scpca_portal/models/project.py
```diff
@@ -141,6 +171,18 @@ def single_cell_anndata_computed_file(self):
...
+            )
+        except ComputedFile.DoesNotExist:
+            pass
+
+    @property
+    def single_cell_anndata_merged_computed_file(self):
+        try:
+            return self.project_computed_files.get(
+                format=ComputedFile.OutputFileFormats.ANN_DATA,
+                includes_merged=True,
```
davidsmejia, line 185: Also remove `type` here for `project_id__isnull=False`

api/scpca_portal/models/project.py
```diff
@@ -599,21 +641,30 @@ def create_computed_files(
...
-            self.process_computed_file(computed_file, clean_up_output_data, update_s3)
+            if computed_file:
+                self.process_computed_file(computed_file, clean_up_output_data, update_s3)
 
         with ThreadPoolExecutor(max_workers=max_workers) as tasks:
             for file_format in (
                 ComputedFile.OutputFileFormats.ANN_DATA,
                 ComputedFile.OutputFileFormats.SINGLE_CELL_EXPERIMENT,
             ):
+                tasks.submit(
```
davidsmejia, line 660: Check if format is in the file mapping first? Then remove the check on `if computed_file` in `create_computed_file`

api/scpca_portal/test/management/commands/test_load_data.py
```diff
@@ -68,18 +63,27 @@ def test_data_clean_up(self):
...
     def test_load_data(self):
+        project_id = "SCPCP999990"
+
         def assert_object_count():
             self.assertEqual(Project.objects.count(), 1)
-            self.assertEqual(ProjectSummary.objects.count(), 5)
-            self.assertEqual(Sample.objects.count(), 5)
-            self.assertEqual(ComputedFile.objects.count(), 9)
+            self.assertEqual(ProjectSummary.objects.count(), 4)
+            # There are 3 Single-cell Samples and 1 Sample that is available in Bulk RNA only.
```
davidsmejia, line 79: 2 single cell, 1 bulk, 1 spatial

api/scpca_portal/test/management/commands/test_load_data.py
```diff
@@ -139,12 +145,194 @@ def assert_object_count():
...
+            # │   │   └── SCPCL999990_celltype-report.html
+            # │   └── SCPCS999997
+            # │       └── SCPCL999997_qc.html
+            # │       └── SCPCL999997_celltype-report.html
+            # └── single_cell_metadata.tsv
+            files = set(project_zip.namelist())
+            self.assertEqual(len(files), 10)
+            self.assertIn("SCPCP999990_merged_rna.hdf5", files)
+
+    def test_no_merged_single_cell(self):
```
davidsmejia, line 307: this is actually no merged single-cell and no-merged anndata

api/scpca_portal/management/commands/load_data.py
```diff
@@ -97,6 +97,7 @@ def download_data(bucket_name, scpca_project_id=None, scpca_sample_id=None):
                 (
                     "--exclude=*",  # Must precede include patterns.
                     "--include=project_metadata.csv",
+                    f"--include=merged/{scpca_project_id}*",
```
avrohomgottlieb, line 100: Doesn't merged data only apply at the project level? Here it's downloading the project when triggered from the sample level.

api/scpca_portal/models/project.py
```diff
@@ -77,6 +79,10 @@ def computed_files(self):
     def input_data_path(self):
         return common.INPUT_DATA_PATH / self.scpca_id
 
+    @property
+    def input_merged_data_path(self):
+        return self.input_data_path / "merged"
```
avrohomgottlieb, line 84: return self.input_data_path / f'{self.scpca_id}_merged.rds'?

avrohomgottlieb (approved):
LGTM

api/scpca_portal/models/project.py
davidsmejia: This gets used on line 95, so it is just used to break up where we source the "merged" files, meaning celltype-report and the actual data in the input folder.

api/scpca_portal/management/commands/load_data.py
davidsmejia: Yeah good catch, I think that we can leave this for now. Issue that I have is that when recreating a sample we would need to update the project anyway. So honestly, at this time we never use this flag and I'd like to remove it in it's entirety anyway. I haven't even tested processing of a single sample because of the reason just stated.

davidsmejia:
Additional Changes since initial PR:
- Merged projects now use correct readme templates
- Moved merged readme templates to `templates/readme` where they should be
- Removed unused additional restrictions template while i was cleaning that up

api/scpca_portal/management/commands/load_data.py
avrohomgottlieb: Right, I've been wondering the same thing about the command to download individual samples for that exact reason. Sounds good.

Issue #663, opened 2024-04-19
Move multiplexed metadata handling to `Project::handle_samples_metadata`


avrohomgottlieb:
### Context

Now that single cell and spatial metadata handling has been moved to the `Project::handle_samples_metadata` method ([PR #652](https://github.com/AlexsLemonade/scpca-portal/pull/652)), multiplexed metadata handling should be moved there as well. 

Multiplexed metadata has certain nuances as to how its parsed and how its sample and library metadata are combined, taking into account the individual demux sample ids within the multiplexed sample. Because of this it was important to separate multiplexed metadata handling and single cell/spatial metadata handling into multiple PRs.

The multiplexed metadata parsing is found in the beginning of `Project::load_data`, the the combining of sample and library dictionaries is in `Project::combine_multiplexed_metadata`. As well, other fields are updated within `Project::get_samples` (which have no businesses being there!).

### Problem or idea

As mentioned above, the handling of multiplexed metadata should be moved out of `Project::load_data`, `Project::combine_multiplexed_metadata` and elsewhere into `Project::handle_samples_metadata`.

### Solution or next step

The multiplexed metadata parsing `Project::load_data` is very similar to that of single cell and spatial, with a few exceptions. Among them are the ways that the fields `sample_cell_estimates`, `seq_units`, and `technology` is handled. This would have to be taken into consideration when adding these fields into `Project::load_libraries_metadata` which is where similar logic is handled for single cell and spatial.

For the logic inside `Project::combine_multiplexed_metadata`, the idea would be to break up the three different sections of that method, to see what can be abstracted out and recomputed in their own methods, and what could be moved to the general `Project::combine_metadata`.

avrohomgottlieb:
Issue closed with successful merging of https://github.com/AlexsLemonade/scpca-portal/pull/666

Issue #638, opened 2024-04-09
Move updating of samples logic in `Project::get_samples` to sample metadata handling


avrohomgottlieb:
### Context

Inside `Project::load_data`, once metadata has been parsed, collected and combined, `Project::get_samples` is called to create `Sample` objects which will soon be saved to the database. 

`Project::get_samples` also edits the following 4 fields on each sample:
- `demux_cell_count_estimate`
- `multiplexed_with`
- `seq_units`
- `technologies`

### Problem or idea

There's no reason why these 4 sample fields should be edited here and not when the other fields are being edited earlier on.

Additionally, the naming isn't clear - samples are not retrieved here, as the name `get_samples` seems to be indicate, but rather objects are created in memory in order to be saved soon to the database.

### Solution or next step

1) Extract all field editing logic from this method and handle it in the normal flow of metadata parsing (or move into its own method).
2) Rename method to `generate_sample_objects` or `create_sample_objects`.

avrohomgottlieb:
Another point connected to this issue is that after the sample objects are created, they should be immediately saved to the database. 

Currently these sample objects are stored in a list and are saved to the db during sample file computation, as follows [^1]:
```python
for sample in Sample.objects.bulk_create(samples):
```

This seems like needless coupling. Better would be to save the sample objects when they're created, and then to query for them from the database when the time comes to begin file computation, as follows:
```python
for sample in Sample.objects.filter(project__scpca_id=kwargs["scpca_project_id"])`
```

[^1]: https://github.com/AlexsLemonade/scpca-portal/blob/dev/api/scpca_portal/models/project.py#L1043

avrohomgottlieb:
Should I wait until the multiplex refactor before tackling this one?

avrohomgottlieb:
Issue addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/697

PR #666, opened 2024-04-22
Project.load_data Handle multiplex metadata logic

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/663

Following up from https://github.com/AlexsLemonade/scpca-portal/pull/652, (which dealt with single-cell and spatial metadata), this PR addresses the handling of multiplexed metadata, specifically:
- updating its sample and library metadata loading inside of `Project::load_samples_metadata` and `Project::load_libraries_metadata` respectively
- integrating `Project::combine_multiplexed_metadata` into the generic `Project::combine_metadata` method

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


avrohomgottlieb:
Bug alert: in `get_library_metadata_keys`, modality/modalities kwarg don't agree between usage and implementation

avrohomgottlieb:
Still to do:
- Adding of docstrings to newly added methods

Thinking about doing:
- More aggressive refactor of `Project::combine_metadata` to move out the key generation at the beginning of the method to its own method. Further down the line it would also be good to look at the `Project::get_sample_metadata_keys` and `Project::get_library_metadata_keys` methods to see if they can be rewritten in a way that's more declarative, and to update the usage of the modality/modalities tuple.
- Tucking of `Project::load_sample_metadata` into  `Project::load_library_metadata` in order to manage a single state of the `samples_metadata`->`updated_samples_metadata` dict.

avrohomgottlieb:
> Bug alert: in `get_library_metadata_keys`, modality/modalities kwarg don't agree between usage and implementation. This is connected to the adding of `demux_cell_count_estimate` and `multiplexed_with` keys in `get_sample_metadata_keys` inside the general `excluded_keys` dict, as opposed to within the multiplexed block. When put within the multiplexed block, errors arise in test suite.

avrohomgottlieb:
This PR also contributes to https://github.com/AlexsLemonade/scpca-portal/issues/638 by removing the mutations of the four multiplexed library metadata fields from `Project::get_samples`, and by moving them to `Project::load_libraries_metadata`.

api/scpca_portal/models/project.py
```diff
@@ -745,6 +463,166 @@ def get_additional_terms(self):
...
+                    multiplexed_json = json.load(multiplexed_json_file)
+
+                multiplexed_json["scpca_library_id"] = multiplexed_json.pop("library_id")
+                multiplexed_json["scpca_sample_id"] = multiplexed_json.pop("sample_id")
+
+                multiplexed_libraries_metadata.append(multiplexed_json)
+
+        return multiplexed_libraries_metadata
+
+    def get_multiplexed_remaining_fields(self):
```
davidsmejia, line 493: Any reason not to call these aggregate fields?

api/scpca_portal/models/project.py
```diff
@@ -745,6 +463,166 @@ def get_additional_terms(self):
...
+    ):
+        """
+        Returns a list of combined metadata of all samples that
+        the scpca_sample_id sample was multiplexed with.
+        The returned list is then immediately written to the open csv file in the caller method.
+        """
+        multiplexed_with_combined_metadata = []
+        multiplexed_sample_ids = sorted(multiplexed_with_mapping[scpca_sample_id])
+
+        multiplexed_with_samples_metadata_filtered_keys = {
```
davidsmejia, line 590: This name is pretty hard to parse without looking at how it is used. I think an inline comment describing the variable's value would be prudent here.

api/scpca_portal/models/project.py
```diff
@@ -750,12 +752,31 @@ def get_metadata_field_names(self, columns, modality):
...
+        Returns a dictionary which maps sample ids to a set of associated library ids.
+        """
+        # sample_libraries_id_mapping = {sample_id: set() for sample_id in sample_ids}
+        sample_libraries_id_mapping = {}
+
         for library_metadata in libraries_metadata:
-            sample_libraries_id_mapping[library_metadata["scpca_sample_id"]].add(
-                library_metadata["scpca_library_id"]
-            )
+            isMultiplexed = "demux_samples" in library_metadata
```
davidsmejia, line 763: `is_multiplexed`

api/scpca_portal/models/project.py
```diff
@@ -750,12 +752,31 @@ def get_metadata_field_names(self, columns, modality):
...
+
+            if not isMultiplexed:
```
davidsmejia, line 765: can we reverse the order of these cases?

api/scpca_portal/models/project.py
```diff
@@ -1213,25 +1213,22 @@ def combine_metadata(self, updated_samples_metadata, libraries_metadata, sample_
                 modality=modality,
             )
 
-            sample_libraries_mapping = self.get_sample_libraries_mapping(
-                libraries_metadata[modality]
+            pre_filtered_samples_metadata = (
```
davidsmejia, line 1216: Would you be alright with `unfiltered_samples_metadata` or `input_samples_metadata`

davidsmejia (commented):
I'd like to take another pass at our naming before merging but this is looking pretty good.

api/scpca_portal/models/sample.py
```diff
@@ -56,25 +56,24 @@ def __str__(self):
     def get_from_dict(cls, data, project):
         """Prepares ready for saving sample object."""
 
-        has_multiplexed_data = bool(data.get("multiplexed_with"))
         sample = cls(
             age_at_diagnosis=data["age_at_diagnosis"],
-            demux_cell_count_estimate=(
-                data.get("demux_cell_count_estimate") if has_multiplexed_data else None
-            ),
+            demux_cell_count_estimate=(data.get("demux_cell_count_estimate", None)),
```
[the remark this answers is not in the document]
avrohomgottlieb: I believe so. The key is dropped because `demux_cell_count_estimate` is not included in the output metadata file, but we keep the property as an attribute on `Sample` objects. 
- When building `Sample` objects we use `updated_samples_metadata`.
- When combining metadata and writing the metadata files, we filter `updated_samples_metadata` and use  `samples_metadata_filtered_keys`.

api/scpca_portal/models/project.py
avrohomgottlieb: Good call, that's a better name. Will make the change

api/scpca_portal/models/project.py
avrohomgottlieb: My reasoning was that the normal case is *not* multiplexed, and the exception *is* multiplexed. Personal preference - I tend to prefer the rule first, and then the exception afterwards. Does the following help at all?
```python
is_not_multiplexed = "demux_samples" not in library_metadata

if is_not_multiplexed:
  ...
```

api/scpca_portal/models/project.py
avrohomgottlieb: Sure. I actually originally used `input_samples_metadata`, but was afraid that the name was too generic and that it could be confused with the different usage of `input_samples_metadata` used elsewhere throughout in the file. I like `unfiltered_samples_metadata`.

api/scpca_portal/utils.py
```diff
@@ -28,3 +28,15 @@ def join_workflow_versions(workflow_versions):
 def get_today_string(format: str = "%Y-%m-%d"):
     """Returns today's date formatted. Defaults to ISO 8601."""
     return datetime.today().strftime(format)
+
+
+def filter_dict_list_by_keys(list_of_dicts, included_keys):
```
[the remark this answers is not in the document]
avrohomgottlieb: Yes, we should.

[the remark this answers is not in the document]
avrohomgottlieb: Added.

davidsmejia (commented):
Looking good, just a couple suggestions.

api/scpca_portal/utils.py
```diff
@@ -28,3 +28,24 @@ def join_workflow_versions(workflow_versions):
 def get_today_string(format: str = "%Y-%m-%d"):
     """Returns today's date formatted. Defaults to ISO 8601."""
     return datetime.today().strftime(format)
+
+
+def filter_dict_list_by_keys(list_of_dicts, included_keys, ignore_value_error=False):
```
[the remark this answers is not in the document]
avrohomgottlieb: Ay ay captain!

api/scpca_portal/utils.py
```diff
@@ -28,3 +28,24 @@ def join_workflow_versions(workflow_versions):
...
+    """
+    Returns a list of dictionaries with keys filtered according to provided include_keys.
+    If included-keys are not a subset of dictionary keys, returns a ValueError.
+    Throwing of ValueError can be disabled.
+    """
+    # Make sure included_keys is subset of dictionary keys
+    if not ignore_value_error and not set(included_keys).issubset(set(list_of_dicts[0].keys())):
```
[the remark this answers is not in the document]
avrohomgottlieb: This is a good point, I was thinking about it from the perspective of the current usage where all inputted sample metadata dictionaries have the same set of keys, but a utils function should have the expectation of being generic and having the ability to accept varied inputs.

Here's a better solution:
```python
if not ignore_value_error:
    if any(not set(included_keys).issubset(dictionary) for dictionary in list_of_dicts):
        raise ValueError("Included keys must be a subset of each dictionary's key set")
```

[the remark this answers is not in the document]
davidsmejia: Yeah this seems more reasonable.

davidsmejia (approved):
👍

PR #683, opened 2024-04-26
Allow downloading of multiple configuration options

davidsmejia:
## Issue Number

N/A

When selecting `Download Now` again (without refreshing) the previously configured download would automatically trigger. This was not the expected behavior. We now clear out the public computed file after downloading so that users can select new download options without refreshing.

- Conditionally set `publicComputedFile` to null when not specified on modal open
- Loosen constraint that `publicComputedFile` be defined to show `DownloadStarted` state.
  - This should be fine because `download` must be true which implies `publicComputedFile` was correctly defined at some point in the hooks's lifecycle

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/client/src/hooks/useDownloadModal.js
+++ b/client/src/hooks/useDownloadModal.js
@@ -22,7 +22,7 @@ export const useDownloadModal = (
   const hasMultipleFiles = hasMultiple(resource.computed_files)
 
   // states that dictate what the modal can show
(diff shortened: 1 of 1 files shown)


dvenprasad (approved):
🐗 🐗 🐗

PR #684, opened 2024-04-26
Production Deploy

davidsmejia:
# Production Deploy
### Client
- After downloading a project and subsequently selecting the same project `Download Now` button we no longer automatically trigger the previously configured download and present the download options form again without requiring a refresh.


dvenprasad (approved):
🚀 🚀 🚀

PR #687, opened 2024-04-30
Production Deploy

davidsmejia:
# Production Deploy
## Change log
- Bug fix: Previously generated `Single-cell` `SingleCellExperiment` computed file zips that contained multiplexed samples were missing sample library metadata for non-multiplexed samples. This has been addressed and all library metadata is available for non-multiplexed and multiplexed samples alike.


allyhawkins (approved):
Based on the metadata on staging, this looks good 👍

PR #697, opened 2024-05-09
Update Sample Object Saving Logic

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/638

This PR finishes the work started in PR https://github.com/AlexsLemonade/scpca-portal/pull/666, which cleaned up `Project::get_samples` by extracting the updating of the four multiplexed metadata aggregate properties from the method, and moving them to `Project::load_libraries_metadata`, inside of `Project::handle_samples_metadata`. 

This PR adds saving logic to the sample objects created in `get_samples`, while updating the method name accordingly to `create_samples`. The references to these samples during sample file computation are addressed by querying and retrieving them from the db as needed.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/models/project.py
+++ b/api/scpca_portal/models/project.py
@@ -843,7 +843,7 @@ class Project(CommonDataAttributes, TimestampedModel):
         """Returns an input data directory based on a sample ID."""
         return self.input_data_path / sample_scpca_id
 
-    def get_samples(
+    def create_samples(
         self,
         samples_metadata,
         sample_id=None,
@@ -857,7 +857,7 @@ class Project(CommonDataAttributes, TimestampedModel):
 
             samples.append(Sample.get_from_dict(sample_metadata, self))
 
-        return samples
+        Sample.objects.bulk_create(samples)
 
     def load_data(self, sample_id=None, **kwargs) -> None:
         """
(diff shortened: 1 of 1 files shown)


api/scpca_portal/models/project.py
```diff
@@ -843,7 +843,7 @@ def get_sample_input_data_dir(self, sample_scpca_id):
         """Returns an input data directory based on a sample ID."""
         return self.input_data_path / sample_scpca_id
 
-    def get_samples(
+    def create_samples(
```
davidsmejia, line 846: While we are here we should do the following:
- this seems like it would be better served as a class method on the Sample model
- I think we can improve the docstring on this method
- since this function essentially just wraps the `get_from_dict` and `bulk_create` methods, we should probably rename this to `bulk_create_from_dicts`

api/scpca_portal/models/project.py
```diff
@@ -925,6 +925,12 @@ def update_spatial_data(future):
 
         non_downloadable_sample_ids = self.get_non_downloadable_sample_ids()
         max_workers = kwargs["max_workers"]
+
+        samples = (
+            Sample.objects.filter(project__scpca_id=self.scpca_id)
+            if sample_id is None
+            else Sample.objects.filter(scpca_id=sample_id).first()
```
davidsmejia, line 932:
```suggestion
            else Sample.objects.filter(scpca_id=sample_id)
```

I think you want this to remain a list regardless. I believe adding `.first()` would give you back the one sample and not a list of samples

davidsmejia (changes requested):

api/scpca_portal/models/project.py
avrohomgottlieb: Good call. If a `sample_id` is provided and I'll want to grab only that sample then it should still be housed within a list in order not to cause problems when calling `len()` and iterating over `for sample in samples` later on. Thanks for pointing this out.

To document, this ternary check is only for the short term, as we'll be removing the ability to pass an individual `sample_id` into `load_data` at some point soon in the future.

davidsmejia (approved):
LGTM

Issue #693, opened 2024-05-03
Extract metadata file writing outside of combine_metadata


avrohomgottlieb:
### Context

The `Project::combine_metadata` method inside of `Project::handle_samples_metadata` accomplishes two things:
1. Combining sample and library metadata, and storing it inside a `combined_metadata[modality]` list.
2. Writing the combined metadata to an output tsv file at the per sample and per project levels.

### Problem or idea

The fact that this method does multiple things causes the unnecessary coupling of these two independent actions.

### Solution or next step

The metadata combination logic should stay within `Project::combine_metadata` , and the file writing logic should be moved to its own method, given a name like `Project::write_metadata_files`. Inside of `Project::write_metadata_files`, there should be two methods: 
- `Project::write_samples_metadata_files` 
- `Project::write_project_metadata_files`

A possible stub for `Project::write_metadata_files` could look as follows:

```python
def write_metadata_files(self, combined_metadata: Dict[List[Dict]]) -> None:
```

avrohomgottlieb:
Another approach would be to write a single method called `write_libraries_metadata`, and to write a generic method (which we can throw into utils.py) which handles writing a list of dictionaries to a file. This generic file would be used to write both sample and project level metadata files.

```python
def write_libraries_metadata(self, combined_metadata: Dict[List[Dict]]) -> None:
  for sample_libraries_metadata in combined_metadata:
    sample_output_file_name = ...
    utils.write_dict_list_to_tsv(sample_libraries_metadata, sample_output_file_name)

  project_output_file_name = ...
  utils.write_dict_list_to_tsv(combined_metadata, sample_output_file_name)

def write_dict_list_to_tsv(self, combined_libraries_metadata: List[Dict], output_file_name) -> None:
  ...
```

avrohomgottlieb:
Issue addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/698

PR #698, opened 2024-05-10
693 - Extract metadata file writing from combine metadata

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/693

Prior to this PR, 
- the combining of sample and library metadata dictionaries, and 
- the writing of those combined libraries to sample and project metadata tsv files 

lived together inside `Project::combine_metadata`. 

In order to decouple these two actions, all file writing logic within `Project::combine_metadata` was removed and placed into a new method called `Project::write_combined_metadata_libraries`. 

The new execution of `Project::load_data` looks as follows 
- Readmes are generated
- `Project::handle_samples_metadata` is called, which parses and combines sample and library metadata, and returns the `combine_metadata` libraries
- Directly afterwards, `Project::write_combined_metadata_libraries` is called, taking the `combined_metadata` libraries and writing them to disk
- Files archives are generated

In addition, with the goal of being more generic, a utils function was created, called `utils::write_dict_list_to_file`, which takes in a list of dictionaries, the output file name/path, the field names, and the type of delimiter, and carries out the file writing to disk. In `Project::write_combined_metadata_libraries`, `utils::write_dict_list_to_file` is used for the actual writing of both the sample and project metadata files to disk.

## Types of changes

## Functional tests

Tests for `write_dicts_to_file`
- `test_write_dicts_to_file_successful_write`
- `test_write_dicts_to_file_read_write_values_match`
- `test_write_dicts_to_file_incomplete_field_names`
- `test_write_dicts_to_file_not_inputted_field_names`
- `test_write_dicts_to_file_invalid_output_file`

Tests for `get_key_superset_from_dicts`
- `test_get_key_superset_from_dicts_same_keys`
- `test_get_key_superset_from_dicts_different_keys`

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/models/project.py
+++ b/api/scpca_portal/models/project.py
@@ -1180,9 +1180,7 @@ class Project(CommonDataAttributes, TimestampedModel):
             unfiltered_samples_metadata = (
                 updated_samples_metadata
                 if modality is not Sample.Modalities.MULTIPLEXED
-                else self.get_multiplexed_samples_metadata(
-                    updated_samples_metadata, sample_metadata_keys, sample_id
-                )
+                else self.get_multiplexed_samples_metadata(updated_samples_metadata, sample_id)
             )
             samples_metadata_filtered_keys = utils.filter_dict_list_by_keys(
                 unfiltered_samples_metadata, sample_metadata_keys
--- a/api/scpca_portal/models/project.py
+++ b/api/scpca_portal/models/project.py
@@ -1163,10 +1163,11 @@ class Project(CommonDataAttributes, TimestampedModel):
                 modalities.add(Sample.Modalities.CITE_SEQ)
 
             sample_metadata_keys = self.get_sample_metadata_keys(
-                set(updated_samples_metadata[0].keys()), modalities=modalities
+                utils.get_key_superset_from_dicts(updated_samples_metadata), modalities=modalities
             )
             library_metadata_keys = self.get_library_metadata_keys(
-                set(libraries_metadata[modality][0].keys()), modalities=modalities
+                utils.get_key_superset_from_dicts(libraries_metadata[modality]),
+                modalities=modalities,
             )
 
             unfiltered_samples_metadata = (
@@ -1223,11 +1224,13 @@ class Project(CommonDataAttributes, TimestampedModel):
                 continue
 
             # Establish field names for tsv
-            key_set = set(combined_metadata[modality][0].keys())
+            key_set = utils.get_key_superset_from_dicts(combined_metadata[modality])
             if modality == Sample.Modalities.MULTIPLEXED:
                 # add in non-multiplexed single-cell metadata keys to samples_metadata field_names
                 key_set = key_set.union(
-                    set(combined_metadata[Sample.Modalities.SINGLE_CELL][0].keys())
+                    utils.get_key_superset_from_dicts(
+                        combined_metadata[Sample.Modalities.SINGLE_CELL]
+                    )
                 )
             field_names = self.get_metadata_field_names(key_set, modality=modality)
 
--- a/api/scpca_portal/models/project.py
+++ b/api/scpca_portal/models/project.py
@@ -525,25 +525,6 @@ class Project(CommonDataAttributes, TimestampedModel):
             "sample_technologies_mapping": multiplexed_sample_technologies_mapping,
         }
 
-    def get_multiplexed_sample_libraries_mapping(self, multiplexed_libraries_metadata: List[Dict]):
-        """
-        Returns a dictionary which maps sample ids to a set of associated library ids.
-        get_sample_libraries_mapping() is suitable for all other modalities but not for Multiplexed,
-        which is why this method is necessary.
-        """
-        multiplexed_sample_library_mapping = {}  # Sample ID to library IDs mapping.
-        for library_metadata in multiplexed_libraries_metadata:
-            multiplexed_library_sample_ids = library_metadata["demux_samples"]
-            for multiplexed_sample_id in multiplexed_library_sample_ids:
-                # Populate multiplexed library mapping.
-                if multiplexed_sample_id not in multiplexed_sample_library_mapping:
-                    multiplexed_sample_library_mapping[multiplexed_sample_id] = set()
-                multiplexed_sample_library_mapping[multiplexed_sample_id].add(
-                    library_metadata["scpca_library_id"]
-                )
-
-        return multiplexed_sample_library_mapping
-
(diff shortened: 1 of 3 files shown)


avrohomgottlieb:
Tests for `write_dict_list_to_file` still need to be added.

api/scpca_portal/models/project.py
```diff
@@ -1217,71 +1192,92 @@ def combine_metadata(self, updated_samples_metadata, libraries_metadata, sample_
...
+        multiplexed_with_mapping = self.get_multiplexed_with_mapping(
+            combined_metadata[Sample.Modalities.MULTIPLEXED]
+        )
+
+        for modality in combined_metadata.keys():
+            if not combined_metadata[modality]:
+                continue
+
+            # Establish field names for tsv
+            key_set = set(combined_metadata[modality][0].keys())
```
davidsmejia, line 1234: Do you think we should add a utility here `get_keys_for_dicts` or something along those lines, that would iterate through the list of dicts and add keys to a set that is returned? I'm not a fan of how we assume uniformity here.

davidsmejia (commented):
LGTM just a couple comments and questions I'd like to clear up before merging.

avrohomgottlieb:
> Tests for `write_dict_list_to_file` still need to be added.

Tests added.

api/scpca_portal/utils.py
```diff
@@ -50,3 +54,15 @@ def filter_dict_list_by_keys(list_of_dicts, included_keys, *, ignore_value_error
         new_list_of_dicts.append(dictionary_copy)
 
     return new_list_of_dicts
+
+
+def write_dict_list_to_file(
+    list_of_dicts: List[Dict],
+    output_file_name: str,
+    field_names: Set,
+    delimiter: str,
```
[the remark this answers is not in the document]
avrohomgottlieb: We can do that. I thought it might be beneficial to have a more generic utils function which could write any csv-like file, but if you think that our only use case is for writing `tsv` files, then I'm with you. I'll remove the delimiter argument and default to a tab delimiter when configuring the  `DictWriter`.

api/scpca_portal/models/project.py
avrohomgottlieb: I see your point here. Let's discuss this.

api/scpca_portal/models/project.py
```diff
@@ -1217,71 +1192,92 @@ def combine_metadata(self, updated_samples_metadata, libraries_metadata, sample_
...
-                    sorted(
-                        [cm for cm in combined_metadata[modality]],
-                        key=lambda cm: (cm["scpca_sample_id"], cm["scpca_library_id"]),
-                    )
-                )
-
-        return combined_metadata
+            utils.write_dict_list_to_file(
+                sorted_combined_metadata_by_modality,
+                getattr(self, project_metadata_path),
```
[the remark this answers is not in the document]
avrohomgottlieb: There actually isn't a property method project_metadata_path in the Project model, unlike in the Sample method where there's a get_output_metadata_file_path method which takes scpca_sample_id and modality as arguments. In the Project model you have individual property methods for each modality, which follow the same output_{modality.lower()}_metadata_file_path structure (see line 1274). This is what getattr is pulling from.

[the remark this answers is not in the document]
davidsmejia: Looks good, I misread this

api/scpca_portal/utils.py
```diff
@@ -50,3 +56,15 @@ def filter_dict_list_by_keys(list_of_dicts, included_keys, *, ignore_value_error
...
+def write_dicts_to_tsv(
+    list_of_dicts: List[Dict],
+    output_file_name: str,
+    field_names: Set,
+) -> None:
+    """Writes a list of dictionaries to a csv-like file (exact delimiter provided)."""
+    with open(output_file_name, "w", newline="") as raw_file:
+        csv_writer = csv.DictWriter(raw_file, fieldnames=field_names, delimiter=common.TAB)
+        csv_writer.writeheader()
+        csv_writer.writerows(list_of_dicts)
```
davidsmejia, line 70: So I think we really just want sensible application wide defaults in this method as it is primarily syntactic sugar. This will allow us to keep implementation dry but retain the power of the underlying `csv.DictWriter` command. For example if we want to specify `N/A` for `restval` we could do that in this method for everything, but if we have a use case that requires empty strings we would still be able to supply that.

```suggestion
def write_dicts_to_file(
    list_of_dicts: List[Dict],
    output_file_name: str,
    **kwargs,
) -> None:
    kwargs["fieldnames"] = kwargs.get("fieldnames", ...)
    kwargs["delimeter"] = kwargs.get("delimiter", common.TAB)
    with open(output_file_name, "w", newline="") as raw_file:
        csv_writer = csv.DictWriter(raw_file, **kwargs)
        csv_writer.writeheader()
        csv_writer.writerows(list_of_dicts)
```

api/scpca_portal/models/project.py
avrohomgottlieb: Added.

api/scpca_portal/utils.py
avrohomgottlieb: Added.

avrohomgottlieb:
Test have to be written for `get_key_superset_from_dicts`, and then we'll be ready to go.

api/scpca_portal/utils.py
```diff
@@ -50,3 +56,26 @@ def filter_dict_list_by_keys(list_of_dicts, included_keys, *, ignore_value_error
...
     return new_list_of_dicts
+
+
+def get_key_superset_from_dicts(list_of_dicts: List[Dict]) -> Set:
+    """Extracts each provided dictionary's key set and returns the superset of all key sets."""
+    key_superset = set()
+    for dictionary in list_of_dicts:
+        key_superset.update(set(dictionary.keys()))
+
+    return key_superset
```
davidsmejia, line 67: I think that we could just have `get_keys_from_dicts` and then the docstring can describe that it is a union of all of the dict's keys.

Also, I don't think `list_of_dicts` is necessary as we have type hinting and we don't generally need to write the type in the variable name.

```suggestion
def get_keys_from_dicts(dicts: List[Dict]) -> Set:
    """Takes a list of dictionaries and returns a set equal to the union of their keys."""
    return set(k for d in dicts for k in d.keys())
```

davidsmejia: Or if you import `chain` from `itertools` you can do this instead but I think calling `keys` explicitly is a bit nicer
```python
return set(key for key in chain(*dicts))
```

avrohomgottlieb: Nice. I'll update accordingly.

Issue #694, opened 2024-05-06
Move project property assignments to project model from load_data command


avrohomgottlieb:
### Context

`management/commands/load_data` (referred to as `load_data` or the `load_data` command) describes the implementation of the interface for the custom django `load_data` management command. The goal of this command is to kick of the execution of the data loading pipeline, which includes 
- Downloading gene expression data objects (SCE and/or Anndata) and their associated metadata from an S3 bucket
- Parsing the metadata files and writing new ones 
- Zipping up the SCE/Anndata objects with the newly produced metadata files at the sample and project levels
- Uploading the new zip archive to another S3 bucket where it will be hosted for the front-end consumption

The current `load_data` command, parses and applies command line arguments which allow for customization of the command, configures aws accordingly, and calls a `load_data::load_data` method which kicks of the program execution, ultimately leading to a call to `Project::load_data`.

### Problem or idea

The `load_data` command should really only be responsible for parsing the command line argument and kicking off the load data pipeline. In this issue, we'll focus on phasing out project property assignments to the project model.

### Solution or next step

`load_data::load_data` creates a new `project` db object, and then calls `load_data::process_project_data` which configures and assigns values to a variety of project properties, and then concludes by calling `Project::load_data` on the `project` object. This method, `load_data::process_project_data`, should be moved to the `Project` model, allowing for the project configuration to happen in its correct domain. As mentioned above, the`load_data` command should be solely responsible for parsing and handling of command line arguments, and for kicking off the `load_data` pipeline, as mentioned above.

As an implementation note - there are some command line arguments which are passed in to `load_data::process_project_data` which will need to be forwarded to it within the `Project` model.

avrohomgottlieb:
Issue addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/700

PR #700, opened 2024-05-14
694 - Move project property assignments from load_data command to Project model

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/694

This PR addresses a design improvement of moving the assignment of project-level properties (to the active project object) from the `load_data` command to `Project::load_data`. 

From an implementation perspective, the necessary changes include moving the `load_data::process_project_data` function to the `Project` model, and sending in, as an additional argument to `Project::load_data`, the project-level properties dictionary (which is parsed from the top-level `project_metadata.csv` file that `load_data::load_data` uses to create the `Project` objects to begin with).

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/models/contact.py
+++ b/api/scpca_portal/models/contact.py
@@ -50,11 +50,11 @@ class Contact(TimestampedModel):
             return
 
         for idx, email in enumerate(emails):
-            # Skip contact if already in db
-            if Contact.objects.filter(email=email):
+            if email in common.IGNORED_INPUT_VALUES:
                 continue
 
-            if email in common.IGNORED_INPUT_VALUES:
+            # Skip contact if already in db
+            if Contact.objects.filter(email=email):
                 continue
 
             contact_data = {"name": names[idx], "email": email, "pi_name": project.pi_name}
--- a/api/scpca_portal/models/contact.py
+++ b/api/scpca_portal/models/contact.py
@@ -34,7 +34,7 @@ class Contact(TimestampedModel):
 
     @staticmethod
     def bulk_create_from_project_data(project_data, project):
-        """Creates a list of contact objects and then save them."""
+        """Creates a list of contact objects and saves them."""
         emails = [
             email.lower().strip()
             for email in project_data["contact_email"].split(common.CSV_MULTI_VALUE_DELIMITER)
--- a/api/scpca_portal/models/contact.py
+++ b/api/scpca_portal/models/contact.py
@@ -1,6 +1,12 @@
+from typing import Dict
+
 from django.db import models
 
+from scpca_portal import common
+from scpca_portal.config.logging import get_and_configure_logger
 from scpca_portal.models.base import TimestampedModel
+
+logger = get_and_configure_logger(__name__)
 
 
 class Contact(TimestampedModel):
@@ -15,3 +21,44 @@ class Contact(TimestampedModel):
 
     def __str__(self) -> str:
         return f"{self.name} <{self.email}>"
+
+    @classmethod
+    def get_from_dict(cls, data: Dict):
+        contact = cls(
+            name=data.get("name"),
+            email=data.get("email"),
+            pi_name=data.get("pi_name"),
+        )
+
+        return contact
+
+    @staticmethod
+    def bulk_create_from_project_data(project_data, project):
+        """Creates a list of contact objects and then save them."""
+        emails = [
+            email.lower().strip()
+            for email in project_data["contact_email"].split(common.CSV_MULTI_VALUE_DELIMITER)
+        ]
+        names = [
+            name.strip()
+            for name in project_data["contact_name"].split(common.CSV_MULTI_VALUE_DELIMITER)
+        ]
+        contacts = []
+
+        if len(emails) != len(names):
+            logger.error("Unable to add ambiguous contacts.")
+            return
+
+        for idx, email in enumerate(emails):
+            # Skip contact if already in db
+            if Contact.objects.filter(email=email):
+                continue
+
+            if email in common.IGNORED_INPUT_VALUES:
+                continue
+
+            contact_data = {"name": names[idx], "email": email, "pi_name": project.pi_name}
+            contacts.append(Contact.get_from_dict(contact_data))
+
+        Contact.objects.bulk_create(contacts)
+        project.contacts.add(*contacts)
--- a/api/scpca_portal/models/contact.py
+++ b/api/scpca_portal/models/contact.py
@@ -37,15 +37,22 @@ class Contact(TimestampedModel):
         """Creates a list of contact objects and saves them."""
         contacts = []
 
-        keys = ["contact_email", "contact_name"]
-        for email, name in utils.get_csv_zipped_values(
-            project_data, *keys, model_name=cls.__name__
-        ):
+        try:
+            zipped_contact_details = utils.get_csv_zipped_values(
+                project_data, "contact_email", "contact_name"
+            )
+        except Exception:
+            logger.error("Unable to add ambiguous contacts.")
+            raise
+
+        for email, name in zipped_contact_details:
             if email in common.IGNORED_INPUT_VALUES:
                 continue
 
-            # Skip contact if already in db
-            if Contact.objects.filter(email=email):
+            # Handle case where contact is already in db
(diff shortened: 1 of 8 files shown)


avrohomgottlieb:
This PR only has one commit, https://github.com/AlexsLemonade/scpca-portal/pull/700/commits/60284049bf68dc941ca6e208e574a11f750a3691. The rest is a result of rebasing from branch connected to PR https://github.com/AlexsLemonade/scpca-portal/pull/698

davidsmejia (commented):
After looking at this closer, I think we should make the following changes:
- Instead of passing in the properties into `load_data` we should create a separate factory method that takes in the project properties.
- In load data check if the project `exists()` in the database if it does do the same exit w/ warning - otherwise get an instance of the project from the new factory method, then call load_data on that project to create the sample / computed files

Looking forward we will want to have a factory method that creates the project and all of it's relationships, ie: `external_accessions`, `contacts`, `publications`, and `samples`. Then `libraries` when creating `samples`. Then we can optionally create metadata only downloads after that is done or we can make that a separate command in the same way that we will for creating the `ComputedFiles` for zips that contain expression data.

davidsmejia:
From our conversation about renaming to `get_from_dict`

```python
        # assign values
        for key in data.keys():
            if hasattr(self, key):
                if key.startswith('includes_') or key.startswith('has_'):
                    setattr(self, key, util.boolean_from_string(data.get(key, False)))
                else:
                    setattr(self, key, data.get(key))

```

avrohomgottlieb:
@davidsmejia 
I added the factory method that we discussed, `Project::get_from_dict`. 

Two points to discuss here:
- Do you have any thoughts on the extensive key massaging in `Project::get_from_dict`? Is it better for us to handle the discrepancies between the csv field names and our model attribute names are on end, or is this a something that we should bring up with the science team to see if the field names can be updated?
- In order to establish the many-to-many relationships inside of `Project::get_from_dict` (`add_contacts`, `add_external_accessions`, `add_publications`), the project object needs to be saved so that it can receive an id. This doesn't feel like the right behavior, as the standard `get_from_dict` API that we discussed is only meant to retrieve an object representation, and allow for the caller to save these representations to the db at a later point.

davidsmejia:
> Two points to discuss here:
> 
>     * Do you have any thoughts on the extensive key massaging in `Project::get_from_dict`? Is it better for us to handle the discrepancies between the csv field names and our model attribute names are on end, or is this a something that we should bring up with the science team to see if the field names can be updated?

This is something we should address here, at least in the context of this PR. I think that there is an argument for creating a new module that would rename dictionary keys. This would take an expected input dict and return what we want to use directly. This file could all of the different transforms that we would use depending on the context. This would allow our individual methods to not care about that and just receive the transformed dictionaries ready for use. I am unsure if we should pass the input structure into the methods or transform them inside of the methods that use them.

>     * In order to establish the many-to-many relationships inside of `Project::get_from_dict` (`add_contacts`, `add_external_accessions`, `add_publications`), the project object needs to be saved so that it can receive an id. This doesn't feel like the right behavior, as the standard `get_from_dict` API that we discussed is only meant to retrieve an object representation, and allow for the caller to save these representations to the db at a later point.

Yeah, `get_from_dict` should just create the unsaved model instance. I think the appropriate thing to do here is remove `add_contacts` from the project model and create a factory method on the `Contact` model. This would be true for all of the relationships.

example:
```python
project = Project.get_from_dict(project_data)
project.save()
contacts = Contact.get_from_project_data(project_data)
project.contacts.add(contacts)
```

avrohomgottlieb:
> 
> Yeah, `get_from_dict` should just create the unsaved model instance. I think the appropriate thing to do here is remove `add_contacts` from the project model and create a factory method on the `Contact` model. This would be true for all of the relationships.
> 
> example:
> 
> ```python
> project = Project.get_from_dict(project_data)
> project.save()
> contacts = Contact.get_from_project_data(project_data)
> Contact.objects.bulk_create(contacts)
> project.contacts.add(contacts)
> ```

Addressed 👍

api/scpca_portal/models/contact.py
```diff
@@ -15,3 +21,44 @@ class Meta:
...
+        contact = cls(
+            name=data.get("name"),
+            email=data.get("email"),
+            pi_name=data.get("pi_name"),
+        )
+
+        return contact
+
+    @staticmethod
+    def bulk_create_from_project_data(project_data, project):
```
davidsmejia, line 36: I was thinking about this a little more and wanted to suggest a new utility method.

```python
def get_delimited_values_as_dicts(data, *args, delimiter=common.CSV_MULTI_VALUE_DELIMITER):
    # Split the values we care about by their delimiter
    split_values = [data[key].split(delimiter) for key in args]
    # collect related values into their own list
    zipped_delimited_values = zip(*split_values, strict=True)
    # create a list of dictionaries that contain zipped key value pairs
    return [dict(zip(args, values)) for values in split_values]
```
The other alternative would be to return the `zipped_delimited_values` (converted to a list probably) and then still do another iteration on them and construct your own dictionaries.

I think this would clean up a lot of the details for each out this could be passed into the `get_from_dict` factory methods, but we would still want to clean up or validate the inputs. We would also want to catch the error from a length mismatch.

davidsmejia (commented):
Looks good, just wanted to make one suggestion that would help DRY this a bit.

api/scpca_portal/models/contact.py
avrohomgottlieb: I looked into this. Because of the varied nature of the data dictionary value transformations (some send parameters to `strip()`, others use `utils.boolean_from_string()`, another is a `project.pi_name` attribute), it was hard to take the extra step and to make the dictionary for `get_from_dict` truly generic, as proposed here. Check out the direction I took and let me know if you think there's a way we can organize better.

avrohomgottlieb: If you like the utils function that I added, then I'll write tests for it and we can close this guy up.

api/scpca_portal/utils.py
```diff
@@ -50,3 +59,52 @@ def filter_dict_list_by_keys(list_of_dicts, included_keys, *, ignore_value_error
...
+    kwargs["fieldnames"] = kwargs.get("fieldnames", get_key_superset_from_dicts(list_of_dicts))
+    kwargs["delimiter"] = kwargs.get("delimiter", common.TAB)
+
+    with open(output_file_path, "w", newline="") as raw_file:
+        csv_writer = csv.DictWriter(raw_file, **kwargs)
+        csv_writer.writeheader()
+        csv_writer.writerows(list_of_dicts)
+
+
+def get_csv_zipped_values(
```
davidsmejia, line 87: 1. We should remove the `model_name` argument as it is only used for error messaging.
2. Don't catch the exception just to log it, we want to fail here as it is indicative of malformed input data which means we probably don't want to process this csv file.
3. zipped values is an empty list but returns a zip object when successful this should just be either a list returned or a zip object.

Going to respond to an anticipated response being, we uniformly handle the errors so this would be less repetitive. I agree with that but it still think the caller should handle the error. Util functions shouldn't be logging, just throwing an expressive error so throwing a ValueError, *args are of non-equal lengths would be sufficient. The caller can decide to log, catch the exception, neither or both.

api/scpca_portal/models/publication.py
```diff
@@ -20,3 +26,39 @@ def __str__(self) -> str:
...
+
+        keys = ["citation_doi", "citation"]
+        for doi, citation in utils.get_csv_zipped_values(
+            project_data, *keys, model_name=cls.__name__
+        ):
+            if doi in common.IGNORED_INPUT_VALUES:
+                continue
+
+            # Skip if already in db
+            if Publication.objects.filter(doi=doi):
```
davidsmejia, line 53: I think this is the general flow we want.
```suggestion
            if existing_pub := Publication.objects.get(doi=doi).exists():
                if not project.publications.get(existing_pub).exists():
                   project.publications.add(existing_pub)
                 continue
```

api/scpca_portal/models/contact.py
davidsmejia: Yeah I think this is fine as an approach

Issue #639, opened 2024-04-09
Move sample ComputedFile generation from `Product::load_data` to its own method inside `ComputedFile` model


avrohomgottlieb:
### Context

This issue is connected to issue https://github.com/AlexsLemonade/scpca-portal/issues/636.

In `Project::load_data`, file computation is first carried out at the sample level and then at the project level. The project level file computation is abstracted out to `Project::create_computed_files`, while the sample level file computation is executed within the `load_data` method.

### Solution or next step

Sample level computation should be abstracted into its own method, similar to project level computation. As well, there should be some name changing to the `Project::create_computed_file` should be renamed for clarity that it only acts on projects (and not samples) and moved to the `ComputedFile` model.

avrohomgottlieb:
Issue addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/702

Issue #636, opened 2024-04-09
Move Project ComputedFile generation logic from `Project::load_data` to new method inside `ComputedFile` model


avrohomgottlieb:
### Context

The `load_data` method inside the `Project` model accomplishes a variety of different tasks. Among those tasks includes writing readmes, parsing metadata, writing metadata, and generating computed files.

Computed file generation is carried out after metadata parsing, metadata writing, first at the sample level, and then at the project level.

### Problem or idea

The file computation logic is too tightly coupled with the other tasks within the `load_data` method.

### Solution or next step

All file computation logic should be moved out of the `load_data` method in the `Project` model into its own method, and then transferred to the `ComputedFile` model.

avrohomgottlieb:
Issue addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/702

PR #702, opened 2024-05-15
639 636 - Extract file computation logic from Project::load_data

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/639
https://github.com/AlexsLemonade/scpca-portal/issues/636

This PR extracts Sample file computation out of `Project::load_data` (something which had already been done for Project file computation) and moves it to its own method `create_sample_computed_files`. The original idea was to move both `create_project_computed_files` and `create_sample_computed_files` to the `ComputedFile` model, as was indicated in the above issues, but the reliance on data structures which use `Sample.Modalities` gave rise to the simpler solution of moving each method to the model of the objects being acted on
- keeping `create_project_computed_files` in the `Project` model, and
- moving `create_sample_computed_files` to the `Sample` model.

This PR also cleans up duplicate code within the `create_<type>_computed_files` methods (including the consolidation of the `update_<modality>_data` callbacks in `create_sample_computed_files` into a generic `create_sample_computed_file` callback), and updates the naming of a few data structures to enhance clarity and naming consistency.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


avrohomgottlieb:
Please note that because of the decisions that we made to not move either method to `ComputedFile`, I've decided to write one PR to address both of the above `issues`.

avrohomgottlieb:
Other items that can be covered in this PR:
- combine ComputedFile::get_<type>_<modality>_mapping methods (this is definitely doable)
- update the way we do file mappings (might be too big of a goal for this PR)

api/scpca_portal/migrations/0043_computedfile_public.py
davidsmejia, line 1: We should set `public` to true when applying this migration. For reference you can check this file `api/scpca_portal/migrations/0032_alter_project_modalities.py`

You would only need to worry about applying the migration, not the reverse.
You should also bulk_update, since there are ~1300 computed files we shouldn't have to worry about pagination.

api/scpca_portal/models/computed_file.py
```diff
@@ -78,6 +78,7 @@ class OutputFileFormats:
     format = models.TextField(choices=OutputFileFormats.CHOICES)
     includes_merged = models.BooleanField(default=False)
     modality = models.TextField(choices=OutputFileModalities.CHOICES)
+    public = models.BooleanField(default=False)
```
davidsmejia, line 81: Let's override the default `objects` manager and replace it with one that the default queryset has `public=True`. 

We can set `all_objects` to be the default manager which would apply to public and private computed files.

Ref: https://docs.djangoproject.com/en/3.2/topics/db/managers/

api/scpca_portal/models/project.py
```diff
@@ -409,37 +386,42 @@ def create_computed_file(future):
...
-                        multiplexed_file_mapping,
-                        multiplexed_workflow_versions,
+                        file_mappings_by_format[file_format].get(Sample.Modalities.MULTIPLEXED),
+                        workflow_versions_by_modality[Sample.Modalities.MULTIPLEXED],
                         file_format,
-                    ).add_done_callback(create_computed_file)
+                    ).add_done_callback(create_project_computed_file)
 
-                if single_cell_file_mapping.get(file_format):
+                if file_mappings_by_format[file_format].get(Sample.Modalities.SINGLE_CELL):
```
davidsmejia, line 408:
```suggestion
                if single_cell_file_mapping := file_mappings_by_format[file_format].get(Sample.Modalities.SINGLE_CELL):
```

I think this is a good place to use the :walrus: operator. 😎

davidsmejia (commented):
Looking good, made a small suggestion for shortening some lines. Also, we should make this PR atomic for handling public/private computed files by creating a manager here and also setting existing computed files to public.

avrohomgottlieb:
As a note, we've decided not to touch add the `public` attribute to the ComputedFile model during this PR, and instead will implement it in a future PR.

davidsmejia (approved):
LGTM!

Issue #706, opened 2024-05-17
Dictionary transformations module


avrohomgottlieb:
### Context

 
Throughout `Project::load_data`, metadata files are parsed and their contents are saved to `Project` and `Sample` objects.

### Problem or idea

The metadata files sometimes have keys which do not match with the relevant model attributes within the `Project` and `Sample` models.

### Solution or next step

A separate module should be created which transforms the mismatched keys of these newly parsed metadata files to the intended key names that match the model attributes. 

The following areas are where transformations should occur:
- Parsing of `project_metadata`
- Parsing of project relationships before passing to factory methods
- Parsing of libraries json before combination with samples
- Parsing of samples tsv before combination with libraries

avrohomgottlieb:
Addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/718

PR #718, opened 2024-05-21
706 - Factory Data Transforms Module

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/706

This PR introduces a `factory_data_transforms` module which includes utility methods that transform mismatched key names, originally taken from parsed metadata files, to match associated model attributes. These functions essentially act as wrappers when the metadata files are being loaded into memory as dictionaries (or lists of dictionaries), and update the key names accordingly.

## Types of changes

## Functional tests

N/A 
## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/management/commands/load_data.py
+++ b/api/scpca_portal/management/commands/load_data.py
@@ -14,6 +14,7 @@ from botocore.client import Config
 
 from scpca_portal import common
 from scpca_portal.models import Contact, ExternalAccession, Project, Publication
+from scpca_portal.models.factory_data_transforms import project_data_transform
 
 ALLOWED_SUBMITTERS = {
     "christensen",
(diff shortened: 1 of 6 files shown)


avrohomgottlieb:
I'm not 100% sure that the right place for this module is inside the `models/` directory. 

Other alternatives for the location of this module include:
- Putting it inside a new `utils/` directory inside of `models/`
- Putting it inside a new `utils/` directory alongside `utils.py` in the `scpca_portal/` directory
- Putting it inside `scpca_portal/` next to `common.py` and `utils.py`

davidsmejia (commented):
This is the right start. I left some comments but I think we should change how we think about this module and make it responsible for the reading and writing of metadata files.

So lets have 3 functions that read
- project_metadata
- sample_metadata
- library_metadata
And one that writes:
library_metadata.tsv (taken from utils)

davidsmejia:
Looks good just need to remove the old `transforms.py`

PR #719, opened 2024-05-22
Remove left over reference to self.project in load_data::clean_up_input_data

avrohomgottlieb:
## Issue Number

N/A

During the recent refactor of `Project::load_data`, project property assignments, which were previously carried out inside of `load_data::load_data`, were converted into a factory method and moved to the project model. During this refactor, all references to `self.project` were removed. It came up that one reference was left by accident, inside of `load_data::clean_up_input_data`, and that is what was cleaned up in this PR.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


davidsmejia (approved):
LGTM

PR #720, opened 2024-05-22
710 - Create library model

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/710

## Purpose/Implementation Notes

This PR contains the initial creation of a `Library` model. 

The `Library` model's initial attributes include general properties about a given library, including its `scpca_id`, `format`, `modality`, where or not it's `multiplexed`, and `seq_unit` and `technology` info parsed from the library's json metadata file. As well, the library's `dir_path` will be saved here.

One open question here pertains to how we intend to handle cell_counts, as non_multiplexed sample libraries pull values from `filtered_cell_count`, while multiplexed sample libraries retrieve their `demux_cell_count_estimate` attributes from a `sample_cell_estimates` dictionary included in library json files that have multiplexed samples.

Future PRs will include the implementation of this new model throughout the codebase.

## Types of changes

What types of changes does your code introduce?

- New feature (non-breaking change which adds functionality)

## Functional tests

N/A

## Checklist

- [x] Lint and unit tests pass locally with my changes

## Screenshots

N/A

Issue #711, opened 2024-05-20
Create a factory method to generate from a dictionary


avrohomgottlieb:
### Context

Within the `Project` and `Sample` models there are factory functions which facilitate the easy creation of objects to be saved and queried when needed. This behavior should be implemented as well in the new `Library` model.

### Solution or next step

A `get_from_dict` class method should be created which takes in a dictionary of relevant `Library` model attributes, and returns a new object to be saved at the caller's discretion. Depending on the exact use case, it likely that a collection of library objects should be formed and then created in bulk with a `bulk_create_from_dicts` method, like in the `Sample` models, to facilitate this behavior.

avrohomgottlieb:
Issue addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/728

PR #728, opened 2024-05-29
711 - Create library model factory method and add it to Project::load_data

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/711

This PR takes the creation the `Library` model forward from PR https://github.com/AlexsLemonade/scpca-portal/pull/720 and begins to implement it. 

The following three items are included in the current PR:

- `Library::get_from_dict` and `Library::bulk_create_from_dicts` methods are built, similar to those in the `Sample` model. These methods are responsible for building a collection of new `Library` objects, and saving them to the db.
- The building of `Library` objects is added to `Project::load_libraries_metadata`. In order to accommodate this, the building and creating of `Sample` objects is moved from the end of `Project::handle_samples_metadata` to `Project::load_samples_metadata`, which in any case is a more appropriate place for this activity. This change was important in order that `Sample` objects should be built before `Library` objects so that the new `Library` objects could be related to the already created `Sample` objects.
- The updating of `Project` purging logic in `Project::purge` to include the purging of related `Library` objects from the db.

## Types of changes

## Functional tests

Still need to be written.

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/models/library.py
+++ b/api/scpca_portal/models/library.py
@@ -34,3 +34,16 @@ class Library(TimestampedModel):
     modality = models.TextField(choices=Modalities.CHOICES)
     scpca_id = models.TextField(unique=True)
     workflow_version = models.TextField()
+
+    @classmethod
+    def get_from_dict(cls, data, formats, is_multiplexed, modality):
+        library = cls(
+            formats=formats,
+            is_multiplexed=is_multiplexed,
+            metadata=data,
+            modality=modality,
+            scpca_id=data["library_id"],
+            workflow_version=data["workflow_version"],
+        )
+
+        return library
--- a/api/scpca_portal/models/library.py
+++ b/api/scpca_portal/models/library.py
@@ -49,7 +49,7 @@ class Library(TimestampedModel):
                 if "filtered_cell_count" in data
                 else Library.Modalities.SPATIAL
             ),
-            scpca_id=data["library_id"],
+            scpca_id=data["scpca_library_id"],
             workflow_version=data["workflow_version"],
         )
 
@@ -66,8 +66,7 @@ class Library(TimestampedModel):
                 if any(sample_dir.glob("*.h5ad")):
                     formats.append(Library.FileFormats.ANN_DATA)
             else:
-                if any(sample_dir.glob("*_spatial/*.rds")):
-                    formats.append(Library.FileFormats.SINGLE_CELL_EXPERIMENT)
+                formats.append(Library.FileFormats.SINGLE_CELL_EXPERIMENT)
 
             libraries.append(Library.get_from_dict(library_json, formats))
 
--- a/api/scpca_portal/models/library.py
+++ b/api/scpca_portal/models/library.py
@@ -1,3 +1,6 @@
+from pathlib import Path
+from typing import Dict, List
+
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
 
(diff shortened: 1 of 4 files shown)


api/scpca_portal/test/management/commands/test_load_data.py
```diff
@@ -430,7 +430,7 @@ def test_multiplexed_metadata(self):
             "disease_ontology_term_id",
             "droplet_filtering_method",
             "filtered_cell_count",  # with non-multiplexed
-            "filtered_cells",
+            # "filtered_cells",
```
avrohomgottlieb, line 433: Is there a reason that we keep this `filtered_cells` property around when testing Multiplexed? In `metadata_file.py`, its one of the properties that we transform, replacing it with `filtered_cell_count`. With Single-cell we've removed it.

api/scpca_portal/models/project.py
```diff
@@ -892,20 +889,22 @@ def load_libraries_metadata(self, samples_metadata: List[Dict]):
                 list(Path(sample_dir).glob("*_metadata.json"))
                 + list(Path(sample_dir).rglob("*_spatial/*_metadata.json"))
             )
```
davidsmejia, line 886: What if instead of only creating one variable here we created two one for single_cell_library_metadata and another for spatial. Then instead of checking a random key in the dict we could just check which list the current one came from is in? We could also attach modality here.

davidsmejia (commented):
Looking good wanted to get some clarity and opinions from you before we finish this review.

api/scpca_portal/test/management/commands/test_load_data.py
davidsmejia: This is ok to remove here and not comment out

api/scpca_portal/models/project.py
```diff
@@ -1063,6 +1079,10 @@ def purge(self, delete_from_s3=False):
                 if delete_from_s3:
                     computed_file.delete_s3_file(force=True)
                 computed_file.delete()
+            for library in sample.libraries.all():
+                # If library has other samples that it is related to, then don't delete it
+                if len(library.samples.all()) == 1:
```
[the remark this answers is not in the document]
avrohomgottlieb: What is accomplished by the first condition in the `if` statement? Isn't the existence relation obvious by the fact that this library was retrieved from a query of this exact sample? 

I understand that what you're going for here in improved code readability, and I'm with you 100%. What do you think about the following?

```suggestion
# Don't delete libraries with relations to other samples
if not (library.samples.count() > 1):
```

api/scpca_portal/models/library.py
```diff
@@ -34,3 +37,38 @@ class Modalities:
...
+    def get_from_dict(cls, data, formats):
+        library = cls(
+            formats=formats,
+            is_multiplexed=("demux_samples" in data),
+            metadata=data,
+            modality=(
+                Library.Modalities.SINGLE_CELL
+                if "filtered_cell_count" in data
+                else Library.Modalities.SPATIAL
+            ),
```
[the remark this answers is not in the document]
avrohomgottlieb: Great suggestion.

api/scpca_portal/models/project.py
avrohomgottlieb: Love this suggestion. Added in.

api/scpca_portal/models/library.py
```diff
@@ -34,3 +37,38 @@ class Modalities:
...
+            scpca_id=data["scpca_library_id"],
+            workflow_version=data["workflow_version"],
+        )
+
+        return library
+
+    @classmethod
+    def bulk_create_from_dicts(cls, library_jsons: List[Dict], sample, sample_dir: Path) -> None:
```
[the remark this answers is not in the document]
avrohomgottlieb: I think we should have a discussion about this. There is a method on `Project` called `get_sample_input_data_dir` which takes in a `sample_id` as an argument and returns the Path object for the `sample_dir`. Part of me feels like we should just move this method to a property method on `Sample`, but there are occurences where this `sample_dir` is needed before the creation of `Sample` objects, so that's a question needed to be addressed. 

As it pertains to your question here, I was able to get around passing through the `sample_dir` into `Libary::bulk_create_from_dicts`. Please see the group of most recent commits.

api/scpca_portal/models/project.py
```diff
@@ -881,30 +872,35 @@ def load_libraries_metadata(self, samples_metadata: List[Dict]):
...
-                sample_seq_units.add(library_json["seq_unit"].strip())
-                sample_technologies.add(library_json["technology"].strip())
-
-                if "spatial" in str(filename_path):
-                    libraries_metadata[Sample.Modalities.SPATIAL].append(library_json)
-                else:
-                    libraries_metadata[Sample.Modalities.SINGLE_CELL].append(library_json)
+            for library_metadata, library_path in zip(
+                all_libraries_metadata, library_metadata_paths
+            ):
```
davidsmejia, line 891:
```suggestion
            all_libraries_metadata = []
            for metadata_path in library_metadata_paths:
                library_metadata = metadata_file.load_library_metadata(metadata_path)
                all_libraries_metadata.applend(library_metadata)
```

davidsmejia: now that we extracted the modality and format outside of bulk create lets remove the zip as well

davidsmejia (commented):
small change looks good

davidsmejia (approved):

PR #745, opened 2024-06-05
Fix issue with duplicate emails

davidsmejia:
## Issue Number

n/a

While processing data for staging, most projects went through successfully however there was an issue where we formatted data and put it in the DB after checking if the unformatted data was in the DB. Obviously this doesn't work so this PR  changes the order so that we check if the formatted values are in the DB.

- also updated models that do something similar to prevent future issues.

## Types of changes

## Functional tests

n/a

## Checklist

## Screenshots

n/a


avrohomgottlieb (approved):
LGTM!

Issue #731, opened 2024-05-30
Add data_file_paths ArrayField attribute to Library model, introduce dynamic data path downloads


avrohomgottlieb:
### Context

The `load_data` management command is responsible for all of the following actions:
- Downloading all of an inputted project's and associated sample's data files (rds/h5ad) and metadata files (csv/json) from an s3 input bucket
- Creating a `Project` instance and calling `Project::load_data` on the `Project` instance (which parses and writes an output metadata files, archives project and sample computed files, and uploads those files to an s3 output bucket)
- Cleaning up the local file system

Currently, `Sample` computed files are generated by writing these data files (and the newly generated output metadata file) to a zip archive created on a per sample basis, and then accumulating the paths to these files in a mapping which maps sample ids to their included data file paths. This file mapping grows as it continues to collect new sample ids with their data file paths as each sample is archived. At the conclusion of all of the samples, this mapping data structure is then passed to the method that handles the archiving of `Project` computed files.

### Problem or idea

A few problems with this approach:
- The way that we accumulate file mappings while archiving `Sample` computed files creates an unnecessary dependency for `Project` file computation that we have to worry about
- The aggregation of these file mappings is a side affect of `Sample` file computation and doesn't have to be carried out then and there
- The fact that we keep this data structure around in memory and send references to it across different methods unnecessarily couples our code in a way that is not desirable, while decreasing readability and code clarity

### Solution or next step

Three next steps are being proposed to clean up this part of the codebase and allow us to better decouple concerns here:
- Add a `data_file_paths` `ArrayField` attribute to the `Library` model which will maintain the library to data file path mappings on an instance by instance basis.
- Add a helper method which facilitates downloading data files lazily, as they will be needed, as opposed to the current approach which downloads everything eagerly right at the beginning.

The implementation would look as follows:
- When `Library` objects are being created inside of `Project::load_libraries_metadata`, the file tree associated with that library's sample's directory will be pulled down from s3, and depending on the modality of the library, the `data_file_paths` array will be populated based on the relevant data file names.
- Later on, when it comes time to archiving files (which begins during our current implementation with the archiving of `Sample` computed files), the relevant libraries will be queried from the database, and each library's associated `data_file_paths` will be downloaded at that point from s3, and then written directly to the archive.

[...]

PR #747, opened 2024-06-06
731 - Add data file paths attribute and getter methods to Library and Project models

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/731

Here is a list of the changes made in this PR:
- `Utils` file
  - Added `list_s3_paths` function which queries an inputted s3 bucket (with an optional `relative_path` to concatenate to end of inputted bucket) and returns all of bucket's entries

- `Library` model
  - Added `data_file_paths` attribute
  - Added `get_data_file_paths` and `add_data_file_paths` methods which call the new `utils::list_s3_paths` function looking for library's data file paths, and then assigns them to inputted objects
  - Added `add_modality_and_formats` method which calculates and assigns these properties to library objects based on `data_files_paths`
  - Updated `bulk_create_from_dicts` to call `add_data_file_paths` and `add_modality_and_formats` on a collection of libraries.

- `Project` model
  - Added `data_file_paths` attribute
  - Added `get_data_file_paths` method which returns all `merged` and `bulk` project files
  - Updated `get_from_dict` to call `get_data_file_paths` and assign the return call to `project.data_file_paths`
  - Integrated loading of multiplexed libraries into `Project::load_libraries_metadata`, and deleted `Project::get_multiplexed_libraries_metadata` method

## Types of changes

## Functional tests

5 unit tests added for `utils::list_s3_paths` inside of `test/test_utils.py`:
- `test_list_s3_paths_success`
- `test_list_s3_paths_with_relative_paths`
- `test_list_s3_paths_with_public_bucket`
- `test_list_s3_paths_command_failure`
- `test_list_s3_paths_with_nested_directory`

## Checklist

## Screenshots

N/A


api/scpca_portal/models/library.py
```diff
@@ -72,3 +81,33 @@ def get_file_formats(sample_dir: Path):
...
+        """
+        Retrieves all data file paths on the aws input bucket associated with the Library object
+        and returns them as a list
+        """
+        project_id = self.samples.first().project.scpca_id
+        sample_id = (
+            self.samples.first().scpca_id
+            if self.samples.count() == 1
+            else ",".join([sample.scpca_id for sample in self.samples.all()])
+        )
```
davidsmejia, line 95: Grab this off of `self.metadata.get("sample_id")` I believe it will be available as a dict

api/scpca_portal/utils.py
```diff
@@ -92,3 +94,41 @@ def get_csv_zipped_values(
...
+    output = result.stdout
+
+    bucket_entries = []
+    for line in output.splitlines():
+        if line.strip().startswith("PRE"):
+            bucket_entries.append(BucketPrefixEntry._make(line.split()))
+        else:
+            bucket_entries.append(BucketObjectEntry._make(line.split()))
+
+    file_paths = [Path(entry.file_path) for entry in bucket_entries]
```
davidsmejia, line 132: before returning this, lets split the path by `/` and if the buck name ends with that path just remove it. put a comment as well

avrohomgottlieb:
FYI, something to keep in mind. Hitting aws every time we encounter a new library and want to grab its data file paths may introduce a bottleneck. Performance wise, my testing suite ran roughly 3x slower when comparing this branch to dev.

`avrohom/731-add-data-file-paths-attribute-to-library-model`
image

`dev`
image


An alternative approach could be:
- Hit aws one time per project in the beginning
- Grab all of the files paths then and there
- Stash them in an `all_file_paths` attribute on the `Project` model
- *Then* query for them as needed when iterating over libraries later on

api/scpca_portal/common.py
```diff
@@ -18,6 +18,9 @@
 
 TEMPLATE_PATH = CODE_PATH / "scpca_portal" / "templates"
 
+TEST_INPUT_BUCKET_NAME = "scpca-portal-public-test-inputs/2024-04-19/"
+INPUT_BUCKET_NAME = TEST_INPUT_BUCKET_NAME if settings.TEST else "scpca-portal-inputs"
```
avrohomgottlieb, line 22: The output bucket is in `config/local.py` and/or `config/production.py`. Should we move the input bucket/s to the `config/` dir as well?

avrohomgottlieb:
Tests for the new utils function still need to be added.

avrohomgottlieb:
> Tests for the new utils function still need to be added.

5 unit tests added for `utils::list_s3_paths` inside of `test/test_utils.py`:
- `test_list_s3_paths_success`
- `test_list_s3_paths_with_relative_paths`
- `test_list_s3_paths_with_public_bucket`
- `test_list_s3_paths_command_failure`
- `test_list_s3_paths_with_nested_directory`

api/scpca_portal/models/library.py
```diff
@@ -58,17 +62,52 @@ def get_from_dict(cls, data):
     def bulk_create_from_dicts(cls, library_jsons: List[Dict], sample) -> None:
         libraries = []
         for library_json in library_jsons:
-            libraries.append(Library.get_from_dict(library_json))
+            if not Library.objects.filter(scpca_id=library_json["scpca_library_id"]).exists():
+                libraries.append(Library.get_from_dict(library_json))
```
davidsmejia, line 66:
```suggestion
                # TODO: remove when scpca_project_id is in source json
                library_json["scpca_project_id"] = sample.project.scpca_id
                libraries.append(Library.get_from_dict(library_json))
```
I was thinking we should break up the `get_modality_and_formats` method into two. But then really, at that point we should just be updating the everything from the context of a single Library instance. Let's drop the static methods and place them directly in `get_from_dict`.

I think this is worth putting in there to avoid accidentally creating a Library instance that is in an "incomplete" state.

api/scpca_portal/models/library.py
```diff
@@ -58,17 +62,52 @@ def get_from_dict(cls, data):
...
 
         Library.objects.bulk_create(libraries)
         sample.libraries.add(*libraries)
+        Library.add_data_file_paths(libraries)
+        Library.add_modality_and_formats(libraries)
+        Library.objects.bulk_update(libraries, ["data_file_paths", "formats", "modality"])
+
+    def get_data_file_paths(self) -> List[Path]:
```
davidsmejia, line 74: In conjunction with the other changes this can become a class method.

davidsmejia:
> FYI, something to keep in mind. Hitting aws every time we encounter a new library and want to grab its data file paths may introduce a bottleneck. Performance wise, my testing suite ran roughly 3x slower when comparing this branch to dev.
> 
> `avrohom/731-add-data-file-paths-attribute-to-library-model` image
> 
> `dev` image
> 
> An alternative approach could be:
> 
>     * Hit aws one time per project in the beginning
> 
>     * Grab all of the files paths then and there
> 
>     * Stash them in an `all_file_paths` attribute on the `Project` model
> 
>     * _Then_ query for them as needed when iterating over libraries later on

This is a valid concern but let's address this down the road. I think the move will be adding an entire projects folder paths as an attribute then updating the library logic to just grab the librarys paths from the project's master list. But we can optimize later on after the initial implementation after we see real world performance

api/scpca_portal/test/management/commands/test_load_data.py
```diff
@@ -13,7 +13,7 @@
 ALLOWED_SUBMITTERS = {"scpca"}
 # NOTE: When INPUT_BUCKET_NAME is changed, please delete the contents of
 # api/test_data/input before testing to ensure test files are updated correctly.
-INPUT_BUCKET_NAME = "scpca-portal-public-test-inputs/2024-04-19/"
+INPUT_BUCKET_NAME = common.INPUT_BUCKET_NAME
```
davidsmejia, line 16: Let's do the following:
- update comment above so it captures the following:
  - test data bucket is defined in common.py
  - when that value changes be sure to wipe your local folder to ensure files are updated
- remove the local `INPUT_BUCKET_NAME` constant
- clean up where we explicitly pass `INPUT_BUCKET_NAME` into `load_data` in this file
- in `load_data` (management command file) replace hard-coded `scpca-portal-inputs` with `common.INPUT_BUCKET_NAME`

davidsmejia (commented):
Looking good just some clean up now

api/scpca_portal/models/library.py
```diff
@@ -31,19 +34,20 @@ class Modalities:
...
+    data_file_paths = ArrayField(models.TextField(), default=list)
     formats = ArrayField(models.TextField(choices=FileFormats.CHOICES), default=list)
     is_multiplexed = models.BooleanField(default=False)
     metadata = models.JSONField(default=dict)
     modality = models.TextField(choices=Modalities.CHOICES)
     scpca_id = models.TextField(unique=True)
     workflow_version = models.TextField()
 
+    def __str__(self):
+        return f"Library id {self.scpca_id}"
```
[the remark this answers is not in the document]
avrohomgottlieb: Here is the above mentioned discussion issue: https://github.com/AlexsLemonade/scpca-portal/issues/756

api/scpca_portal/utils.py
```diff
@@ -92,3 +94,60 @@ def get_csv_zipped_values(
...
+
+    # If there are nested directories in the bucket name, remove them from the returned file paths
+    if bucket_nested_dirs := bucket.parts[1:]:
+        removable_prefix = Path().joinpath(*bucket_nested_dirs)
+        file_paths = [path.relative_to(removable_prefix) for path in file_paths]
```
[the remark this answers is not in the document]
davidsmejia: alternatively remove the conditional and just have this to ignore anything stemming from a bucket path

```python
    # return path relative to bucket path
    bucket_folder = Path().joinpath(*bucket.parts[1:])
    return [Path(entry.file_path).relative_to(bucket_folder) for entry in bucket_entries]
```

api/scpca_portal/models/project.py
```diff
@@ -508,19 +513,11 @@ def get_multiplexed_with_mapping(self, multiplexed_libraries_metadata: List[Dict
...
-                if multiplexed_sample_id not in multiplexed_with_mapping:
-                    multiplexed_with_mapping[multiplexed_sample_id] = set()
-                multiplexed_with_mapping[multiplexed_sample_id].update(
-                    multiplexed_library_sample_ids_copy
+        for multiplexed_dir in sorted(Path(self.input_data_path).rglob("*,*/")):
+            multiplexed_ids = multiplexed_dir.name.split(",")
+            for multiplexed_id in multiplexed_ids:
+                multiplexed_with_mapping[multiplexed_id] = set(
+                    id for id in multiplexed_ids if id is not multiplexed_id
                 )
```
davidsmejia, line 521:
```suggestion
            multiplexed_ids = set(multiplexed_dir.name.split(","))
            for id in multiplexed_ids:
                multiplexed_with_mapping[id] = multiplexed_ids.difference(id)                )
```

I think that doing the substring check made this a bit harder to read. what do you think about this approach instead? - also need to remove the trailing `)`

api/scpca_portal/utils.py
```diff
@@ -92,3 +94,66 @@ def get_csv_zipped_values(
...
+    if "public" in str(bucket_path):
+        command_inputs.append("--no-sign-request")
+
+    try:
+        result = subprocess.run(command_inputs, capture_output=True, text=True, check=True)
+        output = result.stdout
+    except subprocess.CalledProcessError as error:
+        logger.error(
+            """
+            `{}`: Cause of error not returned, note: folder must exist and be non-empty
```
davidsmejia, line 143: Can you add the command_inputs to the error message?

davidsmejia (commented):
LGTM some suggestions for improvements but the only one that is actually important is including the command in the error message

api/scpca_portal/utils.py
avrohomgottlieb: I actually found that the `subprocess.CallProcessError` prints the args (which are the `command_inputs`) when the `print()` function (in this case the logger function) is called on it, so it ends up being redundant.

avrohomgottlieb: This is what it prints
```
 `Command '['aws', 's3', 'ls', 's3://scpca-portal-public-test-inputs/2024-04-19/SCPCP999992/SCPCP999992_bulk', '--recursive', '--no-sign-request']' returned non-zero exit status 1.`: Cause of error not returned, note: folder must exist and be non-empty 
```

api/scpca_portal/models/project.py
avrohomgottlieb: This is great! I'll add it.

davidsmejia (approved):
LGMT

Issue #699, opened 2024-05-13
Simplify sorting of columns when writing metadata TSVs


davidsmejia:
### Context

While doing a code review I noticed that we have what seems to be some over complicated sorting preferences when writing output metadata files. This is specifically about ordering the columns and _not_ the rows. This _does not_ influence which columns are present in the output file, only the order in which we write them to the file header.

I am unsure how much of this is actually necessary and if we could simplify this by having a single sort order for all output metadata file columns independent of modality.

Currently we do the following:
- Pass a list of all `columns` that will be written to the output TSV into the `get_metadata_field_names` function.
- This function checks modality and sorts the provided `columns` based on the sort order for that modality.
- Any keys that were not present in the explicit modality sort order are moved to the end of `columns` in alphabetical order.
- The function then returns the same `columns` keys that were sorted.

The sort order for each modality can be found here:
https://github.com/AlexsLemonade/scpca-portal/blob/dev/api/scpca_portal/models/project.py#L770-L855

### Problem or idea

I suppose there are two questions that should be answered here:
- Does modality sufficiently affect the context in which the user would visually view the metadata files. I say visually because programmatically, order doesn't really mean that much.
- If the answer to the previous question is "no", can we create a single sort order that would influence the ordering of all metadata files independent of modality? That way we would always have `project_id`, `sample_id`, `library_id`, `disease` etc.

### Solution or next step

tagging: @dvenprasad @allyhawkins @jashapiro for input.

allyhawkins:
@davidsmejia there are only a few things that are module specific, so I would agree that we could create a master list. I think the most important things for people to see first are the sample information, so I am proposing a new order that groups things with sample information, library information, project information, and then processing specific information. The processing metadata is the only thing that's modality specific. 

Another thing I noticed when looking at this list is that there are quite a few fields that we have in the metadata that aren't accounted for in the code that you sent. I took the current columns from the two metadata files that you had sent me recently that reflect the current contents of the portal. If we are going to set an order, we should incorporate all the columns that should be present. 

The one question I have is how this order reflects the samples table shown on the Portal? I remember we had quite a few conversations about that sample metadata order and wanted the order to be specific for users, so as long as changing the order in the actual file doesn't affect that table then I would update it. 

Here's a proposed order that I think would make sense: 

```
# use these categories for everything as the first columns in the metadata table 
# there is one column name that is specific to multiplexed in this list that is denoted 
Sample.Modalities.ALL: (
[...]

PR #749, opened 2024-06-06
Add demux samples to sort order

davidsmejia:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/699#issuecomment-2153393073

> > I was working on docs and noticed we forgot some columns that should be included. I am updating the order to reflect the change that project ID should be first and then added the new ontology columns.
> 
> It looks like we are missing the `demux_samples` field for multiplexed data. I would add it after `technology` (in the library metadata section) for multiplex samples only.

- Adds `demux_samples` after technology in the column sort order. When present it will no longer be collected in the "catch all" section.

## Types of changes

## Functional tests

N/A

Will need to update tests probably.

## Checklist

## Screenshots

N/A


avrohomgottlieb (approved):
LGTM

PR #757, opened 2024-06-11
make computed file str method safe for no modality

davidsmejia:
## Issue Number

N/A

Hotfix to address the possibility that a computed file may not have a modality when the str dunder method is called. This was preventing uploads of metadata only downloads.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

n/a


avrohomgottlieb (approved):
This looks great 👍

Issue #732, opened 2024-05-31
Create DOWNLOAD_OPTIONS data structure and use it to query relevant libraries for computed file generation


avrohomgottlieb:
### Context

When it comes to computing files, first `Sample::compute_sample_computed_files` is called, which archives together all of a sample's data and metadata files and then uploads the archive to s3. While files are being computed, `Sample::compute_sample_computed_files` also maintains a file mapping associating sample ids with its associated data file paths (which are available locally after being downloaded from s3 earlier on). This file mapping is then consumed by `Project::create_project_computed_files` when archiving its sample data and metadata files.

Issue https://github.com/AlexsLemonade/scpca-portal/issues/731 proposes an intended solution to remove the need for this file mappings data structure by saving all data file paths in an ArrayField on the `Library` model. With this solution, a `Project` (and its `Samples`) will have direct access to its desired data files necessary for archiving.

The way that files are archived at the `Project` and `Sample` level is by checking for different conditions inside of `Project|Sample::computed_project|sample_computed_files`, and then, based on these different checks, a `ComputedFile::get_project|sample_<modality>_file` subroutine is called which is responsible for the actual archiving.

### Problem or idea

The way that we proceed through `Project|Sample::computed_project|sample_computed_files` and check for a variety of different conditions has led to the following suboptimal design decisions:
- For the sake of code clarity, we currently have separate subroutines inside of `ComputedFile` which archive files, all of which share most of their code
- We are forced to explicitly check for all of the different download conditions, and only then call the correct `ComputedFile` subroutine, which contributes to code bloat and can be carried out more succinctly and programmatically

### Solution or next step

- Initialize a new data structure called `DOWNLOAD_OPTIONS` in a new file called `download_options.py` inside of the `config\` directory. This data structure will  include all possible permutations of computed file download options, organized as a list of dictionaries, each dictionary representing a valid configuration of download options.
- Write a new method on the `Library` model which takes a valid download options configuration with which to query the db and return all relevant libraries.
- Add a helper method to the `Library` model which translates a `data_file_path` (which is an s3 path) to a local file path of the data file which locates where it currently resides on disk. This translation helper method will no longer be necessary when we transition to lazy downloading.

avrohomgottlieb:
Instead of putting DOWNLOAD_OPTIONS into `download_options.py`, let's put it at the bottom of `config/common.py`.

davidsmejia:
just going to comment for posterity,

[...]

PR #758, opened 2024-06-11
732 - Add Valid Download Configurations

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/732

This PR accomplishes the following items:
- Add `VALID_PROJECT_DOWNLOAD_CONFIGURATIONS` and `VALID_SAMPLE_DOWNLOAD_CONFIGURATIONS` data structure to `scpca_portal/common.py` which represent all valid project and sample download options configurations.
- Update the `Library` model to add a foreign key relationship with the `Project` model.
  - This is helpful for a variety of reasons but in this case it allows for cleaner query inside of `Libary::get_project_libraries_from_download_config` by enabling a direct query on the passed project object.
- Add `Libary::get_project_libraries_from_download_config` and `Library:get_sample_libraries_from_download_config` methods to the Library model, allowing for quick retrieval of libraries during file computation.
- Add a helper function `Library::get_local_path_from_data_file_path` quick conversion of relative paths (pulled down from s3) to local file system paths 
  - This will specifically be helpful in ascertaining whether a data file exists locally or not.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


api/scpca_portal/models/library.py
```diff
@@ -104,3 +109,33 @@ def get_formats_from_file_paths(cls, file_paths: List[Path]) -> List[str]:
...
+        cls, project, download_configuration: Dict
+    ) -> List[Self]:
+        if download_configuration not in common.VALID_PROJECT_DOWNLOAD_CONFIGURATIONS:
+            raise ValueError("Invalid download configuration passed. Unable to retrieve libraries.")
+
+        return [
+            library
+            for library in project.libraries.filter(modality=download_configuration["modality"])
+            if download_configuration["format"] in library.formats
+        ]
```
davidsmejia, line 124: I think we should add a test for this because `exclude_multiplexed` isn't accounted for yet.

davidsmejia: For merged something like this but you need to check the format requested and then check against the correct project attribute, `includes_merged_anndata` or `includes_merged_sce`. You might want to make an enum for this.
```python
if download_configuration["includes_merged"] and not project.includes_merged_anndata:
    return []
```

api/scpca_portal/models/library.py
```diff
@@ -104,3 +109,33 @@ def get_formats_from_file_paths(cls, file_paths: List[Path]) -> List[str]:
...
+        cls, sample, download_configuration: Dict
+    ) -> List[Self]:
+        if download_configuration not in common.VALID_SAMPLE_DOWNLOAD_CONFIGURATIONS:
+            raise ValueError("Invalid download configuration passed. Unable to retrieve libraries.")
+
+        return [
+            library
+            for library in sample.libraries.filter(modality=download_configuration["modality"])
+            if download_configuration["format"] in library.formats
+        ]
```
davidsmejia, line 137:
```suggestion
        return sample.libraries.filter(
            modality=download_configuration["modality"],
            format__contains=download_configuration["format"]
        )
```

https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/fields/#contains

You can update the others as well.

davidsmejia (commented):
Looks good, just some small changes here

api/scpca_portal/models/library.py
```diff
@@ -104,3 +109,45 @@ def get_formats_from_file_paths(cls, file_paths: List[Path]) -> List[str]:
...
+        cls, project, download_configuration: Dict
+    ) -> List[Self]:
+        if download_configuration not in common.GENERATED_PROJECT_DOWNLOAD_CONFIGURATIONS:
+            raise ValueError("Invalid download configuration passed. Unable to retrieve libraries.")
+
+        if download_configuration["includes_merged"]:
+            # If the download config requests merged and there is no merged file in the project,
+            # return an empty list
+            if (not project.includes_merged_sce) and (not project.includes_merged_anndata):
+                return []
```
davidsmejia, line 124: This is too broad a check, we need to check if the specific merged data is available for the format in the `download_configuration`. So we only check `includes_merged_sce` when `format` is `SINGLE_CELL_EXPERIMENT` and alternatively check `includes_merged_anndata` when `format` is `ANN_DATA`.

This probably wouldn't break anything today but would be a bug should be have a project that is merged in one format but not another.

api/scpca_portal/models/library.py
```diff
@@ -104,3 +109,45 @@ def get_formats_from_file_paths(cls, file_paths: List[Path]) -> List[str]:
...
+
+        if download_configuration["metadata_only"]:
+            return project.libraries.all()
```
davidsmejia, line 127: I don't think there are any potential pitfalls with having this here but I think it would be better placed at the top since it is the least restrictive check. I can see that you put it here because it is not a "guard"and generally those go first as an escape hatch but I think that because it is checked independently of any other values it's clearer up top (after the first validation guard).

api/scpca_portal/models/library.py
davidsmejia, line 124:
```suggestion
                return Project.libraries.none()
```

api/scpca_portal/models/library.py
```diff
@@ -104,3 +107,53 @@ def get_formats_from_file_paths(cls, file_paths: List[Path]) -> List[str]:
...
+            raise ValueError("Invalid download configuration passed. Unable to retrieve libraries.")
+
+        if download_configuration["metadata_only"]:
+            return project.libraries.all()
+
+        if download_configuration["includes_merged"]:
+            # If the download config requests merged and there is no merged file in the project,
+            # return an empty queryset
+            if (
+                download_configuration["format"] == "SINGLE_CELL_EXPERIMENT"
```
davidsmejia, line 125:
```suggestion
                download_configuration["format"] == FileFormats.SINGLE_CELL_EXPERIMENT
```

Let's avoid hardcoding these. I think this is good for now but maybe we should consider having an instance method on `Project` that takes the format and returns the correct `includes_merged_*` attribute value.

avrohomgottlieb: Yeah, that's a good catch. I'll make this update.

api/scpca_portal/test/models/test_library.py
davidsmejia, line 1: I think coverage for these tests look good, but I think we should just be setting them up by creating fake resources. I don't see why we need to create fake methods on these when we can just populate the test database and allow the queries to be performed normally. Let me know if I am missing something.

avrohomgottlieb: Sure, we can go in this direction.

api/scpca_portal/test/models/test_library.py
```diff
@@ -0,0 +1,119 @@
...
+    def test_get_sample_libraries_from_download_config_single_cell_anndata(self):
+        config = {"modality": "SINGLE_CELL", "format": "ANN_DATA"}
+        result = Library.get_sample_libraries_from_download_config(self.sample, config)
+        self.assertIn(self.library1, result)
+        self.assertIn(self.library3, result)
+
+    def test_get_sample_libraries_from_download_config_spatial_sce(self):
+        config = {"modality": "SPATIAL", "format": "SINGLE_CELL_EXPERIMENT"}
+        result = Library.get_sample_libraries_from_download_config(self.sample, config)
+        self.assertIn(self.library4, result)
```
davidsmejia, line 114: Lets combine these into a single test and iterate over the actual common. we have defined. I would iterate over each of the configs and then check the modality and format in the download_configuration and check that the appropriate samples are there.

api/scpca_portal/test/models/test_library.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+        self.library3 = LibraryFactory(project=self.project, modality=Library.Modalities.SPATIAL)
+
+        self.default_config = {
+            "modality": "SINGLE_CELL",
+            "format": "SINGLE_CELL_EXPERIMENT",
+            "excludes_multiplexed": False,
+            "includes_merged": False,
+            "metadata_only": False,
+        }
```
davidsmejia, line 24: based on our conversation, lets rename this to `download_options` and mutate it directly in each test w/o copying first.

davidsmejia (commented):
Looks good just two comments on the tests

davidsmejia (approved):
LGTM 

I don't think we *need* to take care of this in this PR but the more I look at this the more it looks like `get_project_libraries_from_download_config` should go on the `Project` model and the same for `Sample`.

I think the idea of keeping it in `Library` is mainly for a future benefit of handling user requested downloads uniformly. So either we would move it to the respective model now, or keep it here and in the future we will likely end up combining them and writing `get_libraries_for_config` in a way that would be agnostic of which related model is being used to query `libraries`. This is possible because `Sample` download options is a subset of `Project` download options.

I don't think we need to answer these questions today and can just as easily refactor/move the class methods it if it makes sense once the rest of this epic is implemented.

Issue #736, opened 2024-06-03
Add the metadata download on project level


nozomione:
### Context
Design handoff: https://github.com/AlexsLemonade/scpca-portal/issues/701

On the portal, we are going to provide users with a metadata-only download option by adding buttons to the project card and the samples table. 

During the file download, the download modal UI should display metadata-only download content.

### Problem or idea
Based on the design, we'll implement metadata-only download buttons on projects and display metadata-only download content in the download modal.

### Solution or next step
- Create a new component for the download sample metadata button
- Add this new component to:
  -  `ProjectSearchResult`
  -  `ProjectSamplesTable`
- Render the metadata-only download content to `DownloadStarted`

davidsmejia:
Looking at the Comps, I don't think we need to change to say `Sample` everywhere so let's only address appending `Metadata` to the title and leave the rest as it will render already.

Suggestion for next steps:

### Solution or next step
- Create new component `MetadataDownloadModal.js`
  - accepts the following props: `project`, `disabled`
  - wrap the existing `DownlodModal` component and provide the computed file for the computed file that matches on`metadata_only`
- `DownloadModal.js`
  - update to support checking for `metadata_only` to change the title
- `DownloadStarted.js
  - Check `metadata_only` on computed file to only show the metadata line.
- `Table.js`
  - Add prop `onFilterChange` defaults to noop
  - extract `globalFilter` renamed `currentFilter` where pageIndex is extracted
  - add useEffect that watches currentFilter and calls `onFilterChange`
- Add the download sample metadata button to:
  -  `ProjectSearchResult`
  -  `ProjectSamplesTable`
    - disabled when filter is not-empty (using new callback)

davidsmejia:
Metadata only download is available on staging. Note that modality and format may be undefined which is something you might need to account for when rendering contents.

```
        {
            "created_at": "2024-06-11T13:57:23.470755Z",
            "format": null,
            "has_bulk_rna_seq": false,
            "has_cite_seq_data": false,
            "has_multiplexed_data": false,
            "id": 21011,
            "includes_merged": false,
            "metadata_only": true,
            "modality": null,
            "project": "SCPCP000001",
            "s3_bucket": "scpca-portal-deployer-staging",
            "s3_key": "SCPCP000001_all_metadata.zip",
            "sample": null,
            "size_in_bytes": 21528,
            "updated_at": "2024-06-11T13:57:23.470782Z",
            "workflow_version": "v0.8.0"
[...]

PR #761, opened 2024-06-11
736 - Add the metadata download on project level

nozomione:
## Issue Number

Closing #736

The metadata only download button on project level has been added to:
- the project card on the search results page.
- the project samples table.

@dvenprasad  Please review the latest UI [here](https://scpca-portal-das475b7m-ccdl.vercel.app/).

## Types of changes

## Checklist

## Screenshots


nozomione:
@davidsmejia , thank you for updating the base branch! Does this mean that we no longer use the `feature/client-metadata-downloads` branch for metadata only download changes for FE?

davidsmejia (commented):
Some initial notes here, I believe I will have more feedback but let's start with these items.

client/src/hooks/useDownloadOptionsContext.js
```diff
@@ -42,9 +42,10 @@ export const useDownloadOptionsContext = () => {
       uniqueArray(pick(files, optionName)),
       optionsSortOrder
     )
-    const defaultOption = allOptions.includes(preference)
+    const filteredAllOptions = allOptions.filter((o) => o) // Exclude the metadata_only file
```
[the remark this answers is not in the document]
nozomione: See the commit at acfbb471eb9140283bfe2396a483201c85d21a81.

client/src/hooks/useMetadataOnly.js
```diff
@@ -0,0 +1,21 @@
...
+export const useMetadataOnly = (computedFiles) => {
+  // Check the availability of metadata_only file
+  const isMetadataOnlyAvailable =
+    filterWhere(computedFiles, {
+      metadata_only: true
+    }).length > 0
+
+  const metadataComputedFile = filterWhere(computedFiles, {
+    metadata_only: true
+  })
```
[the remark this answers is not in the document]
nozomione: See the commit at 936829fab508e63736b7d5cb97db3f92f7e49e9e.

client/src/components/DownloadModal.js
```diff
@@ -45,7 +51,17 @@ export const DownloadModal = ({
...
   return (
     <>
-      {icon ? (
+      {icon && sampleMetadataOnly ? (
+        <Anchor
+          icon={icon}
+          onClick={handleClick}
+          disabled={disabled}
+          label={
+            sampleMetadataOnly ? <Text color="brand">{buttonLabel}</Text> : ''
```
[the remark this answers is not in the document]
nozomione: Yes, this is always true. This is to prevent the `buttonLabel` from being rendered on the download sample icon button in the project samples table (pls see the screenshot below).

Screenshot 2024-06-13 at 1 21 56 PM

client/src/components/DownloadModal.js
```diff
@@ -45,7 +51,17 @@ export const DownloadModal = ({
 
   return (
     <>
-      {icon ? (
+      {icon && sampleMetadataOnly ? (
```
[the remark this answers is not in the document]
nozomione: See the commit 43a94f814d4d48c4445abb93cc2deea3da4a150f.

client/src/components/DownloadModal.js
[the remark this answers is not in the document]
nozomione: Currently, the rendering of the metadata-only download button label ("Download Sample Metadata") depends on whether the button is for a metadata-only download, not the availability of a metadata download file.

If the metadata download file is available: 
Screenshot 2024-06-13 at 1 44 01 PM

if it's unavailable:
Screenshot 2024-06-13 at 1 45 19 PM

Unless we hide the metadata-only download button when the metadata download file is not available?

Please let me know your input on this!

(See commit at 43a94f814d4d48c4445abb93cc2deea3da4a150f)

client/src/components/DownloadModal.js
```diff
@@ -45,7 +51,17 @@ export const DownloadModal = ({
...
+        <Anchor
+          icon={icon}
+          onClick={handleClick}
+          disabled={disabled}
+          label={
+            sampleMetadataOnly ? <Text color="brand">{buttonLabel}</Text> : ''
+          }
+          margin={{ top: 'small' }}
```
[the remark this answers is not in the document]
nozomione: See update at 43a94f814d4d48c4445abb93cc2deea3da4a150f.

nozomione:
I've updated the PR based on your feedback and is ready for another look, thank you David!

client/src/components/DownloadModal.js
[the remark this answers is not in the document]
davidsmejia: That isn't quite the point I was trying to make.

In use modal this PR introduces:
```javascript
  const buttonLabel = `Download ${
    sampleMetadataOnly ? 'Sample Metadata' : resourceType
  }`
```

which is functionally equivalent to:
```javascript
  const buttonLabel = `Download ${
    initialPublicComputedFile.metadata_only ? 'Sample Metadata' : resourceType
  }`
```

I think however that this confusion illustrates that the current usage of the `DownloadModal` has drifted from the original plan of "give it a resource and let it figure out how to render itself". So, I think at this point we should make the following changes.

- In the `DownloadModal` component add the props: `color`, `label`. We will pass this in wherever we use the component instead of trying to evaluate it based on the resource or computedFile that is passed in.
- In `useDownloadModal` let's remove the variable `buttonTitle` and add to the `modalTitle` at the end if it is metadata.
- Delete the `MetadataDownloadModal` component, it really only serves as a place to call the hook that gets the metadata file and returns if the file is available. We can do that in `ProjectSearchResult` and `ProjectSamplesTable` components when determining how to render the button.
- Remove all arguments of `sampleMetadataOnly` for components and hooks, we want to determine state based off of the resource and the computedFile or initialComputedFile.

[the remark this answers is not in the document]
nozomione: The separation of the UI presentation logic from the `useDownloadModal` hook is a great idea! Since it eliminates the need for us to conditionally determine the button label's text (especially now that we have more than two resource types).

I applied your input in the following commits: 
> In the `DownloadModal` component add the props: `color`, `label`. We will pass this in wherever we use the component instead of trying to evaluate it based on the resource or `computedFile` that is passed in.

See the commit at c1c929b32cb6c1beaae3c838972ed7f8cd158270.

> In `useDownloadModal` let's remove the variable `buttonTitle` and add to the `modalTitle` at the end if it is metadata.

See the commits at 93068098cc14591b444d62b42ec42de3b67d139c and ffda1378cf771575e97643d3c10447d0fa9a8499.

**NOTE:** Instead of using `initialPublicComputedFile`, `publicComputedFie` is used to evaluate `isSampleMetadataOnly`, since after removing ‘buttonLabel’, this evaluation is only necessary for the downloading modal.

> Delete the `MetadataDownloadModal` component, it really only serves as a place to call the hook that gets the metadata file and returns if the file is available. We can do that in `ProjectSearchResult` and `ProjectSamplesTable` components when determining how to render the button.

See the commitd at dfd55f0a5bb8c2e687a685f620b8469a768971fb and 967baffa6d7e5986afdc65f6a69d0637667001d3.

**NOTE:** In this commit, I've also adjusted the Project Download button (in `ProjectHeader.js`), since `buttonLabel` was removed.

> Remove all arguments of `sampleMetadataOnly` for components and hooks, we want to determine state based off of the resource and the `computedFile` or `initialComputedFile`.

They were completed in above commits, and yes, I used `computedFie` 👍 !

client/src/components/DownloadModal.js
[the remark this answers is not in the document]
davidsmejia: This is a comment on line 55, so remove the margin from inside the component and make it the responsibility of the parent component to space it correctly.

client/src/components/DownloadStarted.js
```diff
@@ -134,7 +134,7 @@ export const DownloadStarted = ({
               </Box>
             </WarningText>
           )}
-          {additionalRestrictions && (
+          {!isSampleMetadata && additionalRestrictions && (
```
davidsmejia, line 137:
```suggestion
          {additionalRestrictions && (
```

Additional restrictions should still be displayed here.

client/src/components/DownloadModal.js
```diff
@@ -11,16 +11,17 @@ import { useDownloadModal } from 'hooks/useDownloadModal'
...
 // Button and Modal to show when downloading
 export const DownloadModal = ({
-  icon,
+  label,
   resource: initialResource,
-  publicComputedFile: initialPublicComputedFile
+  publicComputedFile: initialPublicComputedFile,
+  icon = null,
+  color = 'brand',
+  disabled = false
```
davidsmejia, line 19: We need to update how we handle `disabled` so that they are the same no matter if we render the "Anchor" version or the "Button" version.

I think this should be enough, and then just use `isDisabled` for both versions.
```javascript
const isDisabled = disabled || publicComputedFile === undefined
```

Or remove `publicComputedFile === undefined` and pass in the disabled attribute manually.

client/src/components/ProjectSamplesTable.js
```diff
@@ -23,15 +25,21 @@ export const ProjectSamplesTable = ({
   // We only want to show the applied donwload options.
   // Also need some helpers for presentation.
   const { modality, format, getFoundFile } = useDownloadOptionsContext()
+  const { metadataComputedFile } = useMetadataOnly(project.computed_files)
   const [loaded, setLoaded] = useState(false)
   const [samples, setSamples] = useState(defaultSamples)
   const [showDownloadOptions, setShowDownloadOptions] = useState(false)
+  const [currentFilter, setCurrentFilter] = useState('') // For storing globalFiter for metadata only download
```
davidsmejia, line 32: I think it would be clearer to instead of using the truthiness of the current filter to just have a boolean value.

```suggestion
  const [hasFilter, setHasFilter] = useState(false)
```

client/src/components/ProjectSamplesTable.js
```diff
@@ -23,15 +25,21 @@ export const ProjectSamplesTable = ({
...
   const hasMultiplexedData = project.has_multiplexed_data
   const infoText =
     project && project.has_bulk_rna_seq
       ? 'Bulk RNA-seq data available only when you download the entire project'
       : false
 
+  const onFilterChange = (value) => {
+    setCurrentFilter(value)
```
davidsmejia, line 40:
```suggestion
    setHasFilter(value !== '')
```

client/src/components/DownloadModal.js
```diff
@@ -11,16 +11,17 @@ import { useDownloadModal } from 'hooks/useDownloadModal'
 
 // Button and Modal to show when downloading
 export const DownloadModal = ({
-  icon,
+  label,
   resource: initialResource,
-  publicComputedFile: initialPublicComputedFile
+  publicComputedFile: initialPublicComputedFile,
+  icon = null,
+  color = 'brand',
```
davidsmejia, line 18: So after looking at how we use color and label. I think we should remove color and just have the caller do something like `<DownloadModal label={<Text color="brand">My Label</Text>} />` when they want to override it? This is specifically because it leads to weird behavior if you pass a color and a label that specified a color, the `color` attribute wouldn't affect the label.

davidsmejia (commented):
Looking good, just a few more things to adjust

client/src/components/DownloadModal.js
```diff
@@ -61,7 +68,7 @@ export const DownloadModal = ({
         {hasDownloadOptions && (
           <ModalHeader>
             <Text
-              color="brand"
+              color={color}
```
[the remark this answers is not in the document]
nozomione: See the commit at 50ec00cf5370b93ebaf309789055804ad0e75259.

client/src/components/DownloadModal.js
nozomione: See the commit at 50ec00cf5370b93ebaf309789055804ad0e75259.

client/src/components/ProjectSamplesTable.js
nozomione: Good call 👍!  See the commit at fc9f72d11bca05ea4211ef131cacc3fa3233113a.

client/src/components/ProjectSamplesTable.js
nozomione: See the commit at fc9f72d11bca05ea4211ef131cacc3fa3233113a.

client/src/components/DownloadModal.js
nozomione: Great! See the commit at ea94de48055d089a6f33d49d5a0174dc18dd89bc.

client/src/components/DownloadStarted.js
nozomione: See the commit at d9aa9e95fcbeef55ae961c684749f3d19cbf5bbe.

nozomione:
I've applied your input and this PR is ready for your review, thank you David!

client/src/helpers/getDownloadOptionDetails.js
```diff
@@ -28,8 +29,15 @@ const formatFileItemByKey = (key, computedFile) => {
...
-  const type = project ? 'Project' : 'Sample'
+  const {
+    metadata_only: metadataOnly,
+    modality,
+    project,
+    sample
+  } = computedFile
+  // eslint-disable-next-line no-nested-ternary
+  const type = metadataOnly ? 'Sample Metadata' : project ? 'Project' : 'Sample'
+  const isSampleMetadata = type === 'Sample Metadata'
```
davidsmejia, line 40: Can we remove `isSampleMetadata` or rename `metadataOnly` since they are (and should always be) the same value?

client/src/hooks/useMetadataOnly.js
```diff
@@ -0,0 +1,19 @@
+import { filterWhere } from 'helpers/filterWhere'
+
+// TODO: Implement the logic to handle multiple metadata files (i.e. project and portal-wide) later
+export const useMetadataOnly = (computedFiles) => {
+  const metadataComputedFile = filterWhere(computedFiles, {
```
davidsmejia, line 5:
```suggestion
  const metadataComputedFiles = filterWhere(computedFiles, {
```

Since there are multiple we should have the variable name reflect that.

client/src/hooks/useMetadataOnly.js
```diff
@@ -0,0 +1,19 @@
...
+    metadata_only: true
+  })
+  // Check the availability of metadata_only file
+  const isMetadataOnlyAvailable = metadataComputedFile.length > 0
+
+  // NOTE: The current logic only supports the project level metadata file,
+  // thus the first item is returned
+  return {
+    metadataComputedFile: metadataComputedFile[0],
```
davidsmejia, line 14:
```suggestion
  const isMetadataOnlyAvailable = metadataComputedFiles.length > 0

  // NOTE: The current logic only supports the project level metadata file,
  // thus the first item is returned
  return {
    metadataComputedFile: metadataComputedFiles[0],
```

And then use the new name

client/src/hooks/useDownloadModal.js
```diff
@@ -18,22 +18,23 @@ export const useDownloadModal = (
...
-  const buttonLabel = `Download ${resourceType}`
-
+  // eslint-disable-next-line no-nested-ternary
+  const resourceType = isSampleMetadataOnly
+    ? 'Sample'
+    : resource.samples
+    ? 'Project'
+    : 'Sample'
+  const metadata = isSampleMetadataOnly ? 'Metadata' : ''
+  const modalTitle = `${verb} ${resourceType} ${metadata}`
```
davidsmejia, line 36:
```suggestion
  const resourceType = resource.samples
    ? 'Project'
    : 'Sample'
  const modalTitle = isSampleMetadataOnly
    ? `${verb} Sample Metadata`
    : `${verb} ${resourceType}`
```

I think breaking it up this way is a bit easier to follow.

davidsmejia (commented):
Thanks for updating that, just one last pass of changes and then this looks good to me

client/src/helpers/getDownloadOptionDetails.js
davidsmejia: This is to say do *one* of the following:
- remove `isSampleMetadata` and use `metadataOnly` here
*or*
- rename `metadataOnly` to `isSampleMetadata`

client/src/hooks/useMetadataOnly.js
nozomione: See the commit at 06be077ebea7c8e7c84ac5e5d90f71a377cac85a.

client/src/hooks/useMetadataOnly.js
nozomione: See the commit at 06be077ebea7c8e7c84ac5e5d90f71a377cac85a.

client/src/hooks/useDownloadModal.js
nozomione: See the commit at c080be77ed9adface1fdaa0b722f365ee8ab6f4e.

client/src/helpers/getDownloadOptionDetails.js
nozomione: I've removed `isSampleMetadata` and use  `metadataOnly`, see the commit at ba188e8760d703c7c361fcad3b6a0750ae685411.

client/src/config/downloadOptions.js
```diff
@@ -114,6 +114,14 @@ export const modalityResourceInfo = {
     }
   }
 }
-
+// For metadata only download
```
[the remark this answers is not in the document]
nozomione: See the commit at ba188e8760d703c7c361fcad3b6a0750ae685411.

nozomione:
Applied all of the input, thank you David!

client/src/contexts/DownloadOptionsContext.js
```diff
@@ -34,7 +34,8 @@ export const DownloadOptionsContextProvider = ({
   useEffect(() => {
     if (resource) {
       setComputedFiles(() => {
-        if (!resourceAttribute) return resource.computed_files
+        if (!resourceAttribute)
+          return resource.computed_files.filter((f) => f.modality) // Exclude the metadata_only file
```
davidsmejia, line 38:
```suggestion
          return resource.computed_files.filter((f) => !f.metadata_only)
```

One last change to make here, since we want to omit computed files where `metadata_only` is `true` we should just do this instead of checking the truthiness of `modality`.

davidsmejia: You can put a comment here saying that modality and format would be empty when this is true if you find that helpful, but I think this change makes the implication more obvious.

nozomione: I've also added the comment you suggested, see the commit at 683d3378f21087f42fb6efe96b8dd418573b326d.

davidsmejia (commented):
here are the notes as per our 1:1

client/src/helpers/getDownloadOptionDetails.js
```diff
@@ -28,8 +29,14 @@ const formatFileItemByKey = (key, computedFile) => {
...
+  const {
+    metadata_only: metadataOnly,
+    modality,
+    project,
+    sample
+  } = computedFile
+  // eslint-disable-next-line no-nested-ternary
+  const type = metadataOnly ? 'Sample Metadata' : project ? 'Project' : 'Sample'
   const isProject = type === 'Project'
   const isSample = type === 'Sample'
```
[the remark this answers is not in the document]
nozomione: See the commit at 86af84f10a620b92f62725ed0b5893ee36412679.

client/src/helpers/getDownloadOptionDetails.js
```diff
@@ -28,8 +29,14 @@ const formatFileItemByKey = (key, computedFile) => {
...
-  const { modality, project, sample } = computedFile
-  const type = project ? 'Project' : 'Sample'
+  const {
+    metadata_only: metadataOnly,
+    modality,
+    project,
+    sample
+  } = computedFile
+  // eslint-disable-next-line no-nested-ternary
+  const type = metadataOnly ? 'Sample Metadata' : project ? 'Project' : 'Sample'
```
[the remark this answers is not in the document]
nozomione: See the commit at 86af84f10a620b92f62725ed0b5893ee36412679.

nozomione:
I've applied your input and this PR is ready for another look, thank you David!

davidsmejia (approved):
LGTM thank you for the revisions

Issue #738, opened 2024-06-03
Project level metadata downloads


davidsmejia:
### Context

We want to add project level metadata downloads before refactoring how load data works.

### Problem or idea

This should be as simple as extending the existing functionality to support the creation on one additional metadata file. As well as adding `metadata_only` attribute on `ComputedFile`. The new attribute will be used to determine the contents of the computed file, as well for future filtering.

### Solution or next step
You will have to bounce back and forth between the files to implement this but here is a list of the changes expected:
- `models/computed_file.py`
  - add `metadata_only` attribute to model, boolean, defaults False
  - add `README_METADATA_NAME`
  - add `README_METADATA_PATH`
  - add `README_TEMPLATE_METADATA_PATH`
  - add `get_project_metadata_file`
    - copies over the written metadata (path is property described below)
    - copies new readme (path is ComputedFile.README_METADATA_PATH)
    - set `metadata_only` to true
    - other attributes set as needed
- `models/project.py`
  - add create_metadata_readme_file
  - call new function in `create_readmes`
  - add `@property def output_all_metadata_file_name`
    - return`common.OUTPUT_DATA_PATH / f"{self.scpca_id}_all_metadata.tsv"`
  - at end of `write_combined_metadata_libraries` (outside of loop)
    - create new variable `project_metadata` and make it the union of `combined_metadata` values
    - write dict to `self.output_metadata_file_name`
  - at end of with ThreadPoolExecutor block
    - call get_project_metadata_file
    - rest of flow is similar enough (uploading etc)

avrohomgottlieb:
Addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/746

PR #764, opened 2024-06-14
Project Metadata Download Feedback

davidsmejia:
## Issue Number

#738 

- The readme template for metadata only downloads didn't handle the list of projects correctly when a single project was being created, this fixes that by putting it in a loop (that iterates one time).
- `computed_file.metadata_file_name` was returning `None` when the modality wasn't set. This was addressed by having a default of `metadata.tsv` so we don't end up copying the entire source path into the zip file.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


avrohomgottlieb (approved):
LGTM

Issue #739, opened 2024-06-03
Add method on the Library model to prep combined metadata for writing


avrohomgottlieb:
### Context

After `Sample` and `Library` objects are created and saved to the db, file archiving can begin. In addition to data files which are added to the archive, a metadata file is also added to every archive. This metadata file is comprised of a combination of `Sample` and `Library` metadata, all extracted from `Sample` and `Library` objects.

### Problem or idea

With the addition of the new `Library` model, there is no longer a need to produce a `combined_metadata` data structure and to keep it around until the times comes to write it to a metadata file. Instead, metadata can be combined lazily, when the time comes to start writing.

### Solution or next step

The solution is to query the necessary fields in the `Library` and `Sample` models for the combining of the requested metadata, and to repeat the process depending on the amount of desired libraries (taking the multiplexed option into account). Then, the resulting list of dictionaries will be directly inputted into `metadata_file::write_metadata_dicts`.

Below represents a possible way of addressing metadata combining, taking multiplexed into account:

```python
class Project:
   def get_metadata(self):
       return {
           "project_scpca_id": self.scpca_id,
       }

class Sample:
   def get_metadata(self):
       return {
           "sample_scpca_id": self.scpca_id,
       }
       
class Library:
   def get_metadata() -> Dict:
       if self.is_multiplexed:
          # multiplexed specific presentation
       return # default presentation
            

   def get_combined_metadata(self, exclude_multiplexed=False) -> List[Dict]:
       metadatas = []
       for sample in self.samples:
           metadata = self.project.get_metadata()
               .update(sample.get_metadata())
               .update({
                   ...
               })
```

avrohomgottlieb:
Addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/765

Issue #763, opened 2024-06-14
Correct Name of Project Metadata Attribute - Submitter or Pi-Name


avrohomgottlieb:
### Context

Inside of `METADATA_COLUMN_SORT_ORDER` in `common.py`, we have the metadata properties `submitter` and `submitter_id` listed under `Sample metadata` keys, while we have `pi_name` listed under `Project metadata` keys.

### Problem or idea

In `metadata_file.py`, while transforming metadata keys, we rename `submitter` to `pi_name` (located inside of `PROJECT_METADATA_KEYS`).

Which one do we want to use? As it currently stands, we're sending both keys through to the output `combined_metadata` file with identical values.

davidsmejia:
Calling @allyhawkins for input

allyhawkins:
I think I would keep `pi_name` and remove `submitter`, since that naming is more informative to users.

avrohomgottlieb:
@allyhawkins  We can definitely do that. 

The global sort order that was okayed for metadata columns was as follows:
```
METADATA_COLUMN_SORT_ORDER = [
    "scpca_project_id",
    # Sample metadata
    "scpca_sample_id",
    "scpca_library_id",
    "diagnosis",
    "subdiagnosis",
    "disease_timing",
    "age_at_diagnosis",
    "sex",
    "tissue_location",
    "participant_id",
    "submitter",
    "submitter_id",
    ...
    # Library metadata
    ...
    # Project metadata
    "pi_name",
    "project_title",
    # Processing metadata
    ...
```

If we delete `submitter` and stick with `pi_name`, where should we put `pi_name` in the order here? It should probably be next to `submitter_id`, right? I guess the suggestions are:
- moving up `pi_name` into the location of `submitter` (in which case, should we keep `project_title` in its current location?)
- bringing down `submitter_id` to be next to `pi_name`
- something different

Thoughts?

allyhawkins:
So `submitter_id` is actually sample specific, so we want to keep that where it is. I think you could just remove `submitter` altogether and keep `pi_name` where it is, next to `project_title`. That would keep things as sample, library, project metadata. Let me know if that is unclear!

avrohomgottlieb:
Okay, then that's what we'll do. Makes perfect sense, thanks for being clear. Thanks Ally!

avrohomgottlieb:
[...]

PR #765, opened 2024-06-17
739 - Combine Library Metadata

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/739

- Add `get_metadata` methods to `Project`, `Sample` and `Library` models which collect all metadata fields relevant to each model, whether in the model's general attributes or in the models `(additional_)metadata` attribute, and returns then in a dictionary
- Add `Library::get_combined_library_metadata` method which combines all of these dictionaries on a per library basis and returns a list of a single dictionary (or of multiple dictionaries if the library is multiplexed).

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


api/scpca_portal/models/sample.py
```diff
@@ -120,6 +120,33 @@ def bulk_create_from_dicts(
...
+        sample_metadata.update(
+            {key: getattr(self, key) for key in dict(self) if key in included_sample_attributes}
+        )
+
+        sample_metadata.update(
+            {key: self.additional_metadata[key] for key in self.additional_metadata}
+        )
+
+        return sample_metadata
+
```
davidsmejia, line 149: When you write a ticket about updating how we handle sample metdata / moving away from `sample.additional_metadata` and towards `sample.metadata`. Include this method as a reason why.

Just as a reminder we want to move toward thinking about these attribtues as one of the following:
- metadata (raw inputs from s3)
- attributes (denormalized metadata keys)
- aggregate values (populated after we do bulk create of a related model)

Then we can think about this data in two contexts: persistence / querying and presentation.

This function should be excluded keys (if any) in the same way we do that for `Library.get_metadata`

davidsmejia (commented):
Looks good to me, just the one change for comment and where we remove the `submitter` from the metadata_file module and remove it where we *currently* remove project metadata keys.

Otherwise I'm just requesting that you open a ticket to remind us to handle sample and project (additional) metadata uniformly as outlined in my comment.

After those items are complete this looks good to merge!

avrohomgottlieb:
As referenced in Issue https://github.com/AlexsLemonade/scpca-portal/issues/763, I've added support for `filtered_spots`, `unfiltered_spots`, and `tissue_spots` to be included in the output metadata tsv file. This will affect only `Spatial` libraries. As a note, I did not adjust the old (current) implementation, which is on its way out, but removed these keys from the `excluded_metadata_attributes` list in the `Library::get_metadata` method which will include them in the output metadata tsv file in the new implementation.

api/scpca_portal/models/sample.py
avrohomgottlieb: Here is the described issue: https://github.com/AlexsLemonade/scpca-portal/issues/774

api/scpca_portal/models/project.py
```diff
@@ -786,6 +793,10 @@ def load_samples_metadata(self, sample_id: str = None) -> List[Dict]:
 
             self.add_project_metadata(sample_metadata)
 
+            # This is a temporary fix until we get rid of `load_samples_metadata` imminently.
```
davidsmejia, line 796:
```suggestion
            # TODO: This is a temporary fix until we get rid of `load_samples_metadata` imminently.
```

davidsmejia (approved):
LGTM just one suggestion selfishly so I will see this highlighted.

PR #770, opened 2024-06-20
Production Deploy

davidsmejia:
# Production Deploy

### API
- Adds support to API for generating metadata only downloads when loading a project
- Includes additional support for upcoming `Library` Model 

### Client
- Adds support on the client for project specific metadata only downloads


nozomione (approved):
👍

avrohomgottlieb (approved):
LGTM

PR #772, opened 2024-06-20
Production Deploy

davidsmejia:
# Production Deploy

### Client
- Bugfix: when there is no metadata only download for a project, we now correctly disable the button on the samples table.


nozomione (approved):
👍

Issue #735, opened 2024-06-03
Add generic ComputedFile.get_project|sample_file methods


avrohomgottlieb:
### Context

At the conclusion of Issue https://github.com/AlexsLemonade/scpca-portal/issues/731, 
> which handles the adding of a `data_file_path` attribute to the `Library` model

and thereafter Issue https://github.com/AlexsLemonade/scpca-portal/issues/732,
> which handles the implementation of a new `DOWNLOAD_OPTIONS` data structure as a means of querying the requested `Library` objects for use in creating computed files

the `ComputedFile::get_project_<modality>_file` and `ComputedFile::get_sample_<modality>_file` methods, which handle the actual file archiving, can be refactored accordingly. 

### Problem or idea

With the ability to adequately collect all libraries which satisfy a certain download option, the writing to an archive of each library's associated data and metadata files will be trivial. At this point, there will no longer be a need for the variety of different methods inside of the `ComputedFile` model which independently carry out file computation according to a diverse set of valid download option permutations. Instead, much of this redundant logic can be consolidated into two generic methods called `ComputedFile::get_project_file` and `ComputedFile::get_sample_file`.

### Solution or next step

Here is a possible implementation of `ComputedFile::get_project_file` (`ComputedFile::get_sample_file` would look similar):

```python
@classmethod
def get_project_file(cls, project, download_option, computed_file_name):
    
    libraries = Library.from_download_options(project, download_option)
    libraries_metadata = [library.get_combined_metadata() for library in libraries]
    library_data_files = [library.get_files(download_option) for library in libraries]
    project_data_files = project.get_files(download_option)

    utils.write_metadata_dicts(libraries_metadata, file_path)
    
    # If the query returns empty, abort early
    if not libraries.exists():
        return
        
    with ZipFile(computed_file.zip_file_path, "w") as zip_file:
        # for now keep zipfile.write, but later we should investigate writing to a  string
        zip_file.write(self.get_readme, "readme.md")
        #zip_file.writestr(
        #    self.get_readme(download_options),
        #    "README.md"
        #)
        
        zip_file.writestr(
            file_path, metadata_file_path    
        )
        if not download_option.get("metadata_only", False):

            # this is for library data
            # would return empty if is_merged merged files would come from additonal_data
            if not download_options["is_merged"]
                for local_path, name in library_data_files:
                    zip_file.write(local_path, name)
                
            # this is for bulk and quant and merged
            for local_path, name in project_data_files:
[...]

PR #773, opened 2024-06-21
735 - ComputedFile Project and Sample Generic Methods

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/735

Added `ComputedFile::get_project_file` and `ComputedFile:get_sample_file` generic methods which do the following:
- Take in a `Project`/`Sample` object, a `download_configuration` and a `computed_file_name`
- Write a libraries metadata file
- Compute a zip archive with all data, metadata and readme files
- Create, save, and return a `ComputedFile` object according to the make up of the computed file archive

Added a variety of helper methods for these two methods, including:
- `ComputedFile::get_readme_download_config`
  - Which takes in a `download_configuration` and returns the name of the correct `readme` file path
- `Library::get_filtered_data_file_paths`
  - Which takes in a `download_configuration` and returns all of a library's `data_file_paths` according to the `download_configuration`
- `Project::get_filtered_data_file_paths`
  - Which takes in a `download_configuration` and returns all of a project's `data_file_paths` according to the `download_configuration`

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


api/scpca_portal/models/library.py
```diff
@@ -183,3 +183,16 @@ def get_combined_library_metadata(self) -> List[Dict]:
...
+            common.ANNDATA_EXT
+            if download_config["format"] == Library.FileFormats.SINGLE_CELL_EXPERIMENT
+            else common.SCE_EXT
+        )
+
+        return [
+            file_path
+            for file_path in [Path(fp) for fp in self.data_file_paths]
+            if file_path.suffix != file_extension_to_be_excluded
+        ]
```
davidsmejia, line 198: Just a few suggestions for this:

- rename to `get_download_config_file_paths` so we can specify why subset of filepaths are being returned
- if we keep this implementation I would recommend using `omit_suffix` instead of `file_extension_to_be_excluded`

Maybe we should consider putting the Format names and extensions relationship in `common.py`.

```python
# common.py
# Formats
ANN_DATA = "ANN_DATA"
SINGLE_CELL_EXPERIMENT = "SINGLE_CELL_EXPERIMENT"

FORMAT_EXTENSIONS = {
    ANN_DATA: ".h5ad",
    SINGLE_CELL_EXPERIMENT: ".rds"
}
```

Then we could do something like the following which would allow us to avoid the "if this then the other one" flavor 
of logic in favor of using sets.

```python

    def get_filtered_data_file_paths(self, download_config: Dict) -> List[Path]:
    
        omit_suffixes = set(common.FORMAT_EXTENSIONS.values())
        omit_suffixes.remove(common.FORMAT_EXTENSIONS[download_config["format"]])
        
        return [
            file_path
            for file_path in [Path(fp) for fp in self.data_file_paths]
            if file_path.suffix not in omit_suffixes
        ]
```

Additionally, I don't think that any of the models should "own" the definition of the `ANN_DATA` constant and I'd prefer not to have it defined in more than one place if possible. This does seem to open up more discussion than we strictly need to address here but I think we should consider it before we use `common.ANNDATA_EXT` more broadly.

api/scpca_portal/models/project.py
```diff
@@ -1159,3 +1159,13 @@ def update_downloadable_sample_count(self):
         self.downloadable_sample_count = downloadable_sample_count
         self.unavailable_samples_count = unavailable_samples_count
         self.save()
+
+    def get_filtered_data_file_paths(self, download_config: Dict) -> List[Path]:
```
davidsmejia, line 1163: We support merged data in more than one format so we still need to filter on that here as well.

api/scpca_portal/models/project.py
```diff
@@ -1159,3 +1159,13 @@ def update_downloadable_sample_count(self):
...
+        if download_config["includes_merged"]:
+            return [Path(fp) for fp in self.data_file_paths]
+
+        return [
+            file_path
+            for file_path in [Path(fp) for fp in self.data_file_paths]
```
davidsmejia, line 1169: It might make sense to rename `data_file_paths` as `data_files` and create a property `data_file_paths` which handles the converting the array of strings to an array of lib paths.

davidsmejia (commented):
Left some initial comments for your consideration. Looking good so far.

api/scpca_portal/models/library.py
avrohomgottlieb: I love the idea of moving the format constants `ANN_DATA` and `SINGLE_CELL_EXPERIMENT` outside of the models and into `common.py`. What do you think about moving the Modality constants there also?

For context, here are the places where constants are defined. Doesn't this seem redundant while also introducing unnecessary dependencies?

`computed_file.py`
```python
class OutputFileModalities:
    SINGLE_CELL = "SINGLE_CELL"
    SPATIAL = "SPATIAL"

    CHOICES = (
        (SINGLE_CELL, "Single Cell"),
        (SPATIAL, "Spatial"),
    )

class OutputFileFormats:
    ANN_DATA = "ANN_DATA"
    SINGLE_CELL_EXPERIMENT = "SINGLE_CELL_EXPERIMENT"

    CHOICES = (
        (ANN_DATA, "AnnData"),
        (SINGLE_CELL_EXPERIMENT, "Single cell experiment"),
    )
```

`sample.py`
```python
class Modalities:
    BULK_RNA_SEQ = "BULK_RNA_SEQ"
    CITE_SEQ = "CITE_SEQ"
    MULTIPLEXED = "MULTIPLEXED"
    SINGLE_CELL = "SINGLE_CELL"
    SPATIAL = "SPATIAL"

    NAME_MAPPING = {
        BULK_RNA_SEQ: "Bulk RNA-seq",
        CITE_SEQ: "CITE-seq",
        MULTIPLEXED: "Multiplexed",
        SPATIAL: "Spatial Data",
    }
```

`library.py`
```python
class FileFormats:
    ANN_DATA = "ANN_DATA"
    SINGLE_CELL_EXPERIMENT = "SINGLE_CELL_EXPERIMENT"

    CHOICES = (
        (ANN_DATA, "AnnData"),
        (SINGLE_CELL_EXPERIMENT, "Single cell experiment"),
    )

class Modalities:
    SINGLE_CELL = "SINGLE_CELL"
    SPATIAL = "SPATIAL"

    CHOICES = (
        (SINGLE_CELL, "Single Cell"),
        (SPATIAL, "Spatial"),
    )

```

api/scpca_portal/models/project.py
avrohomgottlieb: What about adding this to the `Library` and `Project` models? This accomplishes the above and allows us to work exclusively with `Path` objects.

```python
def __getattribute__(self, name):
    if name == "data_file_paths":
        raw_paths = super().__getattribute__(name)
        return [Path(p) for p in raw_paths]
    return super().__getattribute__(name)
```

avrohomgottlieb: Let's write an issue abstracting the `data_file_path` attribute and its assorted logic into a base class (per our conversation). Also, let's look into a custom model field to handle what's being discussed here (converting all entries from a str to a Path object).

api/scpca_portal/models/computed_file.py
```diff
@@ -105,6 +108,92 @@ def __str__(self):
...
+                cls.OUTPUT_README_FILE_NAME,
+            )
+            # Metadata file
+            zip_file.write(getattr(cls.MetadataFilenames, metadata_path_var))
+
+            if not download_config.get("metadata_only", False):
+                for file_path in library_data_file_paths:
+                    zip_file.write(file_path)
+
+            for file_path in project_data_file_paths:
```
davidsmejia, line 172: Lets bump this under the if not metadata_only guard

davidsmejia (approved):

Issue #740, opened 2024-06-03
Update computed file creation logic in Project and Sample models


avrohomgottlieb:
### Context

The `Project` and `Sample` models both have computed file driver methods named `Project::create_project_computed_files` and `Sample::create_sample_computed_files` respectively. 

These methods are responsible for 
- Evaluating all possible computed file configurations and calling the corresponding `ComputedFile::get_project|sample_<modality>_file` methods (which create new `ComputedFile` objects and carry out the computed file archiving)
- Calling `ComputedFile::process_computed_file` on them (which uploads the passed computed file to s3).

### Problem or idea

Now that we've moved from explicitly checking configuration conditions (upon which to call a corresponding `ComputedFile::get_project|sample_<modality>_file` method) to iterating over a collection of acceptable prebaked DOWNLOAD_OPTIONS on each `Project` or `Sample`, the `Project::create_project_computed_files` and `Sample::create_sample_computed_files methods` must be updated and refactored accordingly.

### Solution or next step

This is a possible implementation, which would have to be tweaked depending on whether the context was for `Project` or `Sample` computed files:
```python
def create_computed_files(
        self,
        max_workers=8,  # 8 = 2 file formats * 4 mappings.
        clean_up_output_data=True,
        update_s3=False,
    ):
        """Prepares ready for saving project computed files based on generated file mappings."""

        def on_get_project_file(future):
            computed_file = future.result()
            if computed_file:
                computed_file.upload_to_s3(update_s3)
            if clean_up_output_data:
                computed_file.clean_up_output_data()
            connection.close()

        with ThreadPoolExecutor(max_workers=max_workers) as tasks:
            for download_options in config.PREBAKED_PROJECT_DOWNLOADS:
                tasks.submit(
                    ComputedFile.get_project_file,
                    self,
                    download_options,
                ).add_done_callback(on_get_project_file)

        self.update_downloadable_sample_count()
```

avrohomgottlieb:
Addressed in PR https://github.com/AlexsLemonade/scpca-portal/issues/785

PR #776, opened 2024-06-25
740 - Project and Sample Create Computed Files Driver Methods

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/740

- Added `Project::create_computed_files` and `Sample::create_computed_files` methods, which utilize iteration over prebaked download configuration to kick of a ThreadPoolExecutor which creates computed files in parallel.
- Added `Project|Sample::get_download_config_file_output_name` method for dynamic naming of zip files.
- Added `Sample::get_lock_name` utility method to abstract lock naming logic.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/models/sample.py
+++ b/api/scpca_portal/models/sample.py
@@ -192,28 +192,12 @@ class Sample(CommonDataAttributes, TimestampedModel):
         return sorted(multiplexed_sample_ids)
 
     @property
-    def output_multiplexed_computed_file_name(self):
-        return f"{'_'.join(self.multiplexed_ids)}_multiplexed.zip"
-
-    @property
     def output_multiplexed_metadata_file_path(self):
         return Sample.get_output_metadata_file_path(self.scpca_id, Sample.Modalities.MULTIPLEXED)
 
     @property
-    def output_single_cell_computed_file_name(self):
-        return f"{self.scpca_id}.zip"
-
-    @property
-    def output_single_cell_anndata_computed_file_name(self):
-        return f"{self.scpca_id}_anndata.zip"
-
-    @property
     def output_single_cell_metadata_file_path(self):
         return Sample.get_output_metadata_file_path(self.scpca_id, Sample.Modalities.SINGLE_CELL)
-
-    @property
-    def output_spatial_computed_file_name(self):
-        return f"{self.scpca_id}_spatial.zip"
 
     @property
     def output_spatial_metadata_file_path(self):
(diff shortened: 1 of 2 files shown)


api/scpca_portal/models/sample.py
```diff
@@ -265,6 +265,71 @@ def single_cell_file_formats(self):
...
+        # The keys are the sample.multiplexed_ids since that will be unique across shared zip files.
+        non_multiplexed_ids = set(s.scpca_id for s in samples)
+        multiplexed_ids = set(
+            ["_".join(s.multiplexed_ids) for s in samples if s.has_multiplexed_data]
+        )
+        locks = {
+            f'{id_entry}-{config["modality"]}-{config["format"]}': Lock()
+            for id_entry in non_multiplexed_ids.union(multiplexed_ids)
+            for config in common.GENERATED_SAMPLE_DOWNLOAD_CONFIGURATIONS
+        }
```
davidsmejia, line 312: Maybe, instead of all this add a property?

```python
@property
def lock_prefix(self):
    if self.is_multiplexed:
       return "_".join(self.multiplexed_ids)
    return self.scpca_id
```

Then:

```suggestion
        lock_prefixes = set([s.lock_prefix for s in samples])
        locks = {
            f'{lock_prefix}-{config["modality"]}-{config["format"]}': Lock()
            for lock_prefix in lock_prefixes
            for config in common.GENERATED_SAMPLE_DOWNLOAD_CONFIGURATIONS
        }
```

api/scpca_portal/models/sample.py
```diff
@@ -265,6 +265,71 @@ def single_cell_file_formats(self):
...
+            case {"modality": "SPATIAL"}:
+                return self.output_spatial_computed_file_name
+            case {"format": "ANN_DATA"}:
+                return self.output_single_cell_anndata_computed_file_name
+            case {"modality": "SINGLE_CELL"}:
+                return (
+                    self.output_single_cell_computed_file_name
+                    if not self.has_multiplexed_data
+                    else self.output_multiplexed_computed_file_name
+                )
```
avrohomgottlieb, line 279:
```python
    def get_download_config_file_ouput_name(self, download_config: Dict):
        const parts = [self.scpca_id]

        if config.get("includes_merged", False):
            parts.append("merged")

        if self.multiplexed:
            parts.append("multiplexed")

        return f"{'_'.join(parts)}.zip"
```

api/scpca_portal/models/sample.py
```diff
@@ -147,6 +147,11 @@ def get_metadata(self) -> Dict:
 
         return sample_metadata
 
+    def get_lock_name(self, download_config: Dict) -> str:
+        """Returns a unique lock name based on multiplexed status and download config"""
+        lock_prefix = self.scpca_id if not self.is_multiplexed else "_".join(self.multiplexed_ids)
+        return f'{lock_prefix}-{download_config["modality"]}-{download_config["format"]}'
+
```
davidsmejia, line 154:
```suggestion
    def get_config_identifier(self, download_config: Dict) -> str:
        """Returns a unique identifier for the sample and download config combination. Multiplexed samples share an ID and are not unique."""
        sample_id = self.scpca_id if not self.is_multiplexed else "_".join(self.multiplexed_ids)
        return "_".join([sample_id] + sorted(download_config.values()))
```

What do you think about this change? I like the idea of moving away from "lock" in the function name as it doesn't really matter, we just want to indicate that the processing is not unique which is why they share a name.

davidsmejia (approved):
Left a couple small suggestions but this looks good to me.

api/scpca_portal/models/sample.py
avrohomgottlieb: Yeah, this is great. I'll make this addition.

Issue #753, opened 2024-06-10
Samples table: Change 'Other Modalities' to 'Modalities'


dvenprasad:
### Context
During evals, users were confused by the `Other Modalities` field. They consistently assumed that the sample did not have single-cell data at all, but only dat the for the modalities mentioned in that column. 

### Problem or idea
Change the column to `Modalities` and display all the modalities available for the sample, including single-cell or single-nucleus.

### Solution or next step

Tagging @davidsmejia for next steps

davidsmejia:
Requested changes:
- change sample column name from `Other Modalities` to `Modalities`
- manually prefix (put in front) `Single-cell` to all lists ONLY on the clientside sample table column that will now be called `Modality` - this probably means we can clean up any N/A handling as well.

PR #783, opened 2024-07-01
753 - Samples table: Change 'Other Modalities' to 'Modalities' 

nozomione:
## Issue Number
Closing #753

The following changes are included in this PR:
- Changed the name of the sample colum from "Other Modalities" to "Modalities"
- Prefixed the above column values with "Single-cell" and removed the default "NA" value

@dvenprasad Please see the latest UI [here](https://scpca-portal-fay0p0ep5-ccdl.vercel.app/).

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots
N/A

### DIFF
--- a/client/src/components/ProjectSamplesTable.js
+++ b/client/src/components/ProjectSamplesTable.js
@@ -175,7 +175,7 @@ export const ProjectSamplesTable = ({
       accessor: ({ modalities }) => {
         const prefix = 'Single-cell'
         return modalities.length > 0
-          ? `${prefix} ${modalities.join(` , ${prefix}`)}`
+          ? `${prefix} ${modalities.join(` , ${prefix} `)}`
           : prefix
       }
     },
--- a/client/src/components/ProjectSamplesTable.js
+++ b/client/src/components/ProjectSamplesTable.js
@@ -171,8 +171,13 @@ export const ProjectSamplesTable = ({
     { Header: 'Sequencing Units', accessor: 'seq_units' },
     { Header: 'Technology', accessor: 'technologies' },
     {
-      Header: 'Other Modalities',
-      accessor: ({ modalities }) => modalities.join(', ') || 'N/A'
+      Header: 'Modalities',
+      accessor: ({ modalities }) => {
+        const prefix = 'Single-cell'
+        return modalities.length > 0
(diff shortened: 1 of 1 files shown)


dvenprasad (changes requested):
`Single-cell` and `Bulk RNA_seq` are two different modalities. There should be a comma between them.

Screenshot 2024-07-01 at 14 58 20

dvenprasad (commented):
Oh also for the multiplexed project, what is the rationale behind saying both `Single-cell` and `Single-cell Multiplexed`?

nozomione:
> `Single-cell` and `Bulk RNA_seq` are two different modalities. There should be a comma between them.

Hey Deepa, I've removed the extra prefix, pls take a look [here]().

About multiplexed, are we going to remove the "Multiplexed" here?
Screenshot 2024-07-01 at 3 32 58 PM

dvenprasad (approved):
This looks good. No need to remove `Multiplexed`

🐨 🐨 🐨

client/src/components/ProjectSamplesTable.js
```diff
@@ -171,8 +171,13 @@ export const ProjectSamplesTable = ({
...
     { Header: 'Technology', accessor: 'technologies' },
     {
-      Header: 'Other Modalities',
-      accessor: ({ modalities }) => modalities.join(', ') || 'N/A'
+      Header: 'Modalities',
+      accessor: ({ modalities }) => {
+        const prefix = 'Single-cell'
+        return `${prefix} ${modalities.length > 0 ? ',' : ''} ${modalities.join(
+          `, `
+        )}`
```
davidsmejia, line 179:
```suggestion
        return ['Single-cell', ...modalities].join(', ')}`
```

I think we can simplify this a bit since that value will always be an array. Also, we want to be sure that the comma has a trailing space.

davidsmejia: Suggestion is malformed, but the idea is there, I think this could be a one line method

nozomione: Thank you for this great suggestion - I've applied it at 3fa9756577c28f217faec957bc01f2a648e74e2f!

davidsmejia (approved):
LGTM

Issue #741, opened 2024-06-03
Clean up old implementation of metadata parsing and file computation inside of Project::load_data


avrohomgottlieb:
### Context

At the conclusion of all of the additive changes listed below,
- https://github.com/AlexsLemonade/scpca-portal/issues/710
- https://github.com/AlexsLemonade/scpca-portal/issues/711
- https://github.com/AlexsLemonade/scpca-portal/issues/731
- https://github.com/AlexsLemonade/scpca-portal/issues/732
- https://github.com/AlexsLemonade/scpca-portal/issues/739
- https://github.com/AlexsLemonade/scpca-portal/issues/735
- https://github.com/AlexsLemonade/scpca-portal/issues/740

it is now possible to fully integrate these updates into the codebase and to clean up the old implementation of metadata parsing and file computation.

### Solution or next step

Ultimately, this refers to
- The unpacking of `Project::handle_samples_metadata` and pulling out `Project::load_samples_metadata`, `Project::load_libraries_metadata`, and adding a function which then updates `Sample` and `Project` aggregate values.
- The changing of the references inside of `Project::load_data` 
  - From `Project::create_project_computed_file` to `Project::create_computed_file`,
  - From `Sample::create_sample_computed_file` to `Sample::create_computed_file`
- Going through the `Project` and `Sample` models and deleting any helper functions which are no longer in use after the rewiring.

avrohomgottlieb:
According to the following comment: https://github.com/AlexsLemonade/scpca-portal/pull/773#discussion_r1651707595
- Let's handle moving `Modality` and `Format` constants from the models to common.py
(@davidsmejia as per our conversation 2024-07-01)

avrohomgottlieb:
Addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/788

PR #788, opened 2024-07-10
741 - Finalize new implementation of Project::load_data, clean up old implementation

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/741

- Wired up new implementation of `Project::load_data` (which uses the `Library` model)
- Created new Sample aggregate methods: 
  - `Project::update_sample_derived_properties`
  - `Project::update_sample_modality_properties`
  - `Project::update_sample_aggregate_properties`
- Cleaned up and removed old implementation of `Project::load_data (which used in-memory file mappings)
  - Removed a bunch of dead code, including     
    - `Project::handle_samples_metadata`
    - `Project::combine_metadata`
    - `Project::write_combined_metadata_libraries`
    - `Project::create_project_computed_files`
    - `Sample::create_sample_computed_files`
    - `Project::get_demux_sample_ids`
    - `Project::get_multiplexed_aggregate_fields`
    - `Project::get_multiplexed_with_mapping`
    - `Project::get_multiplexed_with_combined_metadata`
    - `Project::get_multiplexed_library_path_mapping`
    - `Project::get_non_downloadable_sample_ids`
    - `Project::get_library_metadata_keys`
    - `Project::get_combined_metadata_by_sample`
    - `Project::get_sample_libraries_mapping`
    - `Project::get_sample_metadata_keys`
    - `Project::get_multiplexed_samples_metadata`
    - `Project::get_sample_input_data_dir`
    - `Project::add_project_metadata`

## Types of changes

## Functional tests

No new tests added, but all tests in `test_load_data` updated to handle new way of determining output file names dynamically via download configurations.

## Checklist

## Screenshots

N/A


api/scpca_portal/models/computed_file.py
```diff
@@ -141,42 +157,73 @@ def get_project_file(cls, project, download_config: Dict, computed_file_name: st
...
-                    zip_file.write(file_path)
-
-            for file_path in project_data_file_paths:
-                zip_file.write(file_path)
+                    output_path = file_path.relative_to(f"{project.scpca_id}/")
+                    # Swap delimiter in multiplexed sample libraries from comma to underscore
+                    if "," in file_path.parent.name:
+                        new_delimiter_dir = Path("_".join(file_path.parent.name.split(",")))
+                        output_path = new_delimiter_dir / file_path.name
+                    zip_file.write(Library.get_local_file_path(file_path), output_path)
```
davidsmejia, line 196:
```suggestion
                    zip_file.write(
                        Library.get_local_file_path(file_path),
                        Library.get_zip_file_path(file_path),
                    )
```

Then make this method abstract and just pop off the "root" of the filepath and then rename any comma separated folders or files in the returned path.

```python

@classmethod
def get_zip_file_path(cls, file_path: Path) -> Path:
    # we don't want the file nested in the project directory
    output_path = # remove the parent path
    # we want to only output _ delimited files and folders
    # rename any folder that is comma separated  
    return new_delimiter_dir / file_path.name
```

api/scpca_portal/models/computed_file.py
```diff
@@ -141,42 +157,73 @@ def get_project_file(cls, project, download_config: Dict, computed_file_name: st
...
+
+                for file_path in project_data_file_paths:
+                    if "bulk" in file_path.name and download_config["modality"] == "SPATIAL":
+                        continue
+                    output_path = file_path.relative_to(f"{project.scpca_id}/")
+                    zip_file.write(Library.get_local_file_path(file_path), output_path)
```
davidsmejia, line 202: use `get_zip_file_path` here as well

api/scpca_portal/test/management/commands/test_load_data.py
```diff
@@ -966,7 +1045,7 @@ def test_spatial_metadata(self):
         # │       ├── tissue_lowres_image.png
         # │       └── tissue_positions_list.csv
         # └── spatial_metadata.tsv
-        self.assertEqual(len(project_zip.namelist()), 19)
+        self.assertEqual(len(project_zip.namelist()), 16)
```
davidsmejia, line 1048: We should look at why we are no longer seeing the directories in namelist, that seems odd to me.

We did see a difference of now nesting the library folder in the sample folder, so we should update the commented out file tree to reflect that change.

davidsmejia (commented):
left some initial comments

api/scpca_portal/models/project.py
```diff
@@ -63,6 +62,21 @@ class Meta:
...
+            if hasattr(project, key):
+                if key.startswith("includes_") or key.startswith("has_"):
+                    setattr(project, key, utils.boolean_from_string(data.get(key, False)))
+                else:
+                    setattr(project, key, data.get(key))
+
+        project.data_file_paths = project.get_data_file_paths()
+
+        return project
+
```
avrohomgottlieb, line 79: This was not added in this PR, but was moved to the top of the file (similar to the placement of similar `get_from_dict` methods on other models)

api/scpca_portal/models/library.py
```diff
@@ -175,21 +189,51 @@ def get_metadata(self) -> Dict:
...
+        Return all of a library's file paths that are suitable for the passed download config.
+        """
+        omit_suffixes = set()
+        # If metadata_only download, then leave omit suffixes as empty set, else populate
+        if not download_config.get("metadata_only", False):
+            omit_suffixes = set(common.FORMAT_EXTENSIONS.values())
+            # If merged download, keep omit suffixes set full as no library data files are included
+            # If not, remove download config format from omit suffixes set
+            if not download_config.get("includes_merged", False):
+                omit_suffixes.remove(common.FORMAT_EXTENSIONS.get(download_config["format"]))
```
davidsmejia, line 222:
```suggestion

        if download_config.get("metadata_only", False):
            return []

        omit_suffixes = set(common.FORMAT_EXTENSIONS.values())

        if not download_config.get("includes_merged", False):
            requested_format = common.FORMAT_EXTENSIONS.get(download_config["format"])
            omit_suffixes.remove(requested_format)
            
```

It would be nice to get rid of the negated check but we still want to remove everything for metadata only download, meaning no files at all.

davidsmejia (commented):
Left some initial thoughts

api/scpca_portal/models/library.py
```diff
@@ -175,21 +189,58 @@ def get_metadata(self) -> Dict:
...
+
+        # Project output paths are relative to project directory
+        if "metadata_only" in download_config:
+            output_path = file_path.relative_to(path_parts[0])
+        # Sample output paths are relative to project and sample directories
+        else:
+            output_path = file_path.relative_to(path_parts[0] / path_parts[1])
+
+        # Comma separated lists of multiplexed samples should become underscore separated
+        return Path(str(output_path).replace(",", "_"))
```
avrohomgottlieb, line 246: Wanted to ask your opinion on this @davidsmejia.

At some point you mentioned transitioning towards one `create_computed_files` method from the two `Project::get_computed_files` and `Sample::get_computed_files`, and one `get_file` method from the two `ComputedFile::get_project_file` and `Computed_File::get_sample_file`.

Obviously line 239 of checking for "metadata_only" is non-obvious that this field only belongs to `Project` download configs. In line with the above point, would it make sense to add to each `download_config` a `type` field, which would have a value of either `"Project"` or `"Sample"`? If so, then that could be used as the conditional identifier in line 239 here as well.

davidsmejia: I think the more correct way to check this is if `download_config` is in `common.GENERATED_PROJECT_DOWNLOAD_CONFIGURATIONS`. That would help clear up the ambiguity.

The other thing is this download config will change when we move to ad hoc but it is too early to tell what that should look like now. Most likely we will be passing in the project/sample/queryset and getting the libraries from that. But again not a problem for today.

api/scpca_portal/models/computed_file.py
```diff
@@ -141,42 +157,70 @@ def get_project_file(cls, project, download_config: Dict, computed_file_name: st
...
+                        )
+                        zip_file.write(
+                            Library.get_local_file_path(file_path),
+                            file_path.relative_to(f"{project.scpca_id}/"),
+                        )
 
         computed_file = cls(
-            has_bulk_rna_seq=project.has_bulk_rna_seq,
+            has_bulk_rna_seq=(
+                download_config["modality"] == "SINGLE_CELL" and project.has_bulk_rna_seq
```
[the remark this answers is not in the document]
avrohomgottlieb: To clarify - what you're saying is that we should add a constant to `common.py` for `modality`? We had originally mentioned just using the `modality` constant in `Library`, but we can't import `Library` into `common.py` due to circular importing.

We could save the reworking on other constants and removing them from the models to a separate PR.

api/scpca_portal/models/library.py
avrohomgottlieb: Yes! A clean solution which is relevant today. Very nice. We'll solve the bigger problems soon enough!

api/scpca_portal/models/computed_file.py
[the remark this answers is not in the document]
davidsmejia: For now I think Library.Modalities.SINGLE_CELL would be correct here. But yes in the future we should collect these constants in common.py or similar and then including them in to each of the models.

api/scpca_portal/test/management/commands/test_load_data.py
```diff
@@ -192,8 +201,8 @@ def test_merged_project_anndata_cite_seq(self):
             # └── single_cell_metadata.tsv
             files = set(project_zip.namelist())
             self.assertEqual(len(files), 8)
-            self.assertIn("SCPCP999992_merged.rds", files)
-            self.assertNotIn("SCPCP999992_merged_adt.h5ad", files)
+            self.assertIn("merged/SCPCP999992_merged.rds", files)
```
davidsmejia, line 204: Update the contents comment just above this line

davidsmejia (commented):
Left some comments for cleaning up

api/scpca_portal/models/sample.py
```diff
@@ -149,7 +137,9 @@ def get_metadata(self) -> Dict:
 
     def get_lock_name(self, download_config: Dict) -> str:
         """Returns a unique lock name based on multiplexed status and download config"""
-        lock_prefix = self.scpca_id if not self.is_multiplexed else "_".join(self.multiplexed_ids)
+        lock_prefix = (
+            self.scpca_id if not self.has_multiplexed_data else "_".join(self.multiplexed_ids)
```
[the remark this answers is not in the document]
avrohomgottlieb: I noticed this also, but I was concerned that it might be confusing. It feels non-obvious that self.multiplexed_ids would return the normal self.scpca_id if its a non-multiplexed sample. It's obvious when you look up at the implementation, but wasn't sure if it would lead to confusion. Thoughts?

api/scpca_portal/models/computed_file.py
```diff
@@ -141,42 +157,70 @@ def get_project_file(cls, project, download_config: Dict, computed_file_name: st
...
+                        file_path = Path(
+                            "/".join(
+                                [
+                                    project.scpca_id,
+                                    library.samples.first().scpca_id,
+                                    f"{library.scpca_id}_spatial",
+                                    f"{library.scpca_id}_metadata.json",
+                                ]
+                            )
+                        )
```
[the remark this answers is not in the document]
davidsmejia: Ok, so our documentation says we do this so we should keep this for now.

api/scpca_portal/test/management/commands/test_load_data.py
avrohomgottlieb: #### A few notes on this
1. There was a bug in the old implementation, where spatial libraries were not nested under their sample folders in the project zip. The bug can be seen below:
  - In `ComputedFile::get_project_single_cell_file` (correct)
```python
    with ZipFile(...) as zip_file:
        ...
        for sample_id, file_paths in sample_to_file_mapping.items():
            for file_path in file_paths:
                # Correct nesting of library data files under sample directory.
                zip_file.write(file_path, Path(sample_id, file_path.name))
```
  - In `ComputedFile::get_project_spatial_file` (bug)
```python
    with ZipFile(...) as zip_file:
        ...
        for sample_id, file_paths in sample_to_file_mapping.items():
            sample_path = project.get_sample_input_data_dir(sample_id)
            for file_path in file_paths:
                # Using relative_to here removes the intended nesting by excluding sample directory.
                zip_file.write(file_path, Path(file_path).relative_to(sample_path))
```
2. The only place that subdirectories seem to be included as their own entries in `ZipFile.namelist()` is during testing of the spatial computed file. In the new implementation with the correction of the above bug, the subdirectories are now no longer included as their own entries in `ZipFile.namelist()`.

avrohomgottlieb: 3. While the tests include in them a commented out file tree (with the number of files supposed to be included and a mention of `(including subdirectory names)`, all of the numbers which are actually asserted differ from the number in the comment. For example:
- `test_single_cell_metadata`
  - `There are 14 files (including subdirectory names)`
   *This comment is incorrect because there are 16 files including subdirectories*
  - `self.assertEqual(len(project_zip.namelist()), 14)`
- `test_multiplexed_metadata`
  - `There are 14 files (including subdirectory names)`
  - `self.assertEqual(len(project_zip.namelist()), 12)`
   *Notice how we test for 12 files, which excludes 2 subdirectories*
- `test_merged_project_anndata_cite_seq`
  - `There are 6 files (including subdirectory names)`
  - `self.assertEqual(len(project_zip.namelist()), 8)`
   *Notice how we test for 8 files, which excludes 3 subdirectories*
- `test_merged_project_anndata_no_cite_seq`
  - `There are 8 files (including subdirectory names)`
  - `self.assertEqual(len(project_zip.namelist()), 10)`
   *Notice how we test for 10 files, which excludes 3 subdirectories*

api/scpca_portal/test/management/commands/test_load_data.py
avrohomgottlieb: Good catch. Upon referring to documentation, it looks like merged data files should be nested under the zip root directory, not under a nested `merged/` directory inside the zip root. I've made the change accordingly to transform merges paths to reflect this.

api/scpca_portal/models/computed_file.py
```diff
@@ -194,223 +239,6 @@ def get_project_file(cls, project, download_config: Dict, computed_file_name: st
...
-        with ZipFile(computed_file.zip_file_path, "w") as zip_file:
-            for src, dst in project_file_mapping.items():
-                zip_file.write(src, dst)
-
-            for sample_id, file_paths in sample_to_file_mapping.items():
-                for file_path in file_paths:
-                    if str(file_path).split("_")[-1] not in sample_file_suffixes:
-                        continue
-                    # Nest these under their sample id.
-                    zip_file.write(file_path, Path("individual_reports", sample_id, file_path.name))
```
avrohomgottlieb, line 265: @davidsmejia Also, I wanted to ask about what this nesting of non-merged library reports inside a `individual_reports` directory during merged file computation was all about. I believe I should update my PR to also accommodate this nested directory.

api/scpca_portal/models/library.py
```diff
@@ -172,21 +189,63 @@ def get_metadata(self) -> Dict:
...
+        if download_config in common.GENERATED_PROJECT_DOWNLOAD_CONFIGURATIONS:
+            output_path = file_path.relative_to(path_parts[0])
+        # Sample output paths are relative to project and sample directories
+        else:
+            output_path = file_path.relative_to(path_parts[0] / path_parts[1])
+
+        # Transform merged project data files to no longer be nested in a merged directory
+        if file_path.parent.name == "merged":
+            output_path = file_path.relative_to(path_parts[0] / path_parts[1])
+
```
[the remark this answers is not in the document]
davidsmejia: I think this is fine for now. Really, we will probably move this behavior by creating a new base class for Project and Library that handles the attributes etc for mapping files.

api/scpca_portal/test/management/commands/test_load_data.py
davidsmejia: Ok, as long as the comments have all been updated we should be good.

api/scpca_portal/models/computed_file.py
davidsmejia: Yeah good catch, it should match whatever we say we are doing in terms of file structure.

api/scpca_portal/models/sample.py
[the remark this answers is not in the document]
davidsmejia: I think this is ok to still use it for now, we should probably rework the name to make it more obvious that it is the scpca id as a list that will contain mutliplexed ids. This is for internal use only so we have some flexibility here to work with.

davidsmejia (approved):
LGTM

Issue #778, opened 2024-06-27
Parse Sample Cell Estimates per sample from library metadata


avrohomgottlieb:
### Context

When parsing multiplexed library Json files, we receive a field populated with a Json Object called `sample_cell_estimates`. The keys of this object represent the `sample_ids` of the samples involved in the multiplexed library, and the values represent the estimated cell count contribution of each demuxed sample to the sequencing carried out (from which the library reading is created). We currently take this Json Object and send it through to the output multiplexed metadata tsv file, keeping it in object (or in Python parlance - dictionary) form for the output tsv file. As opposed to parsing the object and outputting individual values to the necessary rows (i.e. the row that corresponds to a certain sample within the library in question), we pass through the entire object for all rows represented in the multiplexed library.

### Problem or idea

Sending through a Json Object (Python Dictionary) to a tsv file is not good practice, and is not helpful to our end user. As well, a clearer and more transparent fix exists by just parsing the Json Object and writing that sample's `sample_cell_estimates` value to its row in the `sample_cell_estimates` column.

### Solution or next step

```python
# library.py
def combine_library_metadata(self, sample) -> Dict:
    metadata = self.project.get_metadata() | sample.get_metadata() | self.get_metadata()
    # If library.metadata is multiplexed and has the key "sample_cell_estimates", then do two things
    if self.is_multiplexed:
        # 1 - retrieve the value corresponding to the sample being iterated over, and return it as the value for a new "sample_cell_estimate" key
        metadata["sample_cell_estimate"] = metadata["sample_cell_estimates"][sample.scpca_id]
        # 2 - pop off the key "sample_cell_estimates" as it is no longer needed
        del metadata["sample_cell_estimates"]

    return metadata

def get_combined_library_metadata(self) -> List[Dict]:
    return [self.combine_library_metadata(sample) for sample in self.samples]
```

avrohomgottlieb:
Addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/789

PR #789, opened 2024-07-12
778 - Presenting sample sample_cell_estimates per sample in output metadata file

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/778

- Rename `sample_cell_estimates` attribute to `sample_cell_estimate`.
- Exclude `sample_cell_estimates` key from being passed through from `library.metadata` in `Library::get_metadata`
- Include `sample_cell_estiamate` key from Sample model, which pulls from the `Sample.demux_sample_cell_count` (which is updated in Project::update_sample_aggregate_values`
- Update `restval` in `metadata_file::write_metadata_dicts`

Note about addition of `restval`:
- The reason that the updating of the `restval` property of `metadata_file::write_metadata_dicts` is in this PR is because we have to assume that projects with multiplexed libraries will also have samples with non multiplexed libraries. As such, both `sample_cell_count_estimate` and `sample_cell_estimate` fields will be present, and depending on if the library is multiplexed or not, one of these two fields will be blank. Due to the fact that 'NA' `restvals` are present in `samples_metadata.csv`, which we end up passing through to our output metadata files, justifies making this addition here.

## Types of changes

## Functional tests

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/metadata_file.py
+++ b/api/scpca_portal/metadata_file.py
@@ -96,7 +96,6 @@ def write_metadata_dicts(list_of_dicts: List[Dict], output_file_path: str, **kwa
         "fieldnames", utils.get_sorted_field_names(utils.get_keys_from_dicts(list_of_dicts))
     )
     kwargs["delimiter"] = kwargs.get("delimiter", common.TAB)
-    kwargs["restval"] = kwargs.get("restval", "NA")
 
     sorted_list_of_dicts = sorted(
         list_of_dicts,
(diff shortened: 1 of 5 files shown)


api/scpca_portal/metadata_file.py
```diff
@@ -96,6 +96,7 @@ def write_metadata_dicts(list_of_dicts: List[Dict], output_file_path: str, **kwa
         "fieldnames", utils.get_sorted_field_names(utils.get_keys_from_dicts(list_of_dicts))
     )
     kwargs["delimiter"] = kwargs.get("delimiter", common.TAB)
+    kwargs["restval"] = kwargs.get("restval", "NA")
```
davidsmejia, line 99: This was already in dev, you might need to pull the latest into the feature branch and remove this

davidsmejia (approved):
LGTM but `restval` is already in dev so that will probably be a conflict

api/scpca_portal/metadata_file.py
avrohomgottlieb: Handled.

Issue #742, opened 2024-06-03
Break up ComputedFile::process_computed_file method


avrohomgottlieb:
### Context

`ComputedFile::process_computed_file` is responsible for saving the passed computed file, uploading it to s3, removing it from the local filesystem, and then closing the DB connection that was opened for the current thread.

### Solution or next step

This method should be broken up into two methods:
- One called `upload_to_s3` which uploads the inputted computed file to s3
- Another called `clean_up_local_computed_file` (or something similar) which removes/unlinks the local computed file.

As well, the closing of the DB connection should be passed up to the caller function and should be closer to the `ThreadPoolExecutor` for better clarity.

avrohomgottlieb:
Addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/798

PR #798, opened 2024-07-17
742 - Break up ComputedFile::process_computed_file

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/742

This PR deals with the breaking up of `ComputedFile::process_computed_file` into two methods:
- `ComputedFile::upload_s3_file`
- `ComputedFile::clean_up_local_computed_file`. 

The calling of these methods is moved outside of a wrapper (like the role that `ComputedFile::process_computed_file` played), and are called directly inside of `Project::create_computed_files::on_get_project_file` and `Sample::create_computed_files::on_get_sample_file` after computed files have been generated and `computed_file` objects have been created and returned.

As a note - a check has been included before the uploading and cleaning up of local files that makes sure that files are only uploaded and cleaned up once. This is especially a point of concern with computed files that have multiplexed samples. The check ensures that only one sample in a collection of multiplexed samples will be responsible for uploading and cleaning up files, via the new `Sample::is_last_multiplexed_sample` property method, removing duplicate and unwanted behavior.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/models/project.py
+++ b/api/scpca_portal/models/project.py
@@ -329,6 +329,8 @@ class Project(CommonDataAttributes, TimestampedModel):
 
         def on_get_project_file(future):
             if computed_file := future.result():
+                computed_file.save()
+
                 if update_s3:
                     computed_file.upload_s3_file()
                 if clean_up_output_data:
--- a/api/scpca_portal/models/project.py
+++ b/api/scpca_portal/models/project.py
@@ -329,7 +329,11 @@ class Project(CommonDataAttributes, TimestampedModel):
 
         def on_get_project_file(future):
             if computed_file := future.result():
-                computed_file.process_computed_file(clean_up_output_data, update_s3)
+                if update_s3:
+                    computed_file.upload_s3_file()
+                if clean_up_output_data:
+                    computed_file.clean_up_local_computed_file()
+
             # Close DB connection for each thread.
             connection.close()
 
--- a/api/scpca_portal/models/sample.py
+++ b/api/scpca_portal/models/sample.py
@@ -257,9 +257,34 @@ class Sample(CommonDataAttributes, TimestampedModel):
     ):
         """Prepares ready for saving project computed files based on generated file mappings."""
 
+        # Prepare a threading.Lock for each sample, with the chief purpose being to protect
+        # multiplexed samples that shares a zip file.
+        locks = {}
+
         def on_get_sample_file(future):
             if computed_file := future.result():
-                computed_file.process_computed_file(clean_up_output_data, update_s3)
+                # Once created, save immediately, so that it can be queried below
+                computed_file.save()
+
+                config = {
+                    "modality": computed_file.modality,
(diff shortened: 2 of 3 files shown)


api/scpca_portal/models/project.py
```diff
@@ -329,7 +329,13 @@ def create_computed_files(
 
         def on_get_project_file(future):
             if computed_file := future.result():
-                computed_file.process_computed_file(clean_up_output_data, update_s3)
+                computed_file.save()
+
+                if update_s3:
+                    computed_file.upload_s3_file()
+                if clean_up_output_data:
+                    computed_file.clean_up_local_computed_file()
```
avrohomgottlieb, line 337: Because `ComputedFile::upload_s3_file` initiates the s3 uploading by spawning off a subprocess (using the subprocess module), it should be possible from a process scheduling perspective for the `ComputedFile::clean_up_local_computed_file` to be called directly afterwards and to corrupt the s3 upload. How should we handle this?

api/scpca_portal/models/sample.py
```diff
@@ -257,9 +257,34 @@ def create_computed_files(
...
+                with locks.get(sample.get_config_identifier(config)):
+                    nonlocal update_s3
+                    if update_s3:
+                        computed_file.upload_s3
+                        update_s3 = False
+
+                all_samples_created = (
+                    len(computed_file.sample.multiplexed_ids)
+                    == ComputedFile.objects.filter(s3_key=computed_file.s3_key).count()
+                )
```
avrohomgottlieb, line 283: As a general aside, this is the type of intermediate querying that could justify a Public=False or Loading=True attribute on the ComputedFile model. Not saying that that's the route that we need to take, but I guess something like this is what I had in mind.

api/scpca_portal/models/project.py
avrohomgottlieb: In theory yes, but the `subprocess` module in the python standard library is blocking so its not a problem.

api/scpca_portal/models/sample.py
davidsmejia:
```python
        def on_get_sample_file(future):
            if computed_file := future.result():
                # Once created, save immediately, so that it can be queried below

                # Only upload and clean up the last multiplexed sample
                if computed_file.sample.is_last_sample:
                    if update_s3:
                        computed_file.upload_s3_file()
                    if clean_up_output_data:
                        computed_file.clean_up_local_computed_file()

                computed_file.save()
            # Close DB connection for each thread.
            connection.close()

```

davidsmejia: No need for a lock, we can just check if it is the last sample and then upload it / clean up. Otherwise, just save and bypass those steps.

avrohomgottlieb: - Add `Sample::is_last_sample` `@property` method
- Also move `locks` back down below

Issue #766, opened 2024-06-17
Just in Time Readme Creation


avrohomgottlieb:
### Context

Currently, Readme creation is done eagerly at the beginning of `Project::load_data` where all Readme permutations are generated and stored locally until the time for file computation comes around. When file computation begins and we compute files for each type of modality and file format, the each corresponding Readme is retrieved from its location in the local file system and added to the correct archive.

The current Readme permutations include:
- Single Cell SCE
- Single Cell SCE Merged
- Multiplexed Single Cell SCE
- Single Cell Anndata
- Single Cell Anndata Merged
- Spatial
- Metadata only

### Problem or idea

Our goal is to begin relying less on the local file system and eager downloading by transitioning towards a model of just-in-time / lazy downloading. This is especially important as we transition the portal towards implementation of AWS Batch.

### Solution or next step

The first step would be to move all Readme file creation logic outside of the `Project` model and all of the Readme constants (such as the readme name and its template) from the `ComputedFile` model to a `readme_creation.py` module inside of `scpca_portal/`. As we are doing just-in-time creation, there will be no need for all `README_<type>_FILE_NAME` and `README_<type>_FILE_PATH` constants, which represent the eagerly downloaded Readmes before they are included in the desired zipfile. For the remaining constant, `TEMPLATE_PATHS`, it might make sense to put all paths into a `TEMPLATE_PATHS` dictionary which can be accessed more easily/dynamically than through a list of constants.

The dictionary could look as follows:
```python
README_TEMPLATE_PATH = common.TEMPLATE_PATH / "readme"
TEMPLATE_PATHS = {
    "SINGLE_CELL_SINGLE_CELL_EXPERIMENT": README_TEMPLATE_PATH / "single_cell.md",
    "SINGLE_CELL_SINGLE_CELL_EXPERIMENT_MERGED": README_TEMPLATE_PATH / "single_cell_merged.md",
    "SINGLE_CELL_SINGLE_CELL_EXPERIMENT_MULTIPLEXED" : README_TEMPLATE_PATH / "metadata_only.md",
    "SINGLE_CELL_ANN_DATA": README_TEMPLATE_PATH / "anndata.md",
    "SINGLE_CELL_ANN_DATA_MERGED": README_TEMPLATE_PATH / "anndata_merged.md",
    "SPATIAL_SINGLE_CELL_EXPERIMENT": README_TEMPLATE_PATH / "spatial.md",
[...]

PR #802, opened 2024-07-18
766 - Just in time Readme Creation

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/766

Changes include:
- Added a new module `readme_creation` which houses all readme creation logic and constants.
- Added a new method `readme_creation::create_readme_file` which dynamically creates a readme according to a `download_config`.
- Replaced readme archive writing:
  - From inclusion of pregenerated readme files with `zipfile.write()`,
  - To just in time writing of the readme string returned from `readme_creation::create_readme_file` to the zip archive with `zipfile.writestr()` .
- Removed eager readme logic generation from the `Project` model and unused readme constants from the `ComputedFile` model.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


davidsmejia (commented):
After seeing this again, I think we can make it feel more like the `metadata_file` module since they have similar functionality.

davidsmejia (approved):
One comment about kwargs but otherwise looks good

Issue #784, opened 2024-07-01
Remove intermediate rendering of metadata files


avrohomgottlieb:
### Context

In `ComputedFile::get_project_file` and `ComputedFile::get_sample_file` we query for a collection of libraries according to the passed `download_config`, and then write those libraries to an output tsv metadata file. The output metadata file is then included immediately thereafter in the new zip archive created for the passed `download_config`, and then discarded.

### Problem or idea

Because the output metadata file is only written so that it might be included soon after in the upcoming file archive, it can be optimized by writing it to an in-memory buffer, as opposed to writing it to disk and thereafter retrieving it from disk.

We can use the `StringIO` class from the `io` module to create such a buffer that would suite our needs. Even better, `csv.DictWriter`, which we're currently using for file writing in `metadata_file.write_metadata_dicts`, supports direct writing to the `io.StringIO` file buffer, and won't necessitate a refactor.

### Solution or next step

```python
    def get_project_file(cls, project, download_config: Dict, computed_file_name: str) -> Self:
        libraries = Library.get_project_libraries_from_download_config(project, download_config)
        ...
        with ZipFile(zip_file_path, "w") as zip_file:
            ...
            # Include metadata file in zip
            metadata_path_var = (
                f'{download_config["modality"]}_METADATA_FILE_NAME'
                if not download_config["metadata_only"]
                else "METADATA_ONLY_FILE_NAME"
            )
            with io.StringIO() as metadata_buffer:
                metadata_file.write_metadata_dicts(libraries_metadata, metadata_buffer)
                zip_file.writestr(getattr(cls.MetadataFilenames, metadata_path_var), metadata_buffer.get_value())
            ...
```

avrohomgottlieb:
Addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/803

PR #803, opened 2024-07-19
784 - In memory metadata file generation

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/784

- Replace eager writing of `output_metadata_file` to a local file (before zip archival) with writing it to an in memory buffer (during zip archival)
  - The file is written to the buffer and thereafter immediately written to the zip archive
- Move metadata output file naming logic out of zip archival process in `ComputedFile::get_project|sample_file` and into own method inside of `metadata_file` module called `metadata_file::get_metadata_file_name`
- Move MetadataFilenames constants from `ComputedFile` model to `metadata_file` module
- Update `metadata_file::write_metadata_dicts` to handle writing to a buffer

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/metadata_file.py
+++ b/api/scpca_portal/metadata_file.py
@@ -132,31 +132,3 @@ def format_metadata_dict(metadata_dict: Dict) -> Dict:
     Returns a copy of metadata dict that is formatted and ready to be written to file.
     """
     return {k: utils.string_from_list(v) for k, v in metadata_dict.items()}
-
-
-def write_metadata_dicts(list_of_dicts: List[Dict], output_file_path: str, **kwargs) -> None:
-    """
-    Writes a list of dictionaries to a csv-like file.
-    Optional modifiers to the csv.DictWriter can be passed to function as kwargs.
-    """
-    kwargs["fieldnames"] = kwargs.get(
-        "fieldnames", utils.get_sorted_field_names(utils.get_keys_from_dicts(list_of_dicts))
-    )
-    kwargs["delimiter"] = kwargs.get("delimiter", common.TAB)
-    # By default fill missing values with "NA"
-    kwargs["restval"] = kwargs.get("restval", common.NA)
-
-    sorted_list_of_dicts = sorted(
-        list_of_dicts,
-        key=lambda k: (
-            k[common.PROJECT_ID_KEY],
-            k[common.SAMPLE_ID_KEY],
-            k[common.LIBRARY_ID_KEY],
-        ),
-    )
-
-    with open(output_file_path, "w", newline="") as raw_file:
-        csv_writer = csv.DictWriter(raw_file, **kwargs)
-        csv_writer.writeheader()
-        for metadata_dict in sorted_list_of_dicts:
-            csv_writer.writerow(format_metadata_dict(metadata_dict))
--- a/api/scpca_portal/metadata_file.py
+++ b/api/scpca_portal/metadata_file.py
@@ -94,7 +94,7 @@ class MetadataFilenames:
     METADATA_ONLY_FILE_NAME = "metadata.tsv"
 
 
-def get_metadata_file_name(download_config: Dict) -> str:
+def get_file_name(download_config: Dict) -> str:
     """Return metadata file name according to passed download_config."""
     output_file_constant = (
         f'{download_config["modality"]}_METADATA_FILE_NAME'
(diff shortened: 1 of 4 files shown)


api/scpca_portal/metadata_file.py
```diff
@@ -87,7 +88,26 @@ def transform_keys(data_dict: Dict, key_transforms: List[Tuple]):
     return data_dict
 
 
-def write_metadata_dicts(list_of_dicts: List[Dict], output_file_path: str, **kwargs) -> None:
+class MetadataFilenames:
+    SINGLE_CELL_METADATA_FILE_NAME = "single_cell_metadata.tsv"
+    SPATIAL_METADATA_FILE_NAME = "spatial_metadata.tsv"
+    METADATA_ONLY_FILE_NAME = "metadata.tsv"
```
avrohomgottlieb, line 94: I wasn't sure if these constants should be higher up in the file (as we do conventionally with constants), or if they should be close to the place where they're called. I decided to put them here for enhanced clarity during the PR review (so that it would be clear what `get_metadata_file_name` is doing).

api/scpca_portal/metadata_file.py
```diff
@@ -87,7 +88,26 @@ def transform_keys(data_dict: Dict, key_transforms: List[Tuple]):
...
+        f'{download_config["modality"]}_METADATA_FILE_NAME'
+        if not download_config.get("metadata_only", False)
+        else "METADATA_ONLY_FILE_NAME"
+    )
+
+    return getattr(MetadataFilenames, output_file_constant)
+
+
+def write_metadata_dicts(
+    list_of_dicts: List[Dict], output_file_path: str = None, *, buffer: io.StringIO = None, **kwargs
```
avrohomgottlieb, line 109: Do you think the signature is too busy here?

api/scpca_portal/models/computed_file.py
```diff
@@ -126,14 +116,15 @@ def get_project_file(cls, project, download_config: Dict, computed_file_name: st
...
-            )
+            with io.StringIO() as metadata_buffer:
+                libraries_metadata = [
+                    lib for library in libraries for lib in library.get_combined_library_metadata()
+                ]
+                # Write libraries metadata to buffer
+                metadata_file.write_metadata_dicts(libraries_metadata, buffer=metadata_buffer)
+                output_metadata_file_name = metadata_file.get_metadata_file_name(download_config)
+                # Write buffer directly to zip archive
+                zip_file.writestr(output_metadata_file_name, metadata_buffer.getvalue())
```
davidsmejia, line 127:
```suggestion
            libraries_metadata = [
                lib for library in libraries for lib in library.get_combined_library_metadata()
            ]

            metadata_file_name = metadata_file.get_file_name(download_config)
            metadata_file_contents = metadata_file.get_file_contents(libraries_metadata)

            # Write buffer directly to zip archive
            zip_file.writestr(output_metadata_file_name, metadata_file_contents)
```

I think this is the general flow we should be going for. The `metadata_file.get_file_contents` should handle the buffter under the hood.

api/scpca_portal/metadata_file.py
davidsmejia: Yeah we really only need a `metadata_file.get_file_contents` that takes the list of metadata dicts and kwargs for dict writer then returns what is currently `metadata_buffer.getvalue()`. Manager the buffer internally.

If we end up needing a `metadata_file.write_metadata_dict` then we will just call `get_file_contents` internally and write it to `output_file_path`. However, I don't think we need this at this time.

api/scpca_portal/metadata_file.py
davidsmejia: This is fine. You don't really care what the value is, just that it is consistent across the application.

api/scpca_portal/metadata_file.py
```diff
@@ -87,7 +88,26 @@ def transform_keys(data_dict: Dict, key_transforms: List[Tuple]):
...
+
+def get_metadata_file_name(download_config: Dict) -> str:
+    """Return metadata file name according to passed download_config."""
+    output_file_constant = (
+        f'{download_config["modality"]}_METADATA_FILE_NAME'
+        if not download_config.get("metadata_only", False)
+        else "METADATA_ONLY_FILE_NAME"
+    )
+
+    return getattr(MetadataFilenames, output_file_constant)
```
davidsmejia, line 105:
```suggestion

    if download_config.get("metadata_only", False):
        return MetadataFileNames.METADATA_ONLY_FILE_NAME

    return getattr(
        MetadataFilenames,
        f'{download_config["modality"]}_METADATA_FILE_NAME'
    )
```

I think at the minimum we should do the affirmative check, however I think early exit looks best.

davidsmejia (commented):
Looks mostly correct to me, mostly just think we should not be worried about the buffer outside of the `metadata_file` module.

api/scpca_portal/models/computed_file.py
avrohomgottlieb: I think this makes sense. Our goal should be creating a clean interface for the caller while abstracting away implementation details, and this better accomplishes that.

api/scpca_portal/metadata_file.py
avrohomgottlieb: Yes, this is better

avrohomgottlieb:
@davidsmejia I'd like to get your opinion about testing. 

I believe that the next step would be to delete `metadata_file::write_metadata_dicts` and restructure the test cases included in `test_metadata_file.py` for `metadata_file::write_metadata_dicts` to fit `metadata_file::get_file_contents`, but I'd like your input first on how to go about doing that.

Another outstanding question is whether or not `metadata_file::get_file_contents` should generate the file-as-a-string from a collection of libraries (generating the metadata dicts internally), or if the metadata dicts should be retrieved from the caller and passed into `metadata_file::get_file_contents`. It seems like it would be cleaner from the caller's perspective just to pass in the libraries and allow for the callee to take responsibility to grabbing the necessary metadata and generating the file-as-a-string, but there are also advantages to doing it the other way.

All of this affects of necessary testing for the function.

api/scpca_portal/metadata_file.py
```diff
@@ -87,6 +88,49 @@ def transform_keys(data_dict: Dict, key_transforms: List[Tuple]):
...
+    return getattr(MetadataFilenames, f'{download_config["modality"]}_METADATA_FILE_NAME')
+
+
+def get_file_contents(libraries, **kwargs) -> str:
+    """Return newly genereated metadata file as a string for immediate writing to a zip archive."""
+    libraries_metadata = [
+        format_metadata_dict(library_metadata)
+        for library in libraries
+        for library_metadata in library.get_combined_library_metadata()
+    ]
```
davidsmejia, line 111: Making this change would require an overhaul of the tests so let's keep this as the callers responsibility for now. You should open an issue that covers this change and the required changes to testing there.

davidsmejia (commented):
Looking good to me, one change needs to be made in `get_file_contents`.

> I believe that the next step would be to delete metadata_file::write_metadata_dicts and restructure the test cases included in test_metadata_file.py for metadata_file::write_metadata_dicts to fit metadata_file::get_file_contents, but I'd like your input first on how to go about doing that.

Correct, you just need to call `get_file_contents` instead and then you can convert it back to a list of dicts by using a combination of `DictReader` and `StringIO`.

> Another outstanding question is whether or not metadata_file::get_file_contents should generate the file-as-a-string from a collection of libraries (generating the metadata dicts internally), or if the metadata dicts should be retrieved from the caller and passed into metadata_file::get_file_contents. It seems like it would be cleaner from the caller's perspective just to pass in the libraries and allow for the callee to take responsibility to grabbing the necessary metadata and generating the file-as-a-string, but there are also advantages to doing it the other way.

This makes sense but we should postpone that change since it will take longer than the results are worth at this time.

davidsmejia (approved):
LGTM

Issue #797, opened 2024-07-17
[WIP] Management command for portal wide metadata download


nozomione:
> [!note]
> **This note will be deleted:**
> 
> Currently implementation details match this commit [`ComputedFile::get_project_file`](https://github.com/AlexsLemonade/scpca-portal/blob/9542045d1df8dc86cc9b23053898ce9b47bb3fbb/api/scpca_portal/models/computed_file.py#L130) in [`feature/remove-file-mappings`](https://github.com/AlexsLemonade/scpca-portal/tree/feature/remove-file-mappings) but will be adjusted later based on upcoming updates (e.g., using buffers instead of output data files to write and generate downloadable files).
> 
> e.g.) By using a buffer, the content of the portal-metadata readme file can be generated as follows:
> ```python
> from io import StringIO
> from django.template.loader import render_to_string
> 
> def get_portal_metadata_readme():
>   """Generate a portal-wide metadata only README content for zipping"""
>   readme_buffer = StringIO()
>   readme_buffer.write(
>      render_to_string(
>         ComputedFile.README_TEMPLATE_METADATA_PATH,
>         context={
>            "date": utils.get_today_string(),
>            "projects": Project.objects.filter(additional_restrictions__isnull=False).all()
>         }
>      ).strip()
>    )
>   # Make sure to reset the stream position to 0
>   readme_buffer.seek(0)
>   # Return the content of readme file
>   return readme_buffer.getvalue()
>   ```
> 
> The `ComputedFile::get_portal_metadata_file` below temporarily writes the content of readme file using the above example method to illustrate the use of a buffer.

### Context
Parent issue: [Management command to generate portal wide metadata only download](https://github.com/AlexsLemonade/scpca-portal/issues/708)

We'll be adding the portal-wide metadata only download feature to the portal. This allows users to download and review all the metadata information available on the portal without having to download the project's computed files.

The downloadable portal-wide metadata zip file will contain the following files:
- `README.md`
- `metadata.tsv` 

### Problem or idea
To accomplish this, we should do the following:

- Use the `Library` model to query all libraries' metadata to create the computed file instance and the zip file for the portal-wide metadata. 
- Create a new management command that saves the instance to the database and uploads the zip file to S3 bucket.

### Solution or next step
The expected steps are as follows:

1. In the `ComputedFile` model, add the following constants and a model filed.

    (**NOTE:** The readme's output data path is omitted here to demostrate the use of a buffer which is subject to change)

[...]

PR #804, opened 2024-07-19
797 - Scaffolding the porta-wide metadata management command and test files

nozomione:
## Issue Number
#797 

This is the 1st item in the stacked PR list for the linked issue.

The following changes are included:

- Added the `create-portal-metadata.py` file
- Added the `create-portal-metadata.py` file  which setup the database by calling `load_data`
- Added the `create-portal-metadata` command to `sportal`

## Types of changes

## Functional tests
N/A

## Checklist

## Screenshots
N/A

### DIFF
--- a/api/scpca_portal/test/management/commands/test_create_portal_metadata.py
+++ b/api/scpca_portal/test/management/commands/test_create_portal_metadata.py
@@ -4,7 +4,7 @@ from django.test import TransactionTestCase
 
 from scpca_portal import common
 from scpca_portal.management.commands import create_portal_metadata, load_data
-from scpca_portal.models import Library
+from scpca_portal.models import Library, Project, Sample
 
 # NOTE: Test data bucket is defined in `scpca_porta/common.py`.
 # When common.INPUT_BUCKET_NAME is changed, please delete the contents of
@@ -24,6 +24,11 @@ class TestCreatePortalMetadata(TransactionTestCase):
         shutil.rmtree(common.OUTPUT_DATA_PATH, ignore_errors=True)
 
     def load_test_data(self):
+        # Set up the database
+        PROJECT_COUNT = 3
+        SAMPLES_COUNT = 9
+        LIBRARIES_COUNT = 7
+
         self.loader.load_data(
             allowed_submitters=list(ALLOWED_SUBMITTERS),
             clean_up_input_data=False,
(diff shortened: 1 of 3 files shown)


api/scpca_portal/test/management/commands/test_create_portal_metadata.py
```diff
@@ -0,0 +1,50 @@
...
+                reload_all=False,
+                reload_existing=False,
+                scpca_project_id=project_id,
+                update_s3=False,
+            )
+
+    def test_create_portal_metadata(self):
+        # Set up the database
+        project_ids = ["SCPCP999990", "SCPCP999991", "SCPCP999992"]
+        self.setup_database(project_ids)
```
davidsmejia, line 42: This is the right idea, but I think we can omit the `project_ids` and not pass that kwarg. The `load_data` command will load everything all at once and since this is test data we don't have to worry about size limitations.

Also, just for clarity I think `load_test_data` instead of `setup_database` more clearly indicates what that method is limited to.

api/scpca_portal/test/management/commands/test_create_portal_metadata.py
```diff
@@ -0,0 +1,50 @@
...
+
+        # Make sure that the number of projects created matches the length of project_ids
+        self.assertEqual(Project.objects.count(), len(project_ids))
```
davidsmejia, line 45: I think this is also a good idea. The change I would suggest is instead of `project_ids` I would assert `PROJECT_COUNT`, `SAMPLES_COUNT`, and `LIBRARIES_COUNT` because as you probably surmised if those assertions fail upstream, downstream validity cannot be ensured.

davidsmejia (commented):
Just a couple things to take another look at but overall this is the right direction.

api/scpca_portal/test/management/commands/test_create_portal_metadata.py
nozomione: See the commit at f4df1159ca5ee2b20cc3e45b165b7f31e79cf74b.

api/scpca_portal/test/management/commands/test_create_portal_metadata.py
nozomione: For this, I've declared the constants `PROJECT_COUNT`, `SAMPLES_COUNT`, and `LIBRARIES_COUNT` and give them hard-coded values representing the corresponding total count. Then I am using the `Library` model to assert the equality of each count. Let me know your thought on this. Thank you David!

See the commit at f4df1159ca5ee2b20cc3e45b165b7f31e79cf74b.

api/scpca_portal/test/management/commands/test_create_portal_metadata.py
```diff
@@ -0,0 +1,61 @@
...
+            lib for library in libraries for lib in library.get_combined_library_metadata()
+        ]
+
+        self.assertEqual(
+            len(set(lib["scpca_project_id"] for lib in libraries_metadata)), PROJECT_COUNT
+        )
+        self.assertEqual(
+            len(set(lib["scpca_sample_id"] for lib in libraries_metadata)), SAMPLES_COUNT
+        )
+        self.assertEqual(libraries.count(), LIBRARIES_COUNT)
```
davidsmejia, line 56: I think what you want to do here is instead of counting instances of IDs in `libraries_metadata` you should just count rows in the database. Much like how you do libraries.count(). Only you can just do:

```python
self.assertEqual(Library.objects.count(), LIBRARIES_COUNT)
```

Do the same for samples and projects.

api/scpca_portal/test/management/commands/test_create_portal_metadata.py
```diff
@@ -0,0 +1,48 @@
...
+    def load_test_data(self):
+        # Expected object counts
+        PROJECTS_COUNT = 3
+        SAMPLES_COUNT = 9
+        LIBRARIES_COUNT = 7
+
+        self.loader.load_data(
+            allowed_submitters=list(ALLOWED_SUBMITTERS),
+            clean_up_input_data=False,
+            clean_up_output_data=False,
```
davidsmejia, line 35: Just a note: for this test you could have `cleanup_output_data=True` since you aren't testing any output of this command. This is purely just a shortcut to populate the data. This would give you a cleaner output `test_data/output` directory.

davidsmejia (approved):

api/scpca_portal/test/management/commands/test_create_portal_metadata.py
nozomione: Thank you for the note, David!

Issue #767, opened 2024-06-18
Just in Time Data File Downloads


avrohomgottlieb:
### Context

The downloading of data files is done eagerly inside of `load_data::load_data` by a subroutine called `load_data::download_data`. `load_data::download_data` utilizes the `subprocess` module in the Python standard library, which we use to spawn a `bash` process which calls the `aws` cli and configures it to download all files from a given input bucket. When we archive files later on after manipulating input metadata files, we grab the archive's desired data files from the local file system and include them into the active archive.

### Problem or idea

Instead of downloading terabytes of data in one fail swoop, we want to move to lazy just-in-time downloads during archival where files can be included when they're needed into their desired archive and not kept around in the local file system indefinitely. This will also better allow us to transition to AWS Batch, we can divide up the work of archiving Project and Sample computed files between multiple (virtual) servers, and as such need to enable the ability for downloading to happen in a more decoupled and distributed fashion.

### Solution or next step

Now that we have `Project::data_file_paths` and `Library::data_file_paths` attributes, the downloading will be trivial and quick.

- Utility function which downloads data based on a list of file paths from a passed input bucket
- Utility function which checks if a file exists locally before downloading it
- Inside the method which handles file archiving, collect all data file paths (`Project` and `Library`) and calling the download_data utility function with a list of these paths.

An alternative solution, with the data file downloading and file archiving all occurring in memory via IO buffers, could look as follows:
```python
import boto3
import io
import zipfile

# Initialize the S3 client
s3 = boto3.client('s3')

# Specify the S3 bucket and files for download
input_bucket_name = 'input-bucket-name'
output_bucket_name = 'output-bucket-name'
file_keys = ['file1.txt', 'file2.txt', 'file3.txt']  # List of files in S3

# Create an in-memory bytes buffer
zip_buffer = io.BytesIO()

# Create a new zip file in the bytes buffer
with zipfile.ZipFile(zip_buffer, 'w', zipfile.ZIP_DEFLATED) as zip_file:
    for file_key in file_keys:
        # Download the file from S3 into a bytes buffer
        file_obj = io.BytesIO()
        s3.download_fileobj(input_bucket_name, file_key, file_obj)
        file_obj.seek(0)  # Go to the start of the BytesIO object

        # Add the file to the zip archive
        zip_file.writestr(file_key, file_obj.read())

# Seek to the start of the buffer to read the zip file content
zip_buffer.seek(0)

# Save the zip file to the local file system (if needed)
[...]

PR #809, opened 2024-07-24
767 - Just In Time Data File Downloads

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/767

This PR introduces the following changes:
- Move all logic that handles `s3` to a new `s3` module
- Add generic `download_s3_files` function to `s3` module which downloads all files from an S3 bucket according to a list of passed filters
- Add following functions to `s3` module, all of which leverage `download_s3_files`:
  -  `download_metadata_files` (called inside of `load_data` command)
  - `download_data_files` (called inside of `ComputedFile::get_project_file` and `ComputedFile::get_sample_file`)

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/management/commands/configure_aws_cli.py
+++ b/api/scpca_portal/management/commands/configure_aws_cli.py
@@ -1,6 +1,11 @@
+import logging
 import subprocess
 
 from django.core.management.base import BaseCommand
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+logger.addHandler(logging.StreamHandler())
 
 
 class Command(BaseCommand):
@@ -39,5 +44,8 @@ class Command(BaseCommand):
                 f"aws configure set default.s3.max_bandwidth {kwargs['s3_max_bandwidth']}MB/s",
             )
 
+        logger.info("Configuring AWS CLI...")
         for command in commands:
             subprocess.check_call(command.split())
+
+        logger.info("AWS CLI successfully configured!")
--- a/api/scpca_portal/management/commands/configure_aws_cli.py
+++ b/api/scpca_portal/management/commands/configure_aws_cli.py
@@ -30,21 +30,27 @@ class Command(BaseCommand):
     def handle(self, *args, **kwargs):
         self.configure_aws_cli(**kwargs)
 
-    def configure_aws_cli(self, **kwargs):
+    def configure_aws_cli(
+        self,
+        *,
+        s3_max_concurrent_requests: int = 10,
+        s3_multipart_chunk_size: int = 8,
+        s3_max_bandwidth: int,
+        **kwargs,
+    ):
         commands = [
             # https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#payload-signing-enabled
             "aws configure set default.s3.payload_signing_enabled false",
             # https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#max-concurrent-requests
-            "aws configure set default.s3.max_concurrent_requests "
-            f"{kwargs['s3_max_concurrent_requests']}",
+            f"aws configure set default.s3.max_concurrent_requests {s3_max_concurrent_requests}",
             # https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#multipart-chunksize
-            "aws configure set default.s3.multipart_chunksize "
-            f"{kwargs['s3_multipart_chunk_size']}MB",
+            f"aws configure set default.s3.multipart_chunksize {s3_multipart_chunk_size}MB",
         ]
-        if kwargs["s3_max_bandwidth"]:
+
+        if s3_max_bandwidth:
             commands.append(
                 # https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#max-bandwidth
-                f"aws configure set default.s3.max_bandwidth {kwargs['s3_max_bandwidth']}MB/s",
+                f"aws configure set default.s3.max_bandwidth {s3_max_bandwidth}MB/s",
             )
 
         logger.info("Configuring AWS CLI...")
--- a/api/scpca_portal/management/commands/configure_aws_cli.py
+++ b/api/scpca_portal/management/commands/configure_aws_cli.py
@@ -1,0 +1,43 @@
+import subprocess
+
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = """Configures the AWS CLI.
+
+    Current supported options include:
+        --s3-max-bandwidth
+        --s3-max-concurrent-requests
+        --s3-multipart-chunk-size
+
+    The configuration works by writing the config values to ~/.aws/config,
+    which is references by the AWS CLI tool each time its called."""
+
+    def add_arguments(self, parser):
+        parser.add_argument("--s3-max-bandwidth", type=int, default=None, help="In MB/s")
+        parser.add_argument("--s3-max-concurrent-requests", type=int, default=10)
+        parser.add_argument("--s3-multipart-chunk-size", type=int, default=8, help="In MB")
+
+    def handle(self, *args, **kwargs):
+        self.configure_aws_cli(**kwargs)
+
+    def configure_aws_cli(self, **kwargs):
+        commands = [
+            # https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#payload-signing-enabled
+            "aws configure set default.s3.payload_signing_enabled false",
+            # https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#max-concurrent-requests
+            "aws configure set default.s3.max_concurrent_requests "
+            f"{kwargs['s3_max_concurrent_requests']}",
+            # https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#multipart-chunksize
+            "aws configure set default.s3.multipart_chunksize "
+            f"{kwargs['s3_multipart_chunk_size']}MB",
+        ]
+        if kwargs["s3_max_bandwidth"]:
+            commands.append(
+                # https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#max-bandwidth
+                f"aws configure set default.s3.max_bandwidth {kwargs['s3_max_bandwidth']}MB/s",
+            )
+
+        for command in commands:
+            subprocess.check_call(command.split())
--- a/api/scpca_portal/management/commands/configure_aws_cli.py
+++ b/api/scpca_portal/management/commands/configure_aws_cli.py
@@ -1,11 +1,14 @@
-import logging
 import subprocess
 
 from django.core.management.base import BaseCommand
 
-logger = logging.getLogger()
-logger.setLevel(logging.INFO)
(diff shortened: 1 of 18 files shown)


avrohomgottlieb:
As a note - this PR does not represent the permanent change that we're looking to implement with just in time file downloads. Our original goal was to be able to pass an array of `s3` paths (accessed at the library level with each library's `data_file_paths` attribute) to the `aws s3 cli` and receive those files in return, but it looks like this functionality is not currently supported by the `aws s3 cli`. 

The `aws s3 cli` implements two commands: `sync` and `cp`, neither of which exactly fit our use case.
- `sync` does a diff of the source and destination (which can be either s3 buckets or local directories) and downloads all files missing in the destination from the source. The way to specify files is by passing a collection of `include` and `exclude` statements as options to the command.
- `cp` copies one file from a source to a destination and thereafter closes the connection.

As such, this PR represents a sort of intermediate step where we have split up and moved downloading of data and metadata files more local to their ultimate consumption, with the intention of implementing a more correct solution to our use case later on.

avrohomgottlieb:
> As a note - this PR does not represent the permanent change that we're looking to implement with just in time file downloads. Our original goal was to be able to pass an array of `s3` paths (accessed at the library level with each library's `data_file_paths` attribute) to the `aws s3 cli` and receive those files in return, but it looks like this functionality is not currently supported by the `aws s3 cli`.

It looks like there was a bug in my application code, in function `s3::download_s3_files`, which calls `aws s3 sync` under the hood. The bug caused that when, I was attempting to include a list of individual files in `s3::download_data_files`, it kept throwing an error. As I worked on `s3::download_s3_files` while working through a different solution, the roadblock was lifted, and it became clear that our original solution was feasible. At this point, we're where we wanted to be with this ticket.

api/scpca_portal/management/commands/configure_aws_cli.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument("--s3-max-bandwidth", type=int, default=None, help="In MB/s")
+        parser.add_argument("--s3-max-concurrent-requests", type=int, default=10)
+        parser.add_argument("--s3-multipart-chunk-size", type=int, default=8, help="In MB")
+
+    def handle(self, *args, **kwargs):
+        self.configure_aws_cli(**kwargs)
+
+    def configure_aws_cli(self, **kwargs):
```
davidsmejia, line 30: lets define the keyword arguments here instead of **kwargs

api/scpca_portal/test/management/commands/test_load_data.py
```diff
@@ -7,7 +7,8 @@
 from django.test import TransactionTestCase
 
 from scpca_portal import common, metadata_file
-from scpca_portal.management.commands.load_data import Command
+from scpca_portal.management.commands.configure_aws_cli import Command as configure_aws_cli
```
avrohomgottlieb, line 10: We moved the calling of the `configure_aws_cli` command outside of `run_tests.sh` and inside of `test_load_data` due to a permissions error accessing the `~/.aws/config` file (first noted on this GHA https://github.com/AlexsLemonade/scpca-portal/actions/runs/10182489658/job/28165155525?pr=809)

We should open a ticket about this to investigate what happened here.

api/scpca_portal/models/computed_file.py
```diff
@@ -258,7 +257,13 @@ def get_sample_file(
...
     def download_url(self):
         """A temporary URL from which the file can be downloaded."""
-        return self.create_download_url()
+        if self.s3_bucket and self.s3_key:
+            # Append the download date to the filename on download.
+            date = utils.get_today_string()
+            s3_key = Path(self.s3_key)
+            filename = f"{s3_key.stem}_{date}{s3_key.suffix}"
+
+            return s3.generate_pre_signed_link(s3_key, filename)
```
avrohomgottlieb, line 266: I'd like to call into question this `if` check.

The old way that we created `ComputedFile` objects was that we would create the object without attributes before archiving and populate the attributes after archiving. In such a scenario, there was a moment that a `ComputedFile` object could be build before having its `s3_bucket` and `s3_key` populated, a relevant possibility that justifies the `if` check.

The current way that we do things is that we first archive the file and then, at completion, we create `ComputedFile` objects with all of their attributes at once. 

As such, this `if` statement does not currently protect from the no longer necessary early accession (i.e. after the object has been created but before the `s3_bucket` and `s3_key` attributes have been added), but rather for the case that these attributes are malformed or empty when their originally passed in. 

Two possible alternatives include:
- Deleting the `if` check all together
- Just checking for these attributes, and if they don't exist then throwing an error or logging it and `return`ing early

api/scpca_portal/s3.py
```diff
@@ -0,0 +1,148 @@
...
+    command_parts.append("--exclude=*")
+    command_parts.append("--include=*_metadata.*")
+
+    if "public-test" in bucket_name:
+        command_parts.append("--no-sign-request")
+
+    subprocess.check_call(command_parts)
+
+
+def delete_output_file(key: str) -> bool:
```
avrohomgottlieb, line 110: We should have a consistent API for all of the `upload`/`delete`/`download` `s3` module methods where they either return a `boolean` value or they don't. I understand the argument of handling (and logging) errors in the callee, and then allowing the caller to act based on the function's success or failure status, but we currently don't employ that type of error handling in our callers (at least where `s3::delete_output_file` is called in `Project::purge`).

How would you like to handle this?

api/scpca_portal/management/commands/configure_aws_cli.py
avrohomgottlieb: Unfortunately Python doesn't have an analogous feature to ES6's destructuring assignment, where you can do the following:
```python
def handle(self, *args, **kwargs):
    s3_max_bandwidth, s3_max_concurrent_requests, s3_multipart_chunk_size, ...rest = kwargs
    self.configure_aws_cli(s3_max_bandwidth, s3_max_concurrent_requests, s3_multipart_chunk_size)

def configure_aws_cli(
    self,
    *,
    s3_max_concurrent_requests: int = 10,
    s3_multipart_chunk_size: int = 8,
    s3_max_bandwidth: int,
) -> None:
    ...
```

The closest thing that I found was the following (from a stack overflow [link](https://stackoverflow.com/questions/2955412/destructuring-bind-dictionary-contents)):
```python
from operator import itemgetter

params = {'a': 1, 'b': 2}

a, b = itemgetter('a', 'b')(params)
```

For now I went with the following, but I'd be interesting to hear your feedback:
```python
def handle(self, *args, **kwargs):
    self.configure_aws_cli(**kwargs)

def configure_aws_cli(
    self,
    *,
    s3_max_concurrent_requests: int = 10,
    s3_multipart_chunk_size: int = 8,
    s3_max_bandwidth: int,
    **kwargs,
):
    ...
```

api/scpca_portal/management/commands/load_data.py
```diff
@@ -79,12 +74,28 @@ def add_arguments(self, parser):
...
+
+    @staticmethod
+    def clean_up_input_data(project):
+        shutil.rmtree(common.INPUT_DATA_PATH / project.scpca_id, ignore_errors=True)
+
+    @staticmethod
+    def clean_up_output_data():
+        for path in Path(common.OUTPUT_DATA_PATH).glob("*"):
+            path.unlink(missing_ok=True)
+
```
avrohomgottlieb, line 98: This commit `770d404` is actually supposed to be in the branch from PR https://github.com/AlexsLemonade/scpca-portal/pull/836. I'm not sure how it was added to this branch, I feel like it must have had something to do with the commit being shared during a rebase (as this branch was used as the base for the branch in PR https://github.com/AlexsLemonade/scpca-portal/pull/836 branch).

api/scpca_portal/test/management/commands/test_load_data.py
davidsmejia: Yeah let's open the ticket and remove it from being executed before the tests for now.

api/scpca_portal/management/commands/configure_aws_cli.py
davidsmejia: yeah what you went with is correct, but you dont need to pass **kwargs since we dont handle them

api/scpca_portal/management/commands/load_data.py
davidsmejia: I think we can avoid this by using merge by default from now on. Don't cherry pick changes into your branch unless needed to resolve something. Rebasing can be done until you push remote changes. But honestly, 99% of the time we just want to merge changes up and keep git history as is.

api/scpca_portal/s3.py
avrohomgottlieb: Anything that doesn't return a value should do this
- Use `try/catch` block inside of the s3 module functions

Don't need to adjust the caller for now

`list_s3_paths` - change the logging level from `error` to `warning`

api/scpca_portal/s3.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def download_input_files(file_paths: List[Path], *, bucket_name: str = common.INPUT_BUCKET_NAME):
+    """Download all passed data file paths which have not previously been downloaded.'"""
+    command_parts = ["aws", "s3", "sync", f"s3://{bucket_name}", common.INPUT_DATA_PATH]
+
+    download_queue = [fp for fp in file_paths if not fp.exists()]
+    # If download_queue is empty, exit early
+    if not download_queue:
+        return
```
avrohomgottlieb, line 86: Should this return `True` or `False`? I feel like it should return `True` because its returning early because there are no files to download, not because the downloading failed.

api/scpca_portal/models/computed_file.py
avrohomgottlieb: @davidsmejia

davidsmejia (approved):
This looks good to me, just need to address my one comment before merging.

api/scpca_portal/models/project.py
```diff
@@ -213,7 +213,7 @@ def on_get_project_file(future):
                 computed_file.save()
 
                 if update_s3:
-                    computed_file.upload_s3_file()
+                    s3.upload_output_file(computed_file)
```
davidsmejia, line 216:
```suggestion
                    s3.upload_output_file(computed_file.s3_key)
```

api/scpca_portal/models/sample.py
```diff
@@ -267,7 +267,7 @@ def on_get_sample_file(future):
                 # Only upload and clean up the last if multiplexed
                 if computed_file.sample.is_last_multiplexed_sample:
                     if update_s3:
-                        computed_file.upload_s3_file()
+                        s3.upload_output_file(computed_file)
```
davidsmejia, line 270:
```suggestion
                        s3.upload_output_file(computed_file.s3_key)
```

api/scpca_portal/models/project.py
avrohomgottlieb: Good catch

PR #821, opened 2024-07-30
hotfix database migration 47

davidsmejia:
## Issue Number

N/A

Looks like our staging DB is in a intermediate state. One of the project's samples is not populated correctly. Since the migrations weren't applied exploring the DB via the ORM is a bit tricky. I am going to apply this change to see if it is more permissive and allows the migration to go through. This is a minimal change. Otherwise, I will wipe staging db, migrate, and rerun the load data command to repopulate it.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


avrohomgottlieb (approved):
👍

PR #822, opened 2024-07-30
add django extensions to installed apps

davidsmejia:
## Issue Number

N/A

I was debugging the staging database when I needed some commands that are present in django_extensions. We already include this module in our base requirements so this is just a matter of telling django to include it on start.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


avrohomgottlieb (approved):
LGTM

Issue #796, opened 2024-07-15
remove `age_at_diagnosis` rename add `age_timing` to model


davidsmejia:
### Context

Currently we rename `age` to `age_at_diagnosis` when loading sample metadata. However, we will be receiving a new column that changes the meaning of this value and makes the renaming incorrect. The two new metadata keys for this information will be `age` and `age_timing`. `Age` should be the same and have the same meaning, but `age_timing` will indicate when that `age` was collected.

### Problem or idea

We need to update our models, serializers, and client to account for this change. Also, when we sort metadata keys to write them to file we use a common sort order that will need to be updated. Additionally, the there are many references to `age_at_diagnosis` in the tests that will need to be renamed.

These changes need to land with the corresponding client changes before deployment.

### Preflight
- [x] test data issue made
- [x] test data available
- [x] Where to place in sort order

### Solution or next step

- remove `age` / `age_at_diagnosis` rename from `metadata_file.SAMPLE_METADATA_KEYS` list to stop renaming
- add both / remove `age_at_diagnosis` from Sample model properties and makemigrations
- add both / remove `age_at_diagnosis` from all SampleSerializer
- add both / remove `age_at_diagnosis` from common.METADATA_COLUMN_SORT_ORDER
- update the tests to reflect `age` and `age_timing` and not `age_at_diagnosis`
- update `client/src/components/ProjectSamplesTable.js` to reflect this change

avrohomgottlieb:
@davidsmejia we currently represent `age_at_diagnosis` (or `age` in this context) as a TextField in Postgres. Should it not be an Integer Field?

avrohomgottlieb:
Addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/826

PR #826, opened 2024-07-31
796 - Add age_timing Sample attribute and rename age_at_diagnosis

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/796

- Rename Sample attribute `age_at_diagnosis` to `age`
- Add Sample attribute `age_timing`
- Make migrations to db to reflect changes
- Updates serializers, views, tests and affected functions in Sample model
- Update test bucket from `2024-06-28/` to `2024-07-19/`

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


api/scpca_portal/models/sample.py
```diff
@@ -35,8 +35,14 @@ class Modalities:
             SPATIAL: "Spatial Data",
         }
 
+    class AgeTimings(models.TextChoices):
+        COLLECTION = "collection"
+        DIAGNOSIS = "diagnosis"
+        UNKNOWN = "unknown"
```
avrohomgottlieb, line 41: This feels to me like a more concise design decision than taking the other `CHOICES` approach, which would look as follows:
```
class AgeTimings:
    COLLECTION = "collection"
    DIAGNOSIS = "diagnosis"
    UNKNOWN = "unknown"

    CHOICES = (
        (COLLECTION, "Collection"),
        (DIAGNOSIS, "Diagnosis"),
        (UNKNOWN, "Unknown"),
    )
```

api/scpca_portal/models/sample.py
```diff
@@ -35,8 +35,14 @@ class Modalities:
...
+
     additional_metadata = models.JSONField(default=dict)
-    age_at_diagnosis = models.TextField(blank=True, null=True)
+    age = models.TextField(blank=True, null=True)
```
avrohomgottlieb, line 44: Do we want `age` to be a `TextField` or an `IntegerField`?

api/scpca_portal/test/management/commands/test_load_data.py
```diff
@@ -815,7 +819,7 @@ def test_single_cell_metadata(self):
         self.assertIsNone(sample.demux_cell_count_estimate)
         self.assertFalse(sample.has_bulk_rna_seq)
         self.assertFalse(sample.has_cite_seq_data)
-        self.assertEqual(sample.sample_cell_count_estimate, 3417)
+        self.assertEqual(sample.sample_cell_count_estimate, 3432)
```
avrohomgottlieb, line 822: The change here in the assertion stems from a change to the `filtered_cells` attribute in the `SCPCL999990` json file.

api/scpca_portal/models/sample.py
davidsmejia, line 41:
```suggestion
        UNKNOWN = "unknown"
        NA = "NA"
```

We need to account for a potential `NA` value from the input bucket.

api/scpca_portal/test/management/commands/test_load_data.py
davidsmejia: I feel like at this point it wouldn't hurt to add a comment saying that this line will probably fail when switching test data versions.

davidsmejia (commented):
Looks pretty good, left a few suggested changes.

api/scpca_portal/models/sample.py
avrohomgottlieb: Are we sure that we want to support an `NA` value? 

In the updates to the docs from PR https://github.com/AlexsLemonade/scpca-docs/pull/331 it only mentions 3 possible values:
- `diagnosis`
- `collection`
- `unknown`

Effectively `unknown` in this case would be acting as a more descriptive `NA`

davidsmejia: Actually, I think it is safe to remove the choices from this attribute. We should always be populated this based off of the input metadata and the value does not affect anything internally. So we should just keep this as a nullable string.

We need this nullable for the migration and after that we can require it's population at creation time.

davidsmejia: Actually since we never expect to have a null value during loading we should just have both be a text field with the default as common.NA.

api/scpca_portal/migrations/0049_auto_20240731_1820.py
```diff
@@ -0,0 +1,30 @@
...
+            model_name="sample",
+            name="age_timing",
+            field=models.TextField(
+                choices=[
+                    ("collection", "Collection"),
+                    ("diagnosis", "Diagnosis"),
+                    ("unknown", "Unknown"),
+                ],
+                default="unknown",
+            ),
```
davidsmejia, line 28: looks like we need to remove this migration and run `makemigrations` again

avrohomgottlieb: Sorry about that. Will take care of this now.

client/src/components/ProjectSamplesTable.js
nozomione, line 1: I've also noticed that we're using the `age_at_diagnosis` field in `client/src/api/resources/samples` which needs to be updated:

https://github.com/AlexsLemonade/scpca-portal/blob/fc6e742246ed79347020f121e898256a1e2fe38f/client/src/api/resources/samples.js#L12

And the mock data prop name (total 7 places) in storybook `client/.storybook/data/project`:

https://github.com/AlexsLemonade/scpca-portal/blob/fc6e742246ed79347020f121e898256a1e2fe38f/client/.storybook/data/project.js#L76

Other than that, the change looks great 👍 !

nozomione (changes requested):

nozomione (approved):
LGTM 🚀

api/scpca_portal/models/sample.py
jaclyn-taroni: I don't think this matters anymore, but `age_timing` can be `NA` if `age` is `NA`.

PR #849, opened 2024-08-13
Production Deploy

davidsmejia:
# Production Deploy
### API
- refactors s3 logic into separate module
- adds on demand downloads of data files at time of zip creation
- changes `age_at_diagnosis` back to `age` and adds`age_timing`
### Client
- Add `Age timing` column to project samples table


nozomione (approved):
👍

Issue #864, opened 2024-08-26
Add tests for new loader module


avrohomgottlieb:
### Context

Issue https://github.com/AlexsLemonade/scpca-portal/issues/863 describes the creation of a `loader` module. This module will extract redundant logic from the `load_data`, `load_metadata` and `generate_computed_files` commands, and collect it all in one place.

### Problem or idea

A new test module should be created for the `loader` module, which unit tests the functionality of the different functions and of the flow of execution.

### Solution or next step

`test_loader.py`
```python
class TestLoader(TestCase):
    def setUp(self):
        # set up dummy project metadata dicts
        # set up project factory instances

    @patch(scpca_portal.s3.download_input_metadata)
    def test_get_projects_metadata(self, mock_download_input_metadata):
        ...
        mock_download_input_metadata.assert_called_once()
        # assertions about output projects_metadata dicts

    @patch(scpca_portal.s3.download_input_metadata)
    def test_get_projects_metadata_filter_id_passed(self, mock_download_input_metadata):
        ...
        mock_download_input_metadata.assert_called_once()
        # assertions about output project_metadata dict (list of one item)
        
    def test_create_project(self):
        # go from project_metadata to project being created
        self.assertTrue(Project.objects.exists(scpca_id=project_metadata["scpca_project_id"]))
            
    def test_create_project_data_files_dont_exist(self):
        # create some folders, but not others, and assert what the correct output should be

    def test_create_project_submitter_not_in_whitelist(self):
        # pass a custom submitter list and make sure that the project_metadata dict passed has a non matching submitter (pi_name)

    def test_bulk_create_project_relations(self):
        # test that the correct number of Contact, EA and Publication objects have been submitted

    @patch(scpca_portal.Project.load_metadata)
    def test_create_samples_and_libraries(self, mock_project_load_metadata):
        mock_project_load_metadata.assert_called_once()

    

```

avrohomgottlieb:
Addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/878

Issue #863, opened 2024-08-23
Add a loader module which extracts redundant logic from load_data, load_metadata and generate_computed_files command


avrohomgottlieb:
### Context

Recent PRs broke up the `load_data` command into `load_metadata` and `generate_computed_files` in order to accommodate for the `Batch` workflow. The goal is for `load_metadata` to run on the API, and for a `dispatch_generate_computed_files` command on the API to submit jobs to `Batch`, which will use `generate_computed_files` as their entry points to start off the create computed files flow.

### Problem or idea

The `load_data`, `load_metadata` and `generate_computed_files` management commands all share logic that we'd like to abstract away into a `loader` module. Making this functionality more modularly will minimize code duplication, ultimately making it easier for us to unit test. Another point of consideration about testing is that code executed on `Batch` instances will be harder for us to test, and we'll be restricted only to integration tests which test the entire `Batch` workflow. Abstracting away this logic to a `loader` module, however, will give us full coverage over this logic.

### Solution or next step

load_data.py
```python
def load_data(
    self,
    input_bucket_name: str,
    clean_up_input_data: bool,
    clean_up_output_data: bool,
    max_workers: int,
    reload_existing: bool,
    scpca_project_id: str,
    update_s3: bool,
    submitter_whitelist: Set[str],
    **kwargs,
) -> None:

    loader.prepare_data_dirs(clean_up_input_data)
    for project_metadata in loader.get_projects_metadata(scpca_portal_id):
        # validates that a project can be added to the db, and if possible, creates the project
        if project := loader.create_project(project_metadata, submitter_whitelist, input_bucket_name, reload_existing, update_s3)
            loader.bulk_create_project_relations(project, project_metadata)
            loader.create_samples_and_libraries(project)

    loader.prepare_data_dirs(clean_up_input_data)
    for project in loader.get_computed_file_projects(project_id: str = None):
        loader.generate_computed_files(
            project,
            max_workers,
            clean_up_output_data,
            update_s3,
        )

    loader.prepare_data_dirs(clean_up_input_data)
```

loader.py
```python
def prepare_data_dirs(clean_up_input_data: bool = None) -> None:
    # Prepare data input directory.
        common.INPUT_DATA_PATH.mkdir(exist_ok=True, parents=True)
        if clean_up_input_data:
            for path in Path(common.INPUT_DATA_PATH).glob("*"):
                path.unlink(missing_ok=True)       

        # Prepare data output directory.
[...]

PR #868, opened 2024-08-29
863 - Add Loader Module

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/863

Created a new `loader` module, which houses common logic from the `load_data`, `load_metadata` and `generate_computed_files` commands.

A variety of top levels functions were added in the `loader` module to be called by the commands to better consolidate and self-document logic within the flow of execution. As well, some methods previously used in some of these commands have been designated as private methods for internal use.

## Types of changes

## Functional tests

Will be handled in https://github.com/AlexsLemonade/scpca-portal/issues/864

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/loader.py
+++ b/api/scpca_portal/loader.py
@@ -1,0 +1,210 @@
+from concurrent.futures import ThreadPoolExecutor
+from functools import partial
+from pathlib import Path
+from threading import Lock
+from typing import Any, Dict, List, Set
+
+from django.conf import settings
+from django.db import connection
+from django.template.defaultfilters import pluralize
+
+from scpca_portal import common, metadata_file, s3
+from scpca_portal.config.logging import get_and_configure_logger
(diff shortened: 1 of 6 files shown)


api/scpca_portal/loader.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def get_projects_for_computed_file_generation(update_s3: bool = False) -> List[Project]:
+    """
+    Query the database, prep and return all projects that need computed files to be generated.
+    """
+    return [
+        _prep_project_for_computed_file_generation(project, update_s3)
+        for project in Project.objects.filter(project_computed_files__isnull=True)
+    ]
```
davidsmejia, line 175:
```suggestion
```

Let's remove these and instead query the project in the commands (handle missing project there). Then purging should be handled per script.

davidsmejia (approved):
LGTM

Issue #844, opened 2024-08-12
Develop AWS Batch Compute Environment


avrohomgottlieb:
### Context

This ticket is a stub for a ticket to be filled out later on detailing the steps for developing the compute environment for Batch. 

We want to utilize the following resources
- Fargate for container orchestration
- EBS (3.5x size of project) for a mounted volume
  - If EBS is not possible, than EFS
- Ram (match to the API)

### Solution or next step

infrastucture/batch/roles.tf
```terraform
resource "aws_iam_role" "batch_service_role" {
    name = "scpca-portal-batch-service-role-${var.user}-${var.stage}"
    assume_role_policy = <<EOF
{
    "Version": "2012-10-17",
    "Statement": [
    {
        "Action": "sts:AssumeRole",
        "Effect": "Allow",
        "Principal": {
        "Service": "batch.amazonaws.com",
        }
    }
    ]
}
EOF
    tags = var.default_tags
}

resource "aws_iam_role_policy_attachment" "batch_service_role" {
    role = aws_iam_role.batch_service_role.name
    policy_arn = "arn:aws:iam::aws:policy/service-role/AWSBatchServiceRole"
}

resource "aws_iam_role" "ecs_task_role" {
    name = "scpca-portal-ecs-task-role-${var.user}-${var.stage}"

    assume_role_policy = <<EOF
    {
        "Version": "2012-10-17",
        "Statement": [
        {
            "Action": "sts:AssumeRole",
            "Effect": "Allow",
            "Principal": {
                "Service": "ecs-tasks.amazonaws.com"
            }
        }
       ]
   }
   EOF

    tags = var.default_tags
}

resource "aws_iam_role_policy_attachment" "ecs_task_execution_role_policy" {
    role       = aws_iam_role.ecs_task_role.name
    policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"
}

resource "aws_iam_policy" "ecs_task_s3_access_policy" {
    name = "scpca-portal-ecs-task-s3-access-${var.user}-${var.stage}"
  
    policy = <<EOF
    {
        "Version": "2012-10-17",
        "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:GetObject",
                "s3:PutObject",
                "s3:ListBucket"
            ],
            "Resource": [
                "arn:aws:s3:::scpca-portal-inputs",
                # "arn:aws:s3:::output-bucket-name/*"
[...]

PR #899, opened 2024-09-19
844 - Add Batch Compute Environment

avrohomgottlieb:
*Note  - The base for this PR will be updated to `feature/batch` once `avrohom/clean-up-load-data` is merged into `feature/loader`, and `feature/loader` is merged into `feature/batch`

## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/844

Describe the big picture of your changes.

## Types of changes

## Functional tests

Additions in this PR
- `batch/` directory inside of `infrastructure/`
- `batch/compute_environment.tf` file which describes the compute environment
- `batch/roles.tf` which details necessary roles and their role policies for the compute environment
- `batch/security.tf`  which details the necessary security group for the compute environment

## Checklist

## Screenshots

N/A

### DIFF
--- a/infrastructure/batch/compute_environment.tf
+++ b/infrastructure/batch/compute_environment.tf
@@ -1,5 +1,5 @@
-resource "aws_batch_compute_environment" "scpca_portal" {
-  compute_environment_name = "scpca-portal-compute-${var.user}-${var.stage}"
+resource "aws_batch_compute_environment" "scpca_portal_project_compute" {
+  compute_environment_name = "scpca-portal-project-compute-${var.user}-${var.stage}"
 
   compute_resources {
     max_vcpus = 32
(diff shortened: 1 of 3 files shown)


infrastructure/batch/compute_environment.tf
```diff
@@ -0,0 +1,22 @@
+resource "aws_batch_compute_environment" "scpca_portal" {
```
davidsmejia, line 1: I think we will want to change `scpca_portal` here to something more specific. I do believe at this time we will want to allocate something smaller for ad hoc downloads in the future as there wont really be a benefit to throwing more vcpus at that. Perhaps something like `scpca_portal_project_compute` since this would be for generating downloadable files at the project level.

davidsmejia (commented):
This looks good, I'd just like to go over the naming before we merge this in. We don't need to have it 100% ironed out at this point but we should be mindful at least.

infrastructure/batch/compute_environment.tf
avrohomgottlieb: I updated this one, and the name of the ecs_task role. I think the rest are okay for now. My feeling is that we can always adjust things later on, but for now things look consistent.

davidsmejia (approved):
LGTM

Issue #903, opened 2024-09-20
Update test bucket to 2024-09-10


avrohomgottlieb:
### Context

In https://github.com/AlexsLemonade/ScPCA-admin/issues/957 and https://github.com/AlexsLemonade/ScPCA-admin/issues/958, there were updates made to the test data. 

### Solution or next step

We now need to update the test_bucket config variable to point to the new test bucket, located at `s3://scpca-portal-public-test-inputs/2024-09-10/`. As well, any values that are dependent on values from the old test bucket should be updated in accordance with the values in the new bucket.

Note: This issue cannot be completed until `feature/loader` is merged into `dev`.

avrohomgottlieb:
Instead of making the merging of `feature/loader` into `dev` a blocker for this issue, I'm going to make the updates in `feature/loader` itself. The changes will be propogated into `dev` when `feature/loader` is merged into `dev`.

It makes the most sense to update the test bucket in the place where the tests will be the most up to date, which is currently in `feature/loader`.

avrohomgottlieb:
Addressed by https://github.com/AlexsLemonade/scpca-portal/pull/908

Issue #880, opened 2024-09-06
Clean up Project::update_project_aggregate_properties


avrohomgottlieb:
### Context

At the end of `Project::load_metadata`, after a `project`, its `samples` and its `libraries` have been added to the db, `sample` and `project` aggregate values are updated (`samples` aggregate `library` values, and `projects` aggregate `sample` values).  `Project` aggregate logic is split up into two methods - `Project::update_project_modality_properties` and `Project::update_project_aggregate_properties`. 

### Problem or idea

We should look to optimize the `Project::update_project_aggregate_properties` method from a code readability perspective by pulling out the assignment of values from the `for sample in self.samples.all()` loop, and to assign them directly via querying for the different attribute values. The additional querying is not as performant as the current implementation of a single `for` loop, but the actual performance gain is negligible in our use case due the minimal set of attributes, which gives credence to a code readability optimization.

### Solution or next step

```python
    def update_project_aggregate_properties(self):
        """
        The Project and ProjectSummary models cache aggregated sample metadata.
        We need to update these after any project's sample gets added/deleted.
        """

        ...
        self.additional_metadata_keys = ", ".join(sorted(set(sample.additional_metadata_keys for sample in self.samples.all()), key=str.lower)
        self.diagnoses = ", ".join(sorted(set(sample.diagnosis for sample in self.samples.all()))
        ...

        self.save()
```

PR #908, opened 2024-09-24
903 - Update Test Bucket to 2024-09-10

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/903

This PR updates the test bucket from `scpca-portal-public-test-inputs/2024-07-19/` to `scpca-portal-public-test-inputs/2024-09-10/`.

In order to accommodate the change in how the `Project` attribute `diagnoses` is handled, a `PROJECT_METADATA_VALUES_TRANSFORMS` constant and `transform_values` helper function were added to `metadata_file`.

Values in the `expected_values/` files were also updated in order to match any changed values in the metadata files in the new test bucket.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/config/local.py
+++ b/api/scpca_portal/config/local.py
@@ -17,7 +17,7 @@ class Local(Common):
     AWS_REGION = None
 
     # AWS S3
-    TEST_INPUT_BUCKET_NAME = "scpca-portal-public-test-inputs/2024-07-19/"
+    TEST_INPUT_BUCKET_NAME = "scpca-portal-public-test-inputs/2024-09-10/"
     AWS_S3_INPUT_BUCKET_NAME = TEST_INPUT_BUCKET_NAME if Common.TEST else "scpca-portal-inputs"
     AWS_S3_OUTPUT_BUCKET_NAME = "scpca-local-data"
 
--- a/api/scpca_portal/metadata_file.py
+++ b/api/scpca_portal/metadata_file.py
@@ -28,6 +28,8 @@ PROJECT_METADATA_KEYS = [
     ("citation_doi", "doi", None),
 ]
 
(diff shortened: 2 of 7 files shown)


avrohomgottlieb:
@davidsmejia 
This PR is not very controversial, the one thing to discuss is the handling of the way that we update `Project.diagnoses`.
In light of https://github.com/AlexsLemonade/ScPCA-admin/pull/958/files, the `diagnoses` value in the `projects_metadata.csv` file is now correct. As such, we can now use it to populate `Project.diagnoses`, instead of what we used to do, which was to throw that value away and to collect the value from the aggregation of sample `diagnosis` values. 

I proposed transforming the semi-colon separated list which we get in the `project_metadatas.csv` file to our intended comma (and space) separated output inside of `metadata_file`, where we're already doing key transformations, but we can discuss other options as well. The direction should be that this value should be updated in `Project::get_from_dict`, like all other values pulled from the `projects_metadata.csv`, and not to be saved in `Project::update_project_aggregate_properties` at the end.

davidsmejia (approved):
LGTM

I do have some reservations about changing the source of truth for diseases on a project away from the aggregate evaluation. Mainly because I am unsure about possible states where the actual samples and the aggregate value from `projects_metadata.csv` may no longer be in sync.

This is not a problem and only a concern so I will open an issue to discuss this along with some other changes to `Project::update_project_aggregate_properties` for improving code reliability / consistency.

avrohomgottlieb:
> This is not a problem and only a concern so I will open an issue to discuss this along with some other changes to `Project::update_project_aggregate_properties` for improving code reliability / consistency.

Sounds good. I opened up an issue related to aggregate value updating a few weeks ago to have as a place holder until we would devise a plan of attack for this. I'll attach it here for reference.

https://github.com/AlexsLemonade/scpca-portal/issues/880

Issue #901, opened 2024-09-19
Add Batch Job Definition


avrohomgottlieb:
_Please label your issue with a label to specify its priority: breaking, high, medium, or low._

### Context

With the addition of the Batch compute environment was in https://github.com/AlexsLemonade/scpca-portal/issues/844, and the Batch job_queue in https://github.com/AlexsLemonade/scpca-portal/issues/900, we are now ready to develop the configuration for Batch job definitions.

### Problem or idea

_The context should lead to something, an idea or a problem that you’re facing. Put that here._

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

davidsmejia:
Closed with #901

PR #911, opened 2024-09-27
901 - Add Batch Job Definition

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/901

This PR introduces the `aws_batch_job_definition` resource.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/infrastructure/batch/job_definition.tf
+++ b/infrastructure/batch/job_definition.tf
@@ -1,0 +1,31 @@
+resource "aws_batch_job_definition" "generate_computed_files_job" {
+  name = "generate-computed-files-job-${var.user}-${var.stage}"
+  type = "container"
+
+  platform_capabilities = [
+    "FARGATE",
+  ]
+  container_properties = jsonencode({
+    command = ["sportal", "generate-computed-files", "--scpca-id", "<scpca-id>"]
+    image   = "ccdl/scpca_portal_api"
+
+    # t2.medium has 2 vcpus and 4.0 GB of RAM
+    vcpus  = 2
+    memory = 4000
+    volumes = [
+      {
+        host = "efs-volume"
+        efsVolumeConfiguration = {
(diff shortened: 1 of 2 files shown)


infrastructure/batch/job_definition.tf
```diff
@@ -0,0 +1,37 @@
...
+  name = "scpca-portal-project-job-definition-${var.user}-${var.stage}"
+  type = "container"
+
+  platform_capabilities = [
+    "FARGATE",
+  ]
+  container_properties = jsonencode({
+    # dispatch_to_batch command will override the below command during job submission, with
+    # cotaninerOverrides={"command": ["sportal", "generate-computed-files", "scpca-project-id", "<id>"]}
+    command = ["sportal", "generate-computed-files"]
```
davidsmejia, line 11:
```suggestion
    # command definition expected in cotaninerOverrides when using this job definition
    command = []
```
This file doesn't need to know about how we submit batch jobs as that is subject to change.

davidsmejia (commented):
We should solve for passing in environment variables in this PR.

Docs:
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/batch_job_definition#container_properties
- https://docs.aws.amazon.com/batch/latest/APIReference/API_RegisterJobDefinition.html
- Reference to convert object to name, value pairs: https://github.com/karvounis/terraform-aws-batch-container-properties/blob/master/main.tf

davidsmejia (commented):
Let's create an issue to set envars for the container from terraform env. Other than that this looks good after you remove the comment and update the provider version where it is now. Open an issue to move it in the future.

davidsmejia (approved):
LGTM let's open a new issue for provider version.

Issue #774, opened 2024-06-24
Improve Handling of Sample Metadata / Remove additional_metadata Construct


avrohomgottlieb:
### Context

Keys in the `samples_metadata.csv` file that are not represented in corresponding `Sample` model attributes are dumped into an `additional_metadata` `JSONField` attribute on the model. This attribute is used in a variety of different places, including being serialized and then returned to the client, and used to populate the `additional_metadata_keys` `TextField` attribute on the `Project` model.

### Problem or idea

When constructing the `Library` model, we took a different approach by explicitly persisting all raw inputs from library json files as a `metadata` `JSONField` attribute, and then pulling out the relevant attributes to persist as individual fields. This approach allowed us to be more generic in the way that metadata was handled, while also allowing us to be more agnostic to the science team's decision to add or delete certain metadata keys.

When looking at the way that we handle metadata on the `Sample` model, the approach is very much inverted, where we first populate the model attributes and thereafter dump the remaining keys into an `additional_metadata` model attribute.

### Solution or next step

We would like to transition the `Sample` model to be in line with the `Library` model. The goal would be to move towards classifying attributes as one of the following:
- Metadata (raw inputs from s3 metadata files)
- Model attributes (denormalized metadata keys)
- Aggregate values (metadata keys which are populated as a result of the bulk create of a related model)

At such a point, we would be able to more adequately distinguish between two contexts:
- Persistence (for querying)
- Presentation

The `metadata` attribute would be chiefly responsible for the writing of metadata files, and influence how `Sample::get_metadata` populates its returned dictionary. Instead of maintaining a list of `included_sample_attributes`, we would chose to invert the logic to be in line with the `Library` model and to maintain a list of `excluded_metadata_attributes`. We would then iterate over the `Sample.metadata` `JSONField` attribute to populate the `Sample::get_metadata` returned dictionary. This would remove the need for a second for loop like we currently have inside of `Sample::get_metadata`, where first model attributes are iterated over and added, and thereafter `additional_metadata`.

avrohomgottlieb:
Make the following additions to the following models:
- `Project` 
  - add a `@property` method which serializes `additional_metadata_keys` (called `additional_metadata_keys`) and delete the corresponding model attribute.
- `Sample`
  - add a `@property` method which grabs `additional_metadata` (called `additional_metadata`) and delete the corresponding model attribute.

avrohomgottlieb:
[...]

PR #912, opened 2024-09-29
774 - Remove Additional Metadata Construct

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/774

This PR removes the `Sample` model's `additional_metadata` attribute, replacing it with a `metadata` attribute that stores the input sample metadata file as a `JSONField`. An `additional_metadata` property method is added to derive the `additional_metadata` value previously stored in the database, which is necessary for presentation on the front end.

The `Sample::get_metadata` logic was also updated to conform to this new construct, inverting the approach from one that included a select group of predetermined attributes and excluded the rest, to one that excludes a predetermined group of metadata attributes and includes the rest (and described in more detail in the PR's connected issue).

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


avrohomgottlieb:
This PR was also meant to address the `Project` model's `additional_metadata_keys` attribute, but the approach has to be discussed further. 

Unlike with `additional_metadata` on the `Sample` model, which is populated with fields from `samples_metadata.csv` that have no corresponding model attribute, the `additional_metadata_keys` attribute on the `Project` model is made up of an aggregation of Sample metadata keys, and does not originate from the `projects_metadata.csv` file. 

In fact, there are even a field or two on the `projects_metadata.csv` file that we do not save to the db, and essentially throw away. We have to clarify our intended approach on whether or not we want to allow the science team to send through arbitrary fields for us to pass throw to our output file in the `Project` model like we want in the `Sample` model.

davidsmejia (commented):
LGTM

davidsmejia (commented):
Looking at the discrepancy between files sizes. We should build both and then compare to ensure that we aren't adding a new column.

Issue #862, opened 2024-08-23
Create config/test module, move common.py constants to config


avrohomgottlieb:
### Context

There are two locations where application-wide constants are stored, within `common.py` and within the `config/` directory.

The `config/` directory contains `config/local` and `config/production` modules, which both extend `config/common`, to allow for environment specific configurations with environment level constants. 

There is an `__init__` script inside the `config/` directory which contains logic for dynamically importing either module depending on the environment being used.

### Problem or idea

The philosophy that we'd like to more consistently enforce is that
- All environment level constants should live in the `config/` directory
- All application level constants should live in `common.py`

There are currently a handful of environment level constants living inside of `common.py` that we'd like to move over to their suitable locations within the `config` directory.

As well, a `config/test` module should be added, which extends `config/common` and removes the need for conditional logic inside of `common/local` to accommodate the testing environment more explicitly.

The `__init__` script should also be adjusted to add a clause for the dynamic importing of the `config/test` module during testing.

### Solution or next step

config/common.py
```python
...
class Common():
    ...
    # File path locations
    CODE_PATH = (
        code_path
        if (code_path := Path("/home/user/code")) and code_path.exists()
        else Path("/home/user")
    )
    DATA_PATH = CODE_PATH / "data"
    INPUT_DATA_PATH = DATA_PATH / "input"
    OUTPUT_DATA_PATH = DATA_PATH / "output"
    
    TEMPLATE_PATH = CODE_PATH / "scpca_portal" / "templates"

```

config/common.py
```python
...
class Local(Common):
    ...
    # AWS S3
    AWS_S3_INPUT_BUCKET_NAME = "scpca-portal-inputs"
    AWS_S3_OUTPUT_BUCKET_NAME = "scpca-local-data"
    ...
```

config/test.py
```python

import os

from scpca_portal.config.common import Common

class Test(Common):
    # File path locations
    DATA_PATH = CODE_PATH / "test_data"
    INPUT_DATA_PATH = DATA_PATH / "input"
    OUTPUT_DATA_PATH = DATA_PATH / "output"

    DEBUG = True

    # Mail
    EMAIL_HOST = "localhost"
    EMAIL_PORT = 1025
    EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"

    UPDATE_S3_DATA = False

    # AWS
    AWS_REGION = None

    # AWS S3
[...]

PR #914, opened 2024-10-02
862 - Create config/test Module

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/862

This PR adds a `config` module for the testing environment. It inherits from the `local` config module, but changes the `AWS_S3_INPUT_BUCKET_NAME` to the `test_bucket`, and updates the `DATA_PATH` attribute to point to `CODE_PATH / "test_data"`.

As well, this PR moves environment based constants from `common.py` to the `config` modules.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/config/local.py
+++ b/api/scpca_portal/config/local.py
@@ -27,10 +27,7 @@ class Local(Common):
     DEV_HOST = os.getenv("DEV_HOST")
 
     # Code Paths
-    CODE_PATH = Path("/home/user/code")
-
-    DATA_PATH = CODE_PATH / "data"
-    INPUT_DATA_PATH = DATA_PATH / "input"
-    OUTPUT_DATA_PATH = DATA_PATH / "output"
-
-    TEMPLATE_PATH = CODE_PATH / "scpca_portal" / "templates"
+    INPUT_DATA_PATH = Path("/home/user/code/data/input")
+    OUTPUT_DATA_PATH = Path("/home/user/code/data/output")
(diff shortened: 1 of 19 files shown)


api/scpca_portal/config/local.py
```diff
@@ -17,11 +18,19 @@ class Local(Common):
...
     DEV_HOST = os.getenv("DEV_HOST")
+
+    # Code Paths
+    CODE_PATH = Path("/home/user/code")
+
+    DATA_PATH = CODE_PATH / "data"
+    INPUT_DATA_PATH = DATA_PATH / "input"
+    OUTPUT_DATA_PATH = DATA_PATH / "output"
+
+    TEMPLATE_PATH = CODE_PATH / "scpca_portal" / "templates"
```
davidsmejia, line 36:
```suggestion
    INPUT_DATA_PATH = Path("/home/user/code/data/input")
    OUTPUT_DATA_PATH = Path("/home/user/code/data/input")
    TEMPLATE_PATH = Path("/home/user/code/scpca_portal/templates")
```

I think we will be better served by removing the path concats and then explicitly defining the location. This change would apply to the other config files as well. I think it made more sense to keep the intermediate variables when we were evaluating these but we don't benefit from that anymore.

davidsmejia (commented):
LGTM just want to clean up `CODE_PATH` and `DATA_PATH`. (edited)

davidsmejia (approved):
LGTM

Issue #922, opened 2024-10-07
Update demux_cell_count_estimate value for output metadata file


avrohomgottlieb:
### Context

We currently rename `demux_cell_count_estimate` to `sample_cell_estimate` when we're collecting Sample metadata in `Sample::get_metadata`. The decision was made to no longer rename this attribute for the output metadata file.

As well, we would like to assign this attribute's value in the output metadata file to be its corresponding value in the `sample_cell_estimates` field of the `Library` metadata file (which we have access to in `Library.metadata`). We don't plan on changing the way that the `demux_sample_cell_estimate` attribute is calculated in `Project::update_sample_derived_properties`, just the value in the output metadata file. This should be done in `Library::get_metadata`.

avrohomgottlieb:
Addressed in https://github.com/AlexsLemonade/scpca-portal/pull/928

PR #928, opened 2024-10-08
922 - Update demux_cell_count_estimate for metadata output

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/922

This PR updates the value of `demux_cell_count_estimate` presented on the output metadata file. 

Instead of grabbing the value off of the Sample model as was previously done, the `demux_cell_count_estimate` field in the output metadata file now grabs its value from the metadata attribute of its associated library. The output metadata file takes a more granular approach by grouping together Sample-Library combinations, making it more appropriate to pass through the Sample-Library demux_cell value than the aggregate 'all associated libraries cell count' value which lives on the Sample model.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


api/scpca_portal/models/library.py
```diff
@@ -198,9 +198,16 @@ def get_combined_library_metadata(self) -> List[Dict]:
...
+            # NOTE: demux_cell_count_estimate is reassigned here for the output metadata file.
+            # The metadata file writes out sample-library rows, necessitating greater granularity.
+            # In contrast, the Sample model expects the aggregate value of its related libraries.
+            if self.is_multiplexed:
+                metadata["demux_cell_count_estimate"] = self.metadata["sample_cell_estimates"][
+                    sample.scpca_id
+                ]
+            # Only multiplexed sample-libraries should have a demux_cell_count_estimate attribute
+            else:
+                del metadata["demux_cell_count_estimate"]
```
avrohomgottlieb, line 210: An alternative to deleting this here would be to just delete `demux_cell_count_estimate` from `derived_properties` in `Sample::get_metadata`. We're not treating it as a derived property here as we're grabbing the value from elsewhere, so it could be confusing to add it as a derived property only to repopulate it or delete it right thereafter.

davidsmejia: Yeah, let's remove it from the `Sample.get_metadata` `derived_attributes` since it should not exist in the metadata. Additionally, we should just rename `Sample.demux_cell_count_estimate` to be `demux_cell_count_estimate_sum`. Note we will need to update the client code as well so that the samples table renders correctly.

avrohomgottlieb: Files to be updated:
- Migration, serializer, expected values, client side references

avrohomgottlieb:
@nozomione wanted to ping you to check out the change to the `demux_cell_count_estimate` Sample attribute. On the Sample model we renamed it to `demux_cell_count_estimate_sum`, while we're using the name `demux_cell_count_estimate` to represent the metadata value. On the client side, the name doesn't change on the `ProjectSampleTable` itself, but it does in the `accessor` and `id` fields.

nozomione (approved):
👍

davidsmejia (approved):
LGTM

Issue #941, opened 2024-11-01
Debug Batch implementation for spinning up of dev stack


avrohomgottlieb:
### Context

Issue https://github.com/AlexsLemonade/scpca-portal/issues/939 (and its subsequent PR) dealt with adjustments to the deploy script which allowed for the spinning up of a dev stack, something which we'd like to do to test out our new Batch implementation. The deploy script was outdated and needed some tweaks to be brought up to date.

At this point, the deploy script is up to date, and a dev stack can be spun up.

### Next steps

Due to the fact that all of the Batch implementation terraform code has yet to be tested, the spinning up of the stack will be an iterate process of catching typos, referencing errors, and other bugs which have crept into the code while the Batch implementation was being written.

Next steps include going through these iterative processes and making the necessary fixes until arriving at a successful deploy.

As well, the naming conventions should be analyzed during this issue, to make sure that they are consistent within the Batch implementation and across the infrastructure files.

avrohomgottlieb:
Addressed in https://github.com/AlexsLemonade/scpca-portal/pull/945

PR #945, opened 2024-11-03
941 - Debug Batch implementation

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/941

This PR details the debugging changes made to the Batch implementation to produce workable infrastructure. 
Changes include:
- Adding a `batch.tf` module file to declare the batch subdirectory as a collection of related files
- Adding a `batch/variables.tf` file to enable resources and envars to make their way into the batch module
- Updating references and fixing typos to the current implementation

As well, this PR fixed all deprecations warnings by updating arguments and resources to their updated specifications.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/infrastructure/batch/job_definition.tf
+++ b/infrastructure/batch/job_definition.tf
@@ -79,7 +79,7 @@ resource "aws_batch_job_definition" "scpca_portal_project" {
       sizeInGib = 200
     }
 
-    executionRoleArn = aws_iam_role.ecs_task_execution_role.arn
+    executionRoleArn = aws_iam_role.ecs_task_role.arn
   })
 
   retry_strategy {
--- a/infrastructure/batch/job_definition.tf
+++ b/infrastructure/batch/job_definition.tf
@@ -16,7 +16,7 @@ resource "aws_batch_job_definition" "scpca_portal_project" {
       },
       {
         name  = "DJANGO_DEBUG"
-        value = false
+        value = "false"
       },
       {
         name  = "DJANGO_SECRET_KEY"
@@ -24,19 +24,19 @@ resource "aws_batch_job_definition" "scpca_portal_project" {
       },
       {
         name  = "DATABASE_HOST"
-        value = aws_db_instance.postgres_db.address
+        value = var.postgres_db.address
       },
       {
         name  = "DATABASE_PORT"
-        value = aws_db_instance.postgres_db.port
+        value = tostring(var.postgres_db.port)
       },
       {
         name  = "DATABASE_USER"
-        value = aws_db_instance.postgres_db.username
+        value = var.postgres_db.username
       },
       {
         name  = "DATABASE_NAME"
-        value = aws_db_instance.postgres_db.name
+        value = var.postgres_db.db_name
       },
       {
         name  = "DATABASE_PASSWORD"
@@ -48,7 +48,7 @@ resource "aws_batch_job_definition" "scpca_portal_project" {
       },
       {
         name  = "AWS_S3_BUCKET_NAME"
-        value = aws_s3_bucket.scpca_portal_bucket.id
+        value = var.scpca_portal_bucket.id
       },
       {
         name  = "SENTRY_DSN"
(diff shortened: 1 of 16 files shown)


avrohomgottlieb:
I'd like to propose a naming convention for resources in the `infrastructure/` directory.

Regarding the naming of resources, if there are multiple instances of a resource, then each instance should be given a name with a unique descriptor that distinguishes it some other instances of the same resource, such as `aws_cloudwatch_log_stream log_stream_api | log_stream_api_nginx_access | log_stream_api_nginx_error`. 

If there is only one instance of that resource, and we don't expect that there will be another, then it is recommended to name it according to the project, such as `aws_vpc scpca_portal`.

However, we should endeavor not to append the resource type to the end of the resource name, such as `aws_vpc scpca_portal_vpc`. The way that resources are referenced in terraform is by concatenating the resource type and the resource name with a dotted delimiter (dotted notation), so `aws_vpc.scpca_portal_vpc` is redundant.

A nuance with this convention should be with modules (like the `batch` module), where resources are only available if they are passed in as variables to the module. As such, the most straightforward way to name the variables should be to append the resource type to the resource name, which would look as follows.

```terraform
module "batch" {
    ...
    scpca_portal_vpc = aws_vpc.scpca_portal
}

**Comment by davidsmejia** [comment]
> I'd like to propose a naming convention for resources in the `infrastructure/` directory.
> 
> Regarding the naming of resources, if there are multiple instances of a resource, then each instance should be given a name with a unique descriptor that distinguishes it some other instances of the same resource, such as:
> 
> ```hcl
> resource "aws_subnet" "scpca_portal_1a" | "scpca_portal_1b"
> ```
> 
> If there is only one instance of that resource, and we don't expect that there will be another, then it is recommended to name it according to the project, such as:
> 
> ```hcl
> resource "aws_vpc" "scpca_portal"
> ```
> 
> However, we should endeavor not to append the resource type to the end of the resource name, such as `resource "aws_vpc" "scpca_portal_vpc"`. The way that resources are referenced in terraform is by concatenating the resource type and the resource name with a dotted delimiter (dotted notation), so `aws_vpc.scpca_portal_vpc` is redundant.
> 
> A nuance with this convention should be with modules (like the `batch` module), where resources are only available if they are passed in as variables to the module. As such, the most straightforward way to name the variables should be to append the resource type to the resource name, which would look as follows.
> 
> ```hcl
> module "batch" {
>     ...
>     scpca_portal_vpc = aws_vpc.scpca_portal
>     ...
> }
> ```

Thank you for suggesting this, could you remove this change from this PR and have it match what was previously defined. We should make that change in a separate effort.

File: infrastructure/batch/job_queue.tf [hunk 1]
```diff
@@ -3,11 +3,9 @@ resource "aws_batch_job_queue" "scpca_portal_project" {
...
   priority = 1
 
-  # compute_environments = [] array is deprecated
-  compute_environment_order {
-    order               = 1
-    compute_environment = aws_batch_compute_environments.scpca_portal_project.arn
-  }
+  compute_environments = [
+    aws_batch_compute_environment.scpca_portal_project.arn,
+  ]
```
avrohomgottlieb, line 8: `compute_environments` is currently deprecated. The recommended approach is using `compute_environment_order`. Using `compute_environment_order` necessitates upgrading to aws provider version "5.40", which I ran into bugs while trying to implement. In the interest of time I kept `compute_environments` for now, but this should be updated in the future.

infrastructure/batch/roles.tf
```diff
@@ -22,60 +23,59 @@ resource "aws_iam_role_policy_attachment" "batch_service_role" {
   policy_arn = "arn:aws:iam::aws:policy/service-role/AWSBatchServiceRole"
 }
 
-resource "aws_iam_role" "ecs_task_execution_role" {
+resource "aws_iam_role" "ecs_task_role" {
```
davidsmejia, line 26: can you change these names back as well, or let me know if it is a required change

infrastructure/batch/job_definition.tf
```diff
@@ -16,27 +16,27 @@ resource "aws_batch_job_definition" "scpca_portal_project" {
...
       },
       {
         name  = "DATABASE_HOST"
-        value = aws_db_instance.postgres_db.address
+        value = var.postgres_db.address
       },
       {
         name  = "DATABASE_PORT"
-        value = aws_db_instance.postgres_db.port
+        value = tostring(var.postgres_db.port)
```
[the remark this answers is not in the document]
avrohomgottlieb: For some reason, `tostring()` adequately casts the port variable into a `str`, but the string interpolation does not succeed. Here is the error output from the string interpolation:

image

davidsmejia (approved):
LGTM

Issue #943, opened 2024-11-03
Fix multiplexed for Batch


avrohomgottlieb:
### Context

The current way that we generate computed files for multiplexed samples was developed with the expectation of a multithreaded environment with all computed files being generated on the same machine. This current implementation is both insufficient and buggy for a distributed workload, which is a problem for us in our transition to AWS Batch. 

### Problem or idea

The current approach that we take is the introduction of a lock which inhibits multiple multiplexed samples from generating the sample computed file concurrently at the same location on disk, which would corrupt the file. In a distributed computing environment, where multiple samples would be sent to multiple machines, the implication would be that each multiplexed sample would create a computed file identical to the other multiplexed samples, with each one uploading the file to s3, where the last one would overwrite the previous ones. This is suboptimal behavior, and should be avoided. 

A better approach would be, for the Batch implementation, to develop a solution where only one multiplexed sample is sent off to generate the computed file while the other ones are skipped. A nuance which must be addressed is that this individual multiplexed sample must be responsible for creating computed_file objects for each of the other multiplexed samples, which would not be created.

### Solution or next step

See comment below.

avrohomgottlieb:
@davidsmejia I purposefully left the solution section of the Issue comment blank here because I believe that the simplest solution to the computed_file objects piece (discussed at the end of the **Problem or Idea** section) would be to change the relationship between Samples and ComputedFiles from one-to-many to many-to-many. This feels both more aligned with the actual relationship between the two models (with multiplexed in the picture), and would allow for the creation of a singular computed_file object to be created by the multiplexed sample that generates the computed file. There would be other consequences to such a change throughout the codebase (beyond just the migration) with the differences between how ManyToMany and ForeignKey relationships are accessed via the Django Query Language, but I think it would be manageable.

The other aspect of the solution would be to skip all but one multiplexed sample in the caller (dispatch_to_batch), which I don't think would be too difficult.

avrohomgottlieb:
As per a discussion in the Opening Sprint Meeting this morning, we are going to hold off on creating a ManyToMany relationship between the ComputedFile and Sample models at this point. 

We are going to move forward, however, with removing the locking mechanism from the batch and multithreaded implementations by having the caller only send one multiplexed samples through and letting it create the computed file objects for all related multiplexed samples.

avrohomgottlieb:
Addressed in PR https://github.com/AlexsLemonade/scpca-portal/pull/946

PR #946, opened 2024-11-05
943 - Fix Multiplexed for Batch

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/943

This PR addresses the suboptimal implementation for multiplexed sample computed files as it pertains to AWS Batch. As mentioned in the connected issue, the current approach utilizing a lock is only possible on a single machine where multiple threads are vying for compute resources. When each sample and download_config is broken up into a separate job, multiplexed samples will be recomputed and re-uploaded on each machine that the job runs on, which is wasteful and unnecessary.

This PR makes the following changes:
- Adds a Project::all_samples_no_multiplexed_duplicates property method which allows for querying of all non multiplexed samples and one multiplexed sample per multiplexed library in the caller.
- Removes the locking mechanism all together for multiplexed sample computed file generation
- Adds a ComputedFile::bulk_create_multiplexed_files class method which adds all of a representative multiplexed sample's associated computed files to the db
- Updates test_loader in light of the changes

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/loader.py
+++ b/api/scpca_portal/loader.py
@@ -1,7 +1,6 @@
 import shutil
 from concurrent.futures import ThreadPoolExecutor
 from functools import partial
-from threading import Lock
 from typing import Any, Dict, List, Set
 
 from django.conf import settings
@@ -208,9 +207,7 @@ def generate_computed_files(
         update_s3=update_s3,
         clean_up_output_data=clean_up_output_data,
     )
-    # Prepare a threading.Lock for each sample, with the chief purpose being to protect
-    # multiplexed samples that share a zip file.
-    locks = {}
+
     with ThreadPoolExecutor(max_workers=max_workers) as tasks:
         # Generated project computed files
         for config in common.GENERATED_PROJECT_DOWNLOAD_CONFIGS:
@@ -223,12 +220,10 @@ def generate_computed_files(
         # Generated sample computed files
         for sample in project.all_samples_no_multiplexed_duplicates:
             for config in common.GENERATED_SAMPLE_DOWNLOAD_CONFIGS:
-                sample_lock = locks.setdefault(sample.get_config_identifier(config), Lock())
                 tasks.submit(
                     ComputedFile.get_sample_file,
                     sample,
                     config,
-                    sample_lock,
                 ).add_done_callback(on_get_file)
 
     project.update_downloadable_sample_count()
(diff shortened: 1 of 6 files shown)


avrohomgottlieb:
If we like this approach, then I would update `dispatch_to_batch` to match loader::generate_computed_files (i.e. update the caller). Would like to get consensus first, before proceeding.

api/scpca_portal/loader.py
```diff
@@ -150,13 +149,14 @@ def _create_computed_file(
...
-        if clean_up_output_data:
-            computed_file.clean_up_local_computed_file()
+    if update_s3:
+        s3.upload_output_file(computed_file.s3_key, computed_file.s3_bucket)
+    if clean_up_output_data:
+        computed_file.clean_up_local_computed_file()
+
     computed_file.save()
+    if computed_file.sample and computed_file.has_multiplexed_data:
+        ComputedFile.bulk_create_multiplexed_files(computed_file)
```
davidsmejia, line 159: We should enforce a state where all related multiplexed samples either have a computed file or not. We can do that by bulk_creating them instead of saving when the conditions are met.

```suggestion
    if computed_file.sample and computed_file.has_multiplexed_data:
       ComputedFile.bulk_create(
           computed_file.get_multiplexed_computed_files()
       )
    else:
      computed_file.save()
```

```python
def get_multiplexed_computed_files(self):
    computed_files = [self]
    for sample in self.sample.multiplexed_with_samples:
        sample_computed_file = ComputedFile(model_to_dict(computed_file))
        sample.computed_file.pk = None
        sample.computed_file.sample = sample
        computed_files.append(sample_computed_file)
    return computed_files
```

api/scpca_portal/models/computed_file.py
```diff
@@ -354,6 +349,23 @@ def metadata_file_name(self):
...
         return settings.OUTPUT_DATA_PATH / self.s3_key
 
+    def get_multiplexed_computed_files(self):
+        """
+        Return computed file objects for all associated multiplexed samples.
+        The self computed file object is used as a template in building identical computed files,
+        differentiated only by sample, so that they can be bulk created at the same time.
+        """
+        multiplexed_samples = list(self.sample.multiplexed_with_samples) + [self.sample]
+        self.sample = None
```
avrohomgottlieb, line 359: The reason why I went in this direction is because `model_to_dict` doesn't gracefully handle Foreign Keys. What it does is pass (in this case) the sample_id through, which causes an error when creating a new object because the Django ORM is expecting an object and not an id (integer). Combining the samples to iterate over all of them and dropping the sample from the self object allows for the self object to function as a template from which all associated computed_files can be built.

davidsmejia (approved):
LGTM

Issue #944, opened 2024-11-03
Generate computed files via Batch on dev stack


avrohomgottlieb:
### Context and Next Steps

At the conclusion of https://github.com/AlexsLemonade/scpca-portal/issues/941, the infrastructure will be complete for the new AWS Batch implementation, and we will be able to begin testing out the job submission logic. This will include kicking off the `dispatch_to_batch` command, starting with individual projects and working the way up to the entire portal. We should begin with projects which address specific modality/format configurations, and then address any challenges that come up in separate issues/PRs. 

The progression of dispatching to batch should be as follows
- Single-cell project
- Spatial project
- Multiplexed project
- All projects

Upon successful job submission and job completion, there should be a subset of computed file attributes used to determine the validity and correctness of the computed files generated to make sure that the have been generated correctly.

Closing this ticket ultimately means that the outcomes of these jobs are satisfactory.

avrohomgottlieb:
One area of concern is whether or not the 200 GB ephemeral storage limit on job instances is too small for our maximum job sizes. Refactoring or taking a different approach would be necessary at this point.

avrohomgottlieb:
> One area of concern is whether or not the 200 GB ephemeral storage limit on job instances is too small for our maximum job sizes. Refactoring or taking a different approach would be necessary at this point.

I'm happy to say that the 200 GB ephemeral storage limit proved itself not to be a problem with our current set of projects. The largest project and download_config combination is SCPCP000008 with Single-Cell AnnData, which has a download size of 75.56GB. The projected ceiling for this project should be around the 150GB mark, with the file system containing one copy of each project, one as it was downloaded and the other in the zip archive.

If we end up with jobs which get closer to that 200GB storage limit, one optimization would be to be more granular with how we carry out lazy downloading. The solution would be - as opposed to downloading all relevant data files at the beginning of `ComputedFile::get_project_file`, we could download each data file one at a time, include the file into the archive, and then delete it before downloading the next one.

PR #947, opened 2024-11-10
Debug dispatch_to_batch

avrohomgottlieb:
## Issue Number

Related to Issue https://github.com/AlexsLemonade/scpca-portal/issues/944

After debugging the Batch implementation in PR https://github.com/AlexsLemonade/scpca-portal/pull/945, the final step in order to complete the Batch functionality was to debug the batch job submission workflow in the `dispatch_to_batch` management command. This PR fixed a handful of bugs in the implementation, tweaking the command itself as well as the underlying batch infrastructure.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


avrohomgottlieb:
One bug that was discovered is that the input library metadata file is meant to be included in the output computed file amount the other data files for spatial downloads. When the loading of metadata and generation of computed files happen on the same machine, as has been the case up until now, metadata files are loaded when objects are being created and data files are loaded during computed file generation. However, because the loading of metadata and the generation of computed files happen now on separate machines (the former on the API and the latter on fargate instances through batch), the metadata file wasn't being included with the other data files in the output computed file. I've addressed this bug in my most recent commit.

api/scpca_portal/management/commands/dispatch_to_batch.py
```diff
@@ -28,19 +31,23 @@ def handle(self, *args, **kwargs):
...
         resource_id = project_id if project_id else sample_id
-        job_name = f"{resource_id} - {download_config_name}"
+        job_name = f"{resource_id}-{download_config_name}"
 
         response = batch.submit_job(
             jobName=job_name,
-            jobQueue="scpca_portal_project",
-            jobDefinition="scpca_portal_project",
+            jobQueue=os.environ.get("AWS_BATCH_JOB_QUEUE_NAME"),
+            jobDefinition=os.environ.get("AWS_BATCH_JOB_DEFINITION_NAME"),
```
davidsmejia, line 50: Let's move this to settings and import from there.

bin/sportal
```diff
@@ -35,6 +35,9 @@ commands = {
     "create-portal-metadata": run_api.format("./manage.py create_portal_metadata {}"),
     "load-data": run_api.format("./manage.py load_data {}"),
     "mock-data": run_api.format("./manage.py load_mock_data {}"),
+    "load-metadata": run_api.format("./manage.py load_metadata {}"),
+    "generate-computed-files": run_api.format("./manage.py generate_computed_files {}"),
+    "dispatch-to-batch": run_api.format("./manage.py dispatch_to_batch {}"),
```
davidsmejia, line 40:
```suggestion
```

We shouldn't try to dispatch to batch from local via sportal

api/scpca_portal/models/library.py
```diff
@@ -91,13 +91,13 @@ def get_data_file_paths(cls, data: Dict, s3_input_bucket: str) -> List[Path]:
...
-        data_file_paths = [
-            file_path
-            for file_path in s3.list_input_paths(relative_path, s3_input_bucket)
-            if "metadata" not in file_path.name
-        ]
+        file_paths = s3.list_input_paths(relative_path, s3_input_bucket)
+
+        # input metadata json is included in spatial downloads, but excluded from all others
+        if Library.get_modality_from_file_paths(file_paths) == Library.Modalities.SPATIAL:
+            return file_paths
```
davidsmejia, line 98:
```suggestion
        # input metadata json is excluded from single_cell downloads
        if Library.get_modality_from_file_paths(file_paths) == Library.Modalities.SINGLE_CELL:
            return [file_path for file_path in file_paths if "metadata" not in file_path.name]
```

Inverting the exception here would make it easier to maintain. This allows you to pair the logic for the exception and the return value in the same statement.

api/scpca_portal/models/library.py
```diff
@@ -91,13 +91,13 @@ def get_data_file_paths(cls, data: Dict, s3_input_bucket: str) -> List[Path]:
...
 
-        return data_file_paths
+        return [file_path for file_path in file_paths if "metadata" not in file_path.name]
```
davidsmejia, line 100:
```suggestion
        return file_paths
```

bin/sportal
```diff
@@ -35,6 +35,9 @@ commands = {
     "create-portal-metadata": run_api.format("./manage.py create_portal_metadata {}"),
     "load-data": run_api.format("./manage.py load_data {}"),
     "mock-data": run_api.format("./manage.py load_mock_data {}"),
```
davidsmejia, line 37: can you delete this while you are in here, this command doesn't exist anymore

api/scpca_portal/config/production.py
```diff
@@ -16,12 +16,16 @@ class Production(Common):
     UPDATE_S3_DATA = True
 
     # AWS
-    AWS_REGION = os.getenv("AWS_REGION")
+    AWS_REGION = os.getenv("AWS_REGION", "us-east-1")
```
davidsmejia: Let's move this to common.

davidsmejia (approved):
LGTM

PR #948, opened 2024-11-12
Production Deploy

davidsmejia:
# Production Deploy

## API
- Pins Dockerfile to 3.10.14 from 3.10
- Updates how unavailable and downloadable sample counts are calculated. The results from tabulation are equivalent.

## Client
- No Change


nozomione (approved):
👍

Issue #964, opened 2024-11-18
Address terraform and aws version incompatibility in GHA


avrohomgottlieb:
### Context

PR https://github.com/AlexsLemonade/scpca-portal/pull/953 merged feature/batch into dev. The changes made in feature/batch included adding new Batch infrastructure as well as adding management commands to submit jobs to Batch. In order to accommodate one of the arguments in one of the Batch resources, we had to update the aws_version in terraform to 5.12.0. The version of terraform that we currently use is 0.12.26.

### Problem or idea

While there shouldn't be a problem using aws_version 5.12.0 with terraform version 0.12.26, we encountered an error when the GHA runner attempted to deploy a staging stack. See [failed Github Action](https://github.com/AlexsLemonade/scpca-portal/actions/runs/11862498721/).

The GHA warning which led to the failure was as follows:
> [!WARNING]  
> Warning: registry.terraform.io: This version of Terraform has an outdated GPG key and is unable to verify new provider releases. Please upgrade Terraform to at least 0.12.31 to receive new provider updates. For details see: https://discuss.hashicorp.com/t/hcsec-2021-12-codecov-security-event-and-hashicorp-gpg-key-exposure/23512

The reason why this error didn't surface before now is because GHA only spins up a terraform instance and makes deploys on merges to dev and main. While my local version of terraform that I installed through homebrew is a newer one (I wasn't able to install 0.12.x on homebrew because it's too old), the documentation that I relied on said that coupling terraform 0.12.x with aws version 5.12.0 shouldn't be a problem. This being said, but it looks like there's a GHA component that I didn't anticipate (the GPG key?) that could be the cause of the error.

### Solution or next step
It seems like the clearest way forward would be to upgrade the terraform version. We could either do this as a short term patch, like upgrading to 0.12.31 (as mentioned in the GHA warning), or to go for something more modern (the current version is 1.9.8).

avrohomgottlieb:
Now that both the staging and prod have been successfully updated to terraform 1.0.8 and aws_provider version 5.x, this issue can be closed.

PR #969, opened 2024-11-21
Downgrade to Terraform 0.13.0 in Github GHA workflows

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/964

There is an intermediate step when upgrading from TF 0.12.x to 1.x.x.
TF will automatically apply `terraform 0.13upgrade`, which should resolve the `Invalid legacy provider address` error.

First we'll upgrade to 0.13.x, and then transition to 1.x.x.

Discussion here - https://discuss.hashicorp.com/t/unqualified-provider-aws/18554
Upgrade steps -https://developer.hashicorp.com/terraform/language/v1.1.x/upgrade-guides/0-13

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/.github/workflows/deploy_prod_backend.yml
+++ b/.github/workflows/deploy_prod_backend.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Setup Terraform
         uses: hashicorp/setup-terraform@v3
         with:
-          terraform_version: 1.9.8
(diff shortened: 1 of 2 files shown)


davidsmejia (approved):
LGTM

PR #992, opened 2024-11-22
add unlock_state and use it to unlock failed command

davidsmejia:
## Issue Number

N/A

- Adds functionality to unlock a specific state lock in terraform
- passes the current lock to ignore so we can unblock ourselves
- this should be removed from deploy before going to prod

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


nozomione (commented):
👍

PR #1002, opened 2024-11-22
handle nested value in terraform output -json

davidsmejia:
## Issue Number

N/A

- Outputs are now nested so we need to grab the value instead of the whole object.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


nozomione (approved):
👍

PR #1019, opened 2024-12-08
Hot Fix - dispatch_to_batch is null bug

avrohomgottlieb:
## Issue Number

N/A

The correct null check for a foreign key is via `isnull`, instead of `is_null`. This PR makes fixes this error in `dispatch_to_batch`.

## Types of changes

## Functional tests

N/A

## Checklist

N/A

## Screenshots

N/A


davidsmejia (approved):
We could add a test to cover this as well.

PR #1034, opened 2024-12-20
1028 - Sync Original Files - Object Creation

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1028

- Add `sync_original_files` management command
- Add `s3::list_bucket_objects` function
- Add `OriginalFile::sync` function, which calls `OriginalFile::bulk_create_from_dicts` (and will call `OriginalFile::bulk_update_from_dicts` and `OriginalFile::purge_deleted_files` in the future)
- Move dict transformation functions from `metadata_file` to `utils`

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


avrohomgottlieb:
@davidsmejia 

Two questions about this PR:
1. Should we process all attributes on `get_from_dict` and determine calculated values with static methods (like we do in the `Library` and `ComputedFile` model), or should we elect to save instances with all known properties in `get_from_dict` and add computed_values via instance methods after the fact (like we do in the `Project` model)? I took the first approach, but we're not bound by it.
2. What is considered a "single_cell" or "spatial" original file? Are we considering all original files associated with a single_cell library to be `is_single_cell=True` (i.e. even .html and .json files), or only the library's object files (.rds, .h5ad)? And if we say want to classify only the data files as such, how does that work with spatial when there aren't any clear cut data files?

davidsmejia (approved):
LGTM

Issue #1029, opened 2024-12-16
Sync OriginalFiles - Object Updating and Deletion 


avrohomgottlieb:
### Context

Issue https://github.com/AlexsLemonade/scpca-portal/issues/1024 explained the approach to populating and syncing the OriginalFiles. The sequence explained as follows:

> The `OriginalFile` table should be built by making a single call to aws to list all files in the s3 input bucket, and iterating over the results to create `OriginalFile` instances. The most straightforward way to do this would be to put this logic into a management command, called `sync_original_files`, which would be called at regular intervals to keep the database and s3 input bucket in sync.
> The sequence of functions calls to accomplish this should be:
> - `s3::list_input_paths`
> - `OriginalFile::sync_files`
> - `OriginalFile::bulk_update_from_dicts`
> - `OriginalFile::bulk_create_from_dicts`
>    - `OriginalFile::get_from_dict` 
>  - `OriginalFile::purge_deleted_files`

This issue will focus on the updating and deletion of OriginalFiles, including
 - `OriginalFile::sync_files`
  - `OriginalFile::bulk_update_from_dicts`
  - `OriginalFile::purge_deleted_files`

### Solution or next step

As mentioned in issue https://github.com/AlexsLemonade/scpca-portal/issues/1024, the above methods will be implemented as follows:
> - `OriginalFile::sync_files(input_paths: List[Path])`
>    - Iterates over the data dicts and calls `OriginalFile::bulk_update_from_dicts`, `OriginalFile::bulk_create_from_dicts` , and `OriginalFile::purge_deleted_files` 
>.   - Generates a timestamp before calling above methods, adds timestamp to `last_sync` attribute of all file objects
> - `OriginalFile::bulk_update_from_dicts`
>    - Updates `last_sync` attribute on all files
>    - For modified files, updates hash (etag), size_in_bytes, and other affected attributes
>    - Objects are updated and saved using `OriginalFile.objects.bulk_update`
> - `OriginalFile::purge_deleted_files`
>    - Queries all files in the DB that don't have the newest `last_sync` value, and deletes them
>    - For now we're going to stub out this method and just throw an error if a stale file was not included in the `s3::list_input_files` returned list.

PR #1035, opened 2024-12-22
1029 - Sync Original Files - Object Updating and Deletion

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1029

- Add bulk updating to modify existing original files
- Add purging of all files that no longer exist on s3

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


avrohomgottlieb:
This current implementation does not differentiate between unmodified files which existed on s3 the last time we ran the command and those that have been modified.  They both need their timestamps updated, which is carried out here, but the mechanism to alert us when a file has been updated needs to be added in when we decide what that mechanism should be.

api/scpca_portal/models/original_file.py
```diff
@@ -84,7 +84,59 @@ def bulk_create_from_dicts(cls, file_objects, bucket, sync_timestamp):
...
+                    file_object, sync_timestamp
+                )
+                updatable_original_files.append(updated_instance)
+                fields.update(updated_fields)
+
+        if not updatable_original_files:
+            return []
+
+        OriginalFile.objects.bulk_update(updatable_original_files, fields)
+        return updatable_original_files
```
davidsmejia, line 128:
```suggestion
    @classmethod
    def bulk_update_from_dicts(
        cls, file_objects: List[Dict], bucket: str, sync_timestamp
    ) -> List[Self]:
        # keep track of two lists, one for files that actually changed
        touched_original_files = []
        modified_original_files = []
        fields = set()

        for file_object in file_objects:
            if original_instance := OriginalFile.objects.filter(
                s3_bucket=bucket, s3_key=file_object["s3_key"]
            ).first():
                updated_instance, updated_fields = original_instance.update_existing(
                    file_object, sync_timestamp
                )
                touched_original_files.append(updated_instance)
                fields.update(updated_fields)
                
                # we actually updated the file
                if len(fields.difference({"bucket_sync_at"})) > 1 :
                    modified_original_files.append(updated_instance)

        if not touched_original_files:
            return []

        OriginalFile.objects.bulk_update(touched_original_files, fields)
        return touched_original_files, modified_original_files
```

I think we should probably break up the things we are touching and the things we are actually changing. This way we can have more useful logs.

api/scpca_portal/management/commands/sync_original_files.py
```diff
@@ -22,11 +25,44 @@ def add_arguments(self, parser):
...
         self.sync_original_files(**kwargs)
 
+    def generate_formatted_file_string(self, files: List[OriginalFile], file_type: str) -> str:
+        if not files:
+            return ""
+
+        formatted_file_str = f"{file_type}:\n"
+        formatted_file_str += "\n".join(f"- {str(f)}" for f in files) if files else "- None"
+        formatted_file_str += "\n"
+        return formatted_file_str
```
davidsmejia, line 35: Can we remove the `file_type` from this function so it just returns the multiline list of files or "None". Also, we shouldn't prefix the newline inside this method if we do change it and have the caller handle that.

api/scpca_portal/management/commands/sync_original_files.py
```diff
@@ -13,20 +16,49 @@
...
 
     def handle(self, *args, **kwargs):
         self.sync_original_files(**kwargs)
 
-    def sync_original_files(self, bucket_name: str, **kwargs):
+    def get_indented_files(self, files: List[OriginalFile]) -> str:
+        formatted_file_str = "\n".join(f"\t{str(f)}" for f in files) if files else "\tNone"
+        return formatted_file_str
+
+    def log_file_changes(self, updated_files, created_files, deleted_files, sync_timestamp) -> None:
```
nozomione, line 35: I noticed this method is missing typing. Just a small suggestion to match the others: 

```suggestion
def log_file_changes(self, updated_files: List[OriginalFile], created_files: List[OriginalFile], deleted_files: List[OriginalFile], sync_timestamp: datatime) -> None:
```

nozomione (approved):
I've left a small suggestion. Otherwise, this PR looks great 👍!

Issue #1025, opened 2024-12-11
Add test_dispatch_to_batch Module


avrohomgottlieb:
### Context

There are a few conditions that we'd like to test in the `dispatch_to_batch` module. Among them include:
- The querying of projects without computed files when a project id is not passed.
- That project and sample jobs are being submitted correctly

### Solution or next step

A `test_dispatch__to_batch` module should be created to test the above conditions.

To test the querying of projects without computed files we should create two projects, one with computed files and the other without, and we should make sure that the correct one is queried.

PR #1042, opened 2025-01-03
1025 - Add test_dispatch_to_batch

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1025

Add a test module for `dispatch_to_batch` which tests correct project querying and job submission.

Update `ProjectFactory` factory method to adequately handle the library and sample relationship.

## Types of changes

## Functional tests

- `test_dispatch_to_batch`
  - `test_only_projects_without_computed_files_submitted`
  - `test_project_id_submission`

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/management/commands/dispatch_to_batch.py
+++ b/api/scpca_portal/management/commands/dispatch_to_batch.py
@@ -20,11 +20,15 @@ class Command(BaseCommand):
     help = """
     Submits all computed file combinations to the specified AWS Batch job queue
     for projects for which computed files have yet to be generated for them.
-    If a project-id is passed, then computed files are only submitted for that specific project.
+    If regenerate-all is passed, then all computed files are regenerated for all projects.
+    If a project-id is passed, and not combined with regenerate-all,
+    then computed files are only submitted for that specific project if it has no computed files.
+    If it is combined with regenerate-all, then computed files are regenerated regardless.
     """
 
     def add_arguments(self, parser):
-        parser.add_argument("--project-id", type=str)
+        parser.add_argument("--regenerate-all", type=bool, default=False)
+        parser.add_argument("--project-id", type=str, default="")
 
     def handle(self, *args, **kwargs):
         self.dispatch_to_batch(**kwargs)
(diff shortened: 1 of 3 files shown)


davidsmejia (commented):
This looks mostly good, I think that we should also ensure that sample files are being generated until we are no longer supporting that use case.

api/scpca_portal/test/factories.py
```diff
@@ -169,4 +169,15 @@ class Meta:
 class ProjectFactory(LeafProjectFactory):
     computed_file1 = factory.RelatedFactory(ProjectComputedFileFactory, "project")
     sample1 = factory.RelatedFactory(SampleFactory, "project")
+    library1 = factory.RelatedFactory(LibraryFactory, "project")
     summary1 = factory.RelatedFactory(ProjectSummaryFactory, factory_related_name="project")
+
+    @factory.post_generation
+    def add_sample_library_relation(self, create, extracted, **kwargs):
+        if not create:
+            return
```
davidsmejia, line 178: Add a doc string that explains the many to many consideration here

api/scpca_portal/test/management/commands/test_dispatch_to_batch.py
```diff
@@ -0,0 +1,70 @@
...
+        self.assertNotIn(project_with_files.scpca_id, submitted_project_ids)
+
+        sample_no_files = project_no_files.samples.first()
+        submitted_sample_ids = set(
+            call.kwargs.get("sample_id") for call in mock_submit_job.call_args_list
+        )
+        self.assertIn(sample_no_files.scpca_id, submitted_sample_ids)
+
+    @patch("scpca_portal.management.commands.dispatch_to_batch.Command.submit_job")
+    def test_all_projects_have_computed_files(self, mock_submit_job):
```
davidsmejia, line 37: assert that computed files exist here

api/scpca_portal/test/management/commands/test_dispatch_to_batch.py
```diff
@@ -0,0 +1,70 @@
...
+        ProjectFactory()
+
+        self.dispatch_to_batch()
+        mock_submit_job.assert_not_called()
+
+    @patch("scpca_portal.management.commands.dispatch_to_batch.Command.submit_job")
+    def test_project_id_submission(self, mock_submit_job):
```
davidsmejia, line 44:
```suggestion
    def test_project_id_argument(self, mock_submit_job):
```

davidsmejia (commented):
I think the `test_project_id_submission` exposes opaqueness in how `dispatch_to_batch` handles existing computed files.

I think we should add 2 new arguments to the command:
- `--add-missing': to handle if we want to only generate computed files that don't exist
- `--recreate-all`: combined with `--project-id` will not call submit_jobs if a computed file exists on the project

davidsmejia:
Can you also add a log to the end of the dispatch to batch command that summarizes how many jobs were sent.

api/scpca_portal/management/commands/dispatch_to_batch.py
```diff
@@ -62,29 +68,41 @@ def submit_job(
...
         projects = (
             Project.objects.filter(project_computed_files__isnull=True)
-            if not project_id
-            else Project.objects.filter(scpca_id=project_id)
+            if not regenerate_all
+            else Project.objects.all()
         )
 
+        if project_id:
+            projects.filter(scpca_id=project_id)
```
davidsmejia, line 83: This doenst look right to me, my understanding is that querysets are immutable. I would first try to make your tests fail with this implementation then fix this to overwrite the `projects` queryset.

davidsmejia (approved):
LGTM

davidsmejia:
I think that it might be good to discuss when we should add custom assertion failure messages for specific assertion categories like "Factory data not as expected. Cannot complete tests" to help separate the assertions for the test itself

PR #1052, opened 2025-01-14
1046 - Add Dataset model 

### LINKED ISSUE #1046: Add Dataset model
### Context

To prepare the API for cart functionality, we need to create a `Dataset` model.

This issue defines the initial model schema by bootstrapping the model structure, which allows us integrate it into other parts of the process.

### Problem or idea

The `Dataset` model will consist of the following field types:

- User editable fields
- On-request fields
- Dataset status fields
- Process information fields
- Dynamic model properties

The initial model should contain the following fields:

```py
class Dataset(TimestampModel):

    # Choices matched with the ComputedFile/Library models
    class FileFormats:
        ANN_DATA = "ANN_DATA"
        SINGLE_CELL_EXPERIMENT = "SINGLE_CELL_EXPERIMENT"

        CHOICES = (
            (ANN_DATA, "AnnData"),
            (SINGLE_CELL_EXPERIMENT, "Single cell experiment"),
        )
    
    id: UUIDField()

    # User editable fields (with validation)
    format: TextField(choices=FileFormats.CHOICES)
    data: JSONField(default={})   
    regenerated_from: ForeignKey('self', null=True) # validation required
    email: EmailField(null=True) # validation required
    start: BooleanField(null=True) # validation required (with a serializer)

    # On-request fields - values assigned on requests
    token: OneToOneField(Token) # Used to process the dataset
    download_tokens: ManyToManyField(Token) # Used to create download urls

    # Dataset status fields - values set during the processing workflow
    started_at: DateTimeField(null=True)
    is_started: Boolean(default=False)
    is_processing: Boolean(default=False)

    processed_at: DateTimeField(null=True)
    is_processed: Boolean(default=False)
            
    expires_at: DateTimeField(null=True)
    is_expired: Boolean(default=False) # If expires_at exists and has passed
    
    errored_at: DateTimeField(null=True)
    is_errored: Boolean(default=False)
    error_message: TextField(null=True)

    computed_file: OneToOneField(ComputedFile)
```

Other anticipated fields to be implemented later that require further discussion:

```py
    # Implement later
    # Process information fields        
    api_release: CharField() # A commit hash from Docker
    last_edit_hash: CharField() # A hash for original files when the data attribute was last modified
    input_bucket_sync: DateTimeField() # The last time the original files were synced with the input bucket
    job_id: CharField() # The batch job ID

    # Dynamic model properties
    @property
    stats: Dict # For client UI presentation
    original_files # The result of getting the original files from the data attribtue 
```

### Solution or next step

Based on the above requirements:
- Implement the `Dataset` model
- Generate the migration to apply the updates

nozomione:
## Issue Number

Closes #1046

Changes include:
- Implemented the initial `Dataset` model
- Applied the migration to update the database schema
- To avoid reverse query name clashes:
  - Prefixed `related_name` with `dataset_` for the following fields:
    - `computed_file` 
    - `download_tokens`
    - `token` 
  - Set `related_name` to `original_dataset` for `regenerated_from`
- Added `help_text` to fields that are not fully self-explanatory 

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+    start = models.BooleanField(
+        null=True,
+        help_text="Indicates if the dataset process has started.",
+    )
+
+    token = models.OneToOneField(
+        APIToken,
+        null=True,
+        on_delete=models.CASCADE,
+        related_name="dataset_token",
```
nozomione, line 44: To avoid a reverse query name clash, I’ve prefixed the `related_name` with "dataset_".

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+        help_text="Token used to process the dataset.",
+    )
+    download_tokens = models.ManyToManyField(
+        APIToken,
+        related_name="dataset_download_tokens",
```
nozomione, line 49: To avoid a reverse query name clash, I’ve prefixed the `related_name` with "dataset_".

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+    is_expired = models.BooleanField(default=False)
+    errored_at = models.DateTimeField(null=True)
+    is_errored = models.BooleanField(default=False)
+    error_message = models.TextField(null=True)
+
+    computed_file = models.OneToOneField(
+        ComputedFile,
+        null=True,
+        on_delete=models.CASCADE,
+        related_name="dataset_computed_file",
```
nozomione, line 68: To avoid a reverse query name clash, I’ve prefixed the `related_name` with "dataset_".

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+    )
+    start = models.BooleanField(
+        null=True,
+        help_text="Indicates if the dataset process has started.",
+    )
+
+    token = models.OneToOneField(
+        APIToken,
+        null=True,
+        on_delete=models.CASCADE,
```
avrohomgottlieb, line 43:
```suggestion
        on_delete=models.SET_NULL,
```

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+        related_name="dataset_token",
+        help_text="Token used to process the dataset.",
+    )
+    download_tokens = models.ManyToManyField(
```
avrohomgottlieb, line 47: Why do we think this is a ManyToManyField and not a ForeignKey (on `APIToken`)?

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+    regenerated_from = models.ForeignKey(
+        "self",
+        null=True,
+        on_delete=models.CASCADE,
+        related_name="original_dataset",
+        help_text="Reference to the original dataset if regenerated",
+    )
+    start = models.BooleanField(
+        null=True,
+        help_text="Indicates if the dataset process has started.",
```
avrohomgottlieb, line 37: I'm feeling that if we add `help_text` for one BooleanField, then we should add for the others. It's a conventions question.

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+    expires_at = models.DateTimeField(null=True)
+    is_expired = models.BooleanField(default=False)
+    errored_at = models.DateTimeField(null=True)
+    is_errored = models.BooleanField(default=False)
+    error_message = models.TextField(null=True)
+
+    computed_file = models.OneToOneField(
+        ComputedFile,
+        null=True,
+        on_delete=models.CASCADE,
```
avrohomgottlieb, line 67: This ones more subtle. There is a parent-child relationship here, but in order to represent reality we'd have to also actually delete the file from s3. Until we have that set up, we should resort to SET_NULL. 

```suggestion
        on_delete=models.SET_NULL,
```

avrohomgottlieb (changes requested):
You did a great job here, Nozomi! I have a couple items for feedback, but overall it looks really good. Ping me if you'd like to discuss any of my feedback items.

One other general note is that it would be good to add comments above each group of attributes describing the attribute group (see the OriginalFile model for an example).

PR https://github.com/AlexsLemonade/egg/pull/20 might also be helpful to look at (this PR discusses conventions for Django models ).

Cheers!

api/scpca_portal/migrations/0055_dataset.py
```diff
@@ -0,0 +1,100 @@
...
+import uuid
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("scpca_portal", "0054_tokendownload"),
```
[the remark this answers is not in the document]
nozomione: @davidsmejia I've reapplied the migration as it was just a fix. Thank you!

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+        ordering = ["updated_at", "id"]
+
+    class FileFormats:
+        ANN_DATA = "ANN_DATA"
+        SINGLE_CELL_EXPERIMENT = "SINGLE_CELL_EXPERIMENT"
+
+        CHOICES = (
+            (ANN_DATA, "AnnData"),
+            (SINGLE_CELL_EXPERIMENT, "Single cell experiment"),
+        )
```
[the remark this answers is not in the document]
nozomione: This is great! We will be addressing this in the next sprint 👍

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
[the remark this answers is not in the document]
davidsmejia: So for this and for computed files in the future we will be using UUIDs so they can't be randomly guessed.

api/scpca_portal/models/dataset.py
davidsmejia: This should be foreignKey as well because multiple datasets can be started with any given token.

api/scpca_portal/models/dataset.py
davidsmejia: I think this is the correct relationship as a token can be used to download multiple datasets and every download will be tracked via the token.

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+        help_text="Tokens used to create download.",
+    )
+
+    started_at = models.DateTimeField(null=True)
+    is_started = models.BooleanField(default=False)
+    is_processing = models.BooleanField(default=False)
+    processed_at = models.DateTimeField(null=True)
+    is_processed = models.BooleanField(default=False)
+    expires_at = models.DateTimeField(null=True)
+    is_expired = models.BooleanField(default=False)
```
davidsmejia, line 59: Can we add a comment there saying that this will be evaluated / set in a cron?

api/scpca_portal/models/dataset.py
nozomione: That’s a great point and makes sense! David and I chatted, and we’re thinking of going over the codebase later to add `help_text` where needed. But for now, we’ll hold off since it’s not part of the current pattern - TBD for later.

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+        CHOICES = (
+            (ANN_DATA, "AnnData"),
+            (SINGLE_CELL_EXPERIMENT, "Single cell experiment"),
+        )
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    data = models.JSONField(default=dict)
+    format = models.TextField(choices=FileFormats.CHOICES, null=True)
+    email = models.EmailField(max_length=254, null=True)
+    regenerated_from = models.ForeignKey(
```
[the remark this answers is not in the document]
nozomione: We'll keep this as a `ForeignKey`. A processed and expired dataset can be the source for multiple new datasets, and `regenerated_from` will simply point to the dataset it was generated from. I also applied your feedback on `on_delete` and updated `related_name` to regenerated_datasets.

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+
+        CHOICES = (
+            (ANN_DATA, "AnnData"),
+            (SINGLE_CELL_EXPERIMENT, "Single cell experiment"),
+        )
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    data = models.JSONField(default=dict)
+    format = models.TextField(choices=FileFormats.CHOICES, null=True)
+    email = models.EmailField(max_length=254, null=True)
```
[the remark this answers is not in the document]
nozomione: Great catch on `max_length`, I’ve removed it! As for `email`, it can be `null` initially before the dataset starts processing, so we’ll keep it as is.

api/scpca_portal/models/dataset.py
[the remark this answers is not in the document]
nozomione: Based on the above, we'll keep it as a randomly generated UUID value.

api/scpca_portal/models/dataset.py
```diff
@@ -0,0 +1,73 @@
...
+    regenerated_from = models.ForeignKey(
+        "self",
+        null=True,
+        on_delete=models.CASCADE,
```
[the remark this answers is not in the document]
nozomione: Great insight, thank you!

api/scpca_portal/models/dataset.py
nozomione: Great insights, thank you!

nozomione:
I've applied both of your feedback, and this PR is ready for another look. Thank you again for the valuable insights!

api/scpca_portal/models/dataset.py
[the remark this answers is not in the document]
nozomione: About `format`, I ran into this error during the migration,`You are trying to add a non-nullable field 'test' to dataset without a default; we can't do that`. So to work around it, I made the field nullable to avoid the database constraint issue.

Should we set the default to a placeholder string (e.g., `default='None'`) for the field? Let me know what you think!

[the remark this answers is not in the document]
nozomione: Another thought: would it make sense to define a default placeholder in `common.py` for consistency, so we can standardize the default placeholder value for cases like this across the codebase?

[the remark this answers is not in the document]
davidsmejia: That is because you already have created the table. You need to roll back your migration and re-run makemigrations so that it is a single file. You can have non-nullable columns on new tables. If you already have a table the ORM doesnt know if there will be rows which would become invalid by the migration.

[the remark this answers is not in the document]
nozomione: I've reapplied the migration again and now the `format` field is no longer nullable. Thank you all!

avrohomgottlieb (approved):
Looks great Nozomi! 🚀

Issue #1049, opened 2025-01-13
Test module for sync_original_files


avrohomgottlieb:
### Context

The `sync_original_files` command was recently added to the code base, allowing us to update the OriginalFile table on demand.

### Problem or idea

Tests should be created which test the functionality of the command. We plan on mocking the s3 response.

Issue #1047, opened 2025-01-13
Replace data_file_paths with original_files


avrohomgottlieb:
### Context

With the original file model up and running, we no longer have a need for the data_file_paths attribute on the `Project` and `Library` models.

### Problem or idea

All references to `data_file_paths` should return their corresponding original files.

### Solution or next step

There should be a new property method added to the `Project` and `Library` models which called `original_files`, which queries all of a Project or Library's corresponding original files.

PR #1054, opened 2025-01-17
1049 - Add test_sync_original_files module

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1049.

Add testing for sync_original_files command which tests
- creating files
- updating files
- deleting a single file
- attempting to delete all files without passing the `--allow-bucket-wipe` flag
- deleting all files while passing the `--allow-bucket-wipe` flag

## Types of changes

## Functional tests

- `test_sync_original_files`

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/test/management/commands/test_sync_original_files.py
+++ b/api/scpca_portal/test/management/commands/test_sync_original_files.py
@@ -120,7 +120,7 @@ class TestSyncOriginalFiles(TestCase):
             sorted(modified_files.values_list("s3_key", flat=True)),
         )
 
-        # TEST ATTEMPTED DELETION OF ALL FILES - ALLOW_BUCKET_WIPE FLAG `NOT` PASSED
+        # TEST ATTEMPTED DELETION OF ALL FILES - allow_bucket_wipe flag NOT passed
         with patch("scpca_portal.s3.list_bucket_objects", return_value=self.empty_objects_list):
             self.sync_original_files()
         self.assertListEqual(
@@ -128,7 +128,7 @@ class TestSyncOriginalFiles(TestCase):
             sorted(OriginalFile.objects.all().values_list("s3_key", flat=True)),
         )
 
-        # TEST SUCCESSFUL DELETION OF ALL FILES - ALLOW_BUCKET_WIPE FLAG PASSED
+        # TEST SUCCESSFUL DELETION OF ALL FILES - allow_bucket_wipe flag passed
         with patch("scpca_portal.s3.list_bucket_objects", return_value=self.empty_objects_list):
             self.sync_original_files(allow_bucket_wipe=True)
         self.assertFalse(OriginalFile.objects.exists())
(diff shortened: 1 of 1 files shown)


avrohomgottlieb:
One interesting note about the `aws s3api list-objects` command that came up during testing is that it does not return data-less files (files with a size of 0B). These files are displayed as existing when using the aws console and when queried in the `aws s3 ls` command, but not in `aws s3api list-objects`. This will have to be addressed when we refactor `test_loader` in Issue https://github.com/AlexsLemonade/scpca-portal/issues/1047, as all data files in the public test bucket are all data-less, and the syncing of the `OriginalFile` table will yield incorrect results.

nozomione (approved):
LGTM 👍!

api/scpca_portal/test/management/commands/test_sync_original_files.py
```diff
@@ -0,0 +1,134 @@
...
+                "s3_key": "SCPCP999990/SCPCS999990/SCPCL999990_filtered_rna.h5ad",
+                "size_in_bytes": 801,
+                "hash": "modified_hash",
+            },
+        ]
+
+        self.empty_objects_list = []
+
+    def test_sync_original_files(self):
+        self.assertFalse(OriginalFile.objects.exists())
```
davidsmejia, line 87:
```suggestion
```

api/scpca_portal/test/management/commands/test_sync_original_files.py
```diff
@@ -0,0 +1,134 @@
...
+
+        non_modified_files = OriginalFile.objects.filter(hash_change_at=first_sync_timestamp)
+        self.assertListEqual(
+            sorted(
+                file.get("s3_key")
+                for file in self.modified_objects_list
+                if file.get("hash") != "modified_hash"
+            ),
+            sorted(non_modified_files.values_list("s3_key", flat=True)),
+        )
```
davidsmejia, line 111:
```suggestion
        non_modified_s3_keys =           OriginalFile.objects.filter(hash_change_at=first_sync_timestamp).values_list("s3_key", flat=True)
        self.assertQuerysetEqual(
  non_modified_s3_keys,
            [
                'SCPCP999990/SCPCS999990/SCPCL999990_filtered.rds',
                'SCPCP999990/SCPCS999990/SCPCL999990_filtered_rna.h5ad'
            ]
        )
```

1. can we update to order the assertion to be actual, then expected
2. can we hard code the expected here to be explicit

davidsmejia (approved):
LGTM just a couple tweaks for consistency

avrohomgottlieb:
> One interesting note about the `aws s3api list-objects` command that came up during testing is that it does not return data-less files (files with a size of 0B). These files are displayed as existing when using the aws console and when queried in the `aws s3 ls` command, but not in `aws s3api list-objects`. This will have to be addressed when we refactor `test_loader` in Issue #1047, as all data files in the public test bucket are all data-less, and the syncing of the `OriginalFile` table will yield incorrect results.

It turns out that I was mistaken. `aws s3api list-objects` does return 0B files. The reason that 0 byte files weren't showing up was because they were being filtered out in `s3::remove_listed_directories`. The check as to whether or not a file was a dir was whether or not it's size was 0B. I've since updated that logic to check for a trailing "/", and we're back in business.

PR #1058, opened 2025-01-23
Temporary Patch for Issue Templating

avrohomgottlieb:
## Issue Number

N/A

It looks like Github dropped support for the old style of Issue templating and now requires a header section in order for templates to be picked up.

We should discuss whether or not we want to keep the one general issue or if we'd prefer to break it up into different templates like other repos do.

## Types of changes

## Functional tests

N/A

## Checklist

N/A

## Screenshots

N/A


davidsmejia (approved):
LGTM

Issue #1048, opened 2025-01-13
Replace querying and download with OriginalFile methods


avrohomgottlieb:
### Context

At the conclusion of https://github.com/AlexsLemonade/scpca-portal/issues/1047, an `original_files` property method was added to the `Project` and `Library` files, replacing the `data_file_paths` attributes that both models had.

### Problem or idea

This ticket takes it a step further by refactoring and optimizing the methods that query for original files that calls Project::original_files and Library::original_files under the hood.

PR #1060, opened 2025-01-24
1047 - Replace data file paths attribute with property method

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1047.

The main purpose of this PR is to remove the `data_file_paths` attributes on the `Project` and `Library` models, and to replace them with `data_file_paths` property methods which query the `OriginalFile` table under the hood. This PR is the first in a two part series which aims to integrate the `OriginalFile` model into the code base. The next PR, described in Issue https://github.com/AlexsLemonade/scpca-portal/issues/1048, will introduce a more comprehensive refactor, restructuring downstream query logic and removing no longer used code.

## Types of changes

## Functional tests

- Updated tests in `test_loader` and `test_create_portal_metadata`

## Checklist

## Screenshots

N/A


api/scpca_portal/models/project.py
```diff
@@ -99,9 +97,15 @@ def valid_download_configs(self) -> List[Dict]:
...
         ]
 
+    @property
+    def data_file_paths(self):
+        return sorted(self.original_files.values_list("s3_key", flat=True))
+
     @property
     def original_files(self):
-        return OriginalFile.objects.filter(project_id=self.scpca_id, library_id=None)
+        return OriginalFile.objects.filter(
```
davidsmejia, line 106: This can be cached in the table at time of creation

api/scpca_portal/models/original_file.py
```diff
@@ -59,16 +60,17 @@ def get_from_dict(cls, file_object, bucket, sync_timestamp):
...
-            is_merged=is_merged,
-            is_metadata=is_metadata,
+            project_id=relationship_ids["project_id"],
+            sample_id=relationship_ids["sample_id"],
+            library_id=relationship_ids["library_id"],
+            is_single_cell=modalities["is_single_cell"],
+            is_spatial=modalities["is_spatial"],
+            is_single_cell_experiment=formats["is_single_cell_experiment"],
+            is_anndata=formats["is_anndata"],
+            is_cite_seq=formats["is_cite_seq"],
```
davidsmejia, line 70:
```suggestion
            is_cite_seq=formats["is_cite_seq"], # indicates if the file is exclusively cite_seq
```

api/scpca_portal/utils.py
```diff
@@ -114,11 +116,21 @@ def transform_keys(data_dict: Dict, key_transforms: List[Tuple]):
...
+    they are passed here as *args.
     """
-    for key, transformation_function in value_transforms.items():
-        data_dict[key] = transformation_function(data_dict[key])
+    for key, func in value_transforms.items():
+        signature = inspect.signature(func)
+        if len(signature.parameters) == 1:
+            data_dict[key] = func(data_dict[key])
+        else:
+            data_dict[key] = func(data_dict[key], *args)
```
davidsmejia, line 130: I would trim this to match instead of handling 2 cases (1 arg or 1+ args). Add to the doc string that the transforms are called with the same args, if you need substantially different args then you should call the method again with a different `value_transforms` dict

api/scpca_portal/models/original_file.py
```diff
@@ -140,69 +154,85 @@ def purge_deleted_files(
...
+        if self.is_project_file:
+            return False
+        # spatial and single_cell are mutually exclusive
+        return not self.is_spatial
+
+    @property
+    def is_cite_seq(self):
+        if self.is_project_file:
+            return False
+        return common.CITE_SEQ_FILE_SUFFIX in self.s3_key.name
```
davidsmejia, line 205:
```suggestion
        return self.s3_key.name.endswith(common.CITE_SEQ_FILENAME_ENDING)
        
        # where CITE_SEQ_FILENAME_ENDING = "_adt.h5ad"
```
Since we know that Cite-Seq files are only for anndata files we can assert this instead. Would also make sense to reference the FORMAT_EXTENSIONS["ANN_DATA"] as this will change if we update the AnnData extension. (we have before)

davidsmejia (commented):
Some suggestions for helping reduce interdependence between properties. 

After addressing this feedback we should be good to merge this in.

I do think that we should discuss the possibility of making this dataclass the return type of `s3.list_bucket_objects`.  This would be a much larger change that would require reviewing other conventions etc. But for now, this really helps clarify how these properties are logically evaluated.

api/scpca_portal/models/original_file.py
```diff
@@ -140,69 +154,79 @@ def purge_deleted_files(
...
+    @property
+    def library_id(self):
+        if self.library_id_part:
+            return self.library_id_part.split("_")[0]
+        return self.library_id_part
+
+    @property
+    def is_project_file(self):
+        """Project files have project dirs but don't have sample dirs"""
+        return not self.sample_id
```
davidsmejia, line 184:
```suggestion
        return self.project_id and not self.sample_id
```

Your comment makes me think we should add this.

api/scpca_portal/models/original_file.py
```diff
@@ -140,69 +154,79 @@ def purge_deleted_files(
...
+            return False
+
+        # all spatial files have "spatial" appended to the libary part of their file path
+        return self.library_id_part.endswith("spatial")
+
+    @property
+    def is_single_cell(self):
+        if not self.library_id_part:
+            return False
+
```
davidsmejia, line 198:
```suggestion
```

We could remove this as the library part can be either None, a folder, or a filename.

davidsmejia (approved):
LGTM just a couple small tweaks.

api/scpca_portal/models/original_file.py
avrohomgottlieb: Like this idea. Had to cast the expression to a bool because `None and <something>` evaluates to None for some reason, which violates the `BooleanField` constraints.

Issue #1069, opened 2025-01-28
Allow batch jobs read access to secrets manager secrets


davidsmejia:
### Context

After moving sensitive data from envars to the secrets manager for batch jobs, we need to allow them to read those secrets.

### Problem or idea

Create a policy that allows batch jobs to read from the scpca portals secrets manager.

### Solution or next step

- Add and attach policy to terraform infrastructure

davidsmejia:
Closed with #1084

PR #1084, opened 2025-02-03
Attach policy to ecs task role off of batch job role

davidsmejia:
## Issue Number

#1069 

As per the error, the permissions that were missing were on the ecs task role and not the batch job role. This PR moves that attachment off of the batch job role and onto the ecs task role.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


avrohomgottlieb (approved):
👍

Issue #1045, opened 2025-01-10
SES Infrastructure - API and Batch Permissions


avrohomgottlieb:
### Context

Issue https://github.com/AlexsLemonade/scpca-portal/issues/1037 created the domain verification for SES. Once configured on aws, we'll be able to set straight the DNS records and send test emails from the console.

### Solution or next step

Now we'd like to add IAM policies to allow for both the API and Batch instances to be able send ses emails.

Definition of Done:

- Test emails can successfully be sent from the API
- Test emails can successfully be sent from Batch

avrohomgottlieb:
A corresponding PR was merged into dev so this Issue was automatically closed. However, it should be open until we can verify that emails are being sent on Batch and the API by successfully running the send_test_email and dispatch_send_test_email commands.

avrohomgottlieb:
Emails were successfully sent on the API and Batch. This issue can now be closed.

PR #1097, opened 2025-02-07
Attach ses send policy to batch job role

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/1045

Add `aws_iam_role_policy_attachment` to attach `batch_ses_send_email` policy to `batch_job_role`.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/infrastructure/batch/roles.tf
+++ b/infrastructure/batch/roles.tf
@@ -97,7 +97,6 @@ EOF
   tags = var.batch_tags
 }
 
-
 resource "aws_iam_policy" "batch_job_s3_access" {
   name = "scpca-portal-batch-job-s3-access-${var.user}-${var.stage}"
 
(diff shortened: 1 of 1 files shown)


infrastructure/batch/roles.tf
```diff
@@ -63,7 +63,7 @@ resource "aws_iam_policy" "batch_ses_send_email" {
 EOF
 }
 
-resource "aws_iam_role_policy_attachment" "batch_ses_send_policy" {
+resource "aws_iam_role_policy_attachment" "ecs_task_ses_send_policy" {
```
davidsmejia, line 66: I don't think we need to attach the policy to the `ecs_task_role` at all. I would just move the policy down after the `batch_job_role` and keep the attachment there.

infrastructure/batch/roles.tf
```diff
@@ -121,6 +97,7 @@ EOF
   tags = var.batch_tags
 }
 
+
```
davidsmejia, line 100:
```suggestion
```

davidsmejia (approved):
LGTM 
one extra empty line looks to have sneaked in

Issue #1091, opened 2025-02-06
Separate concerns with S3keyinfo helper class and OriginalFile model


avrohomgottlieb:
### Context

In PR https://github.com/AlexsLemonade/scpca-portal/pull/1085 the `S3KeyInfo` dataclass was introduced to handle the parsing of the `s3_key`, from which many OriginalFile attributes are derived from. This issue will take the above mentioned PR a step further by further developing the S3KeyInfo dataclass and cleaning up OriginalFile querying.

### Problem or idea
- S3KeyInfo should be concerned only with the s3key, disconnected from the OriginalFile model
- The query in get_original_files_by_download_config should be easier to understand and should better leverage the OriginalFile model

### Solution or next step

- Make `utils/` a folder, move current `utils.py` folder into it and rename file `helpers.py`.
- Convert S3KeyInfo from dataclass into a normal class and put it inside of `utils/` in a file named `s3keyinfo.py`.
- Add `__init__.py` which imports `helpers.py` and `s3keyinfo.py` (we want to be able to able to access the methods and classes directly as `utils.<class/function name>`)
- Remove current property methods from S3KeyInfo
- Clean up S3KeyInfo attributes to read as follows:
  - s3_key: Path
  - project_id_part: str | None
  - sample_id_part: str | None
  - library_id_part: str | None
  - has_merged_part: bool
  - has_bulk_part: bool 
- Add test module which tests that s3keyinfo properties are returning correct values
- Add attributes to the OriginalFile model
  - `is_library_file = models.BooleanField(default=False)`
    - assigned as `bool(s3keyinfo.library_part)`
  - `is_supplementary_file = models.BooleanField(default=False)`
    - assigned as `self.s3_key.suffix in common.SUPPLEMENTARY_FILE_EXTENSIONS` (similar to how we assign is_metadata)
- In `OriginalFile::get_from_dict`
  - Instance of S3KeyInfo should be instantiated, passing it the s3 key  
  - Outside of the constructor, is_project_file and is_library_file vars should be declared and assigned
  - All OF attributes should be set inside the constructor, leveraging the instance of S3KeyInfo
- Query inside of Project|Library::get_original_files_by_download_config should be made a compound query, which is the union of the passed download_config format and project|library supplementary files

avrohomgottlieb:
After s3keyinfo is moved to a helper class inside of `utils`, pause to evaluate.

PR #1105, opened 2025-02-11
1091 - Move s3keyinfo to helper class

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1091

This PR refactors the `S3KeyInfo` class, converting it from a dataclass inside of the `OriginalFile` model to a helper class in the new `utils/` folder.  The `S3KeyInfo` class was also renamed to `InputBucketS3KeyInfo` to better clarify the class' scope. New `is_supplementary` and `format` attributes/properties were added to `OriginalFile` and `InputBucketS3KeyInfo`, while a `modalities` property method was added to `InputBucketS3KeyInfo` to more elegantly packaged like attributes. A test module `InputBucketS3KeyInfo` was also added.

This PR is viewed as an intermediate step in order to clarify the defining and assignment of `OriginalFile` attributes, as well as preparing for `OriginalFile` querying inside of the new `Dataset` model.

Additionally in this PR, the file `utils.py` was renamed `helpers.py` and placed inside the new `utils/` directory.

## Types of changes

## Functional tests

- `InputBucketS3KeyInfo` test module, which includes 4 tests:
  - `test_id_parsing`
  - `test_is_merged`
  - `test_modalities`
  - `test_format`

## Checklist

## Screenshots

N/A


avrohomgottlieb:
To me it makes sense that every OriginalFile should have a populated `format` attribute. If files like `projects_metadata` and `sample_metadata`, which are non downloadable files, get the format attribute of `FileFormats.Metadata`, then I think that the bulk `.tsv` files should either get `FileFormats.Metadata` or `FileFormats.Bulk` (though bulk is not a technically file format).

api/scpca_portal/migrations/0062_originalfile_format.py
```diff
@@ -0,0 +1,47 @@
...
+        elif original_file.is_anndata:
+            original_file.format = FileFormats.ANN_DATA
+        elif original_file.is_supplementary:
+            original_file.format = FileFormats.SUPPLEMENTARY
+        elif original_file.is_metadata:
+            original_file.format = FileFormats.METADATA
+        else:
+            original_file.format = None
+
+        original_file.save()
```
davidsmejia, line 23: Instead of saving these individually, can we instead make the changes, then bulk save?

api/scpca_portal/models/original_file.py
```diff
@@ -73,20 +77,37 @@ def get_from_dict(cls, file_object, bucket, sync_timestamp):
...
             is_merged=s3_key_info.is_merged,
             is_project_file=s3_key_info.is_project_file,
-            is_downloadable=s3_key_info.is_downloadable,
+            is_downloadable=OriginalFile._is_downloadable(modalities, format),
         )
 
         return original_file
 
+    @staticmethod
+    def _is_downloadable(modalities, format):
```
davidsmejia, line 97: Does it make more sense to just pass s3_key_info? I think we should type hint this method as well.

api/scpca_portal/test/utils/test_input_bucket_s3_key_info.py
davidsmejia, line 1: Do we want .txt files to show up as single-cell experiment files?
I think it would be helpful if we listed these files together to help demonstrate the expected bucket structure. Let me know if you think there is a better alternative to collecting these in `setUp`.

api/scpca_portal/test/utils/test_input_bucket_s3_key_info.py
```diff
@@ -0,0 +1,89 @@
...
+        single_cell_file = Path("SCPCP999999/SCPCS999999/SCPCL999999_file.txt")
+        spatial_file = Path("SCPCP999999/SCPCS999999/SCPCL999999_spatial/SCPCL999999_file.txt")
+        cite_seq_file = Path("SCPCP999999/SCPCS999999/SCPCL999999_file_adt.h5ad")
+        bulk_file = Path("SCPCP999999/bulk/SCPCP999999_bulk.txt")
+
+        single_cell_file_s3_key_info = InputBucketS3KeyInfo(single_cell_file)
+        self.assertIn(Modalities.SINGLE_CELL, single_cell_file_s3_key_info.modalities)
+        self.assertNotIn(Modalities.SPATIAL, single_cell_file_s3_key_info.modalities)
+        self.assertNotIn(Modalities.CITE_SEQ, single_cell_file_s3_key_info.modalities)
+        self.assertNotIn(Modalities.BULK_RNA_SEQ, single_cell_file_s3_key_info.modalities)
```
davidsmejia, line 44:
```suggestion
        self.assertIn(Modalities.SINGLE_CELL, single_cell_file_s3_key_info.modalities)
        self.assertEqual(1, len(single_cell_file_s3_key_info.modalities))
```

I think this is a bit more concise and can be applied to the others.

davidsmejia:
> To me it makes sense that every OriginalFile should have a populated `format` attribute. If files like `projects_metadata.csv` and `sample_metadata.csv`, which are non-downloadable files, get the format attribute of `FileFormats.Metadata`, then I think that the bulk `.tsv` files should get one, either `FileFormats.Metadata` or `FileFormats.Bulk` (though bulk is not a technically file format).
> 
> If all files have a format, then the attribute would be nullable. If not, then it would be nullable.

First, we should probably limit the blast radius of this PR by not implementing here what I am going to describe next.

I think we could improve on how we are defining `FileFormats` in general. Initially we were really only thinking about `ANN_DATA` and `SINGLE_CELL_EXPERIMENT`. Now, with `OriginalFile` this has been extended to include other categories of files. Also, we should consider this In combination with what we spoke about earlier today re: splitting `DatasetFormat` and `FileFormat` as well as what has been hashed out for the future `Dataset` and pulling in `OriginalFiles`.

I would propose the following:

```python
class DatasetFormat:
    ANN_DATA
    SINGLE_CELL_EXPERIMENT
    SPATIAL_SPACERANGER # still thinking this one through

class FileType:
    ANN_DATA # .h5ad
    SINGLE_CELL_EXPERIMENT # .rds
    SPATIAL_SPACERANGER # in spatial folder?
    METADATA # json / csv
    SUPPLEMENTARY # html
    INGRESS # projects.csv / samples_metadata.csv
```

This would allow us to distinguish everything when it comes time to do any of our current procedures. ie pulling the input bucket, fetching files to go in a download, loading metadata.

We can slot this into pre-planning discussions and see if we have enough information to make a decision now or if we should wait for SPATIAL changes have been better defined. I still think we can move ahead without it and then refine our usage after splitting up these enums.

davidsmejia (commented):
I think this looks good, We should still collect the file paths in `test_input_bucket_s3_key_info` into setup and reuse the same paths. In the future we can extract them to an expected values class so that we can ensure better coverage by testing each type of file.

davidsmejia (approved):
LGTM

PR #1107, opened 2025-02-12
1099 - Add the Job model

### LINKED ISSUE #1099: Add the Job model
### Context

As part of the preparation for adding ad hoc downloads to the portal, we want to introduce the `Job` model to represent batch jobs in our database. 

This model will allow us to track the lifecycle of AWS Batch jobs, including timestamps for job state changes, the number of job runs on the same dataset, and the details of the processing pipeline and compute environment.

We should support the following behaviors:
- Creating a model instance before submission:
    - Ensure the model instance data is valid
    - The instance `job.id` must exist as it'll be passed as an argument when dispatching a job to AWS Batch
    - Enable sending job requests in bulk
    - Handle jobs that were unable to submit due to network errors:
        - Recoverable (e.g., AWS service issue) 
        - Unrecoverable(e.g., Invalid data)
- Submitting the instance to AWS Batch from the model instance method (via `job.submit()`)
- Terminating a running job in AWS Batch from the model instance method (via `job.terminate()`) 
- Retrying a job by creating a new model instance and submitting it to AWS Batch from the model instance method (via `job.get_retry_job()`)
- Tracking the current state and recording timestamps on job status changes
- Storing the pipeline, job definition, and commands sent to AWS Batch via the request upon submission
- Storing the batch job ID and status received from AWS Batch via the response upon submission
- Establishing a many-to-one relationship with the `Dataset` model

### Problem or idea

The initial model should contain the following fields:

#### Internal Attributes 
|Field|Type|Usage|
|:--|:--|:--|
|`id`|Integer|Auto-generated primary key which is used to uniquely identify each job request.|
|`attempt`|Integer|A manually incremented number representing the job attempt, if the previous job failed. 1 by default.|
|`state`|Enum|The current state of the job (`CREATED`, `SUBMITTED`, `COMPLETED`, `TERMINATED`). `CREATED` by default, `SUBMITTEED` after a job dispatched to AWS Batch, and `COMPLETED` when the job finishes. This will be updated during the job execution, and the corresponding timestamps will be updated on state changes. <br/><br/>This is for an internal use and not the same as AWS Batch job "status".|
|`created_at`|Datatime|Inherited from `TimestampModel`. Timestamp when the model instance was `state.CREATED`.|
|`submitted_at`|DateTime|Timestamp when the model instance was `state.SUBMITTED` to AWS Batch.|
|`completed_at`|DateTime|Timestamp when the job processing was `state.COMPLETED` (either success or failure).|
|`terminated_at`|DateTime|Timestamp when the job was `state.TERMINATED`.|
|`failure_reason`|Text|Details of the job failure.|
|`critical_error`|Boolean|A flag indicates whether a job should be recoverable. It is set when terminating a job. If `True`, the job is irrecoverable, otherwise recoverable. `False` by default.|
|`retry_on_termination`|Boolean|A flag indicates whether a terminated job should be retried. It is set when retrying a job. If `True`, the job will be retried, otherwise will not. `False` by default.<br/><br/>We'll only retry terminated jobs only when `critical_error` is `False` and this flag is set to `True`.|
|`updated_at`|Datatime|Inherited from `TimestampModel`. Timestamp when the model instance was last updated.|

#### Job Information Sent to AWS (via Request)

|Field|Type|Usage|
|:--|:--|:--|
|`batch_job_name`|Text|Name of the job for processing.|
|`batch_job_queue`|Text|Name of pipeline used for processing the job requests.|
|`batch_job_definition`|Text|Name of the specified job definition template.|
|`batch_container_overrides`|JSON|Arguments to override default container settings (i.e., Docker image or the job definition).|

#### Job Information Defined from AWS (via Response)

|Field|Type|Usage|
|:--|:--|:--|
|`batch_job_id`|Text|The AWS job UUID generated on job submission|
|`batch_status`|Text|The current state of the job (i.e., one of the following: `SUBMITTED`, `PENDING`, `RUNNABLE`, `STARTING`, `RUNNING`, `SUCCEEDED`, `FAILED`).|

#### Relationship

|Field|Type|Usage|
|:--|:--|:--|
|`dataset`|ForeignKey|Many-to-one relationship to the `Dataset` model.|

It should also include the following instance methods:

- `__str__(self)`: String representation of each model instance.

- `submit(self)`: Calls when submitting a job via `boto3`. Updates `batch_job_id` and `state`.

- `terminate(self, retry_on_termination: Bool = False)`: Calls to terminate a currently processing job via `boto3`. Updates `state`, `critical_error` if an error, and if true, the job is irrecoverable.

- `get_retry_job(self, ...)`: Calls to retry the job with the new model instance via `boto3`. Updates `attempt` and `state`. Called only when no `critical_error` and `retry_on_termination` is true.

**NOTE:** The names of the fields and methods are subject to change, and the actual implementations of those methods will be included in later issues.

### Solution or next step

Based on the above requirements:
- Implement the `Job` model 
- Generate the migration to apply the updates

nozomione:
## Issue Number

Close #1099 

Changes include:
- Implemented the initial `Job` model
- Added `JobState` in `enums` for the `state` field (temporarily a regular class)
- Applied the migration to update the database schema

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


api/scpca_portal/enums/job_state.py
```diff
@@ -0,0 +1,12 @@
...
+    SUBMITTED = "SUBMITTED"
+    COMPLETED = "COMPLETED"
+    TERMINATED = "TERMINATED"
+
+    CHOICES = (
+        (CREATED, "Created"),
+        (SUBMITTED, "Submitted"),
+        (COMPLETED, "Completed"),
+        (COMPLETED, "Terminated"),
+    )
```
nozomione, line 12: This is currently a regular class, but will be converted to a subclass of `Enum` in a later issue.

api/scpca_portal/models/job.py
```diff
@@ -0,0 +1,57 @@
...
+    batch_container_overrides = models.JSONField(default=dict)
+
+    # Job Information Defined from AWS (via Response)
+    batch_job_id = models.TextField(null=True)
+    batch_status = models.TextField(null=True)
+
+    dataaset = models.ForeignKey(Dataset, null=True, on_delete=models.SET_NULL, related_name="jobs")
+
+    def __str__(self):
+        return f"Job {self.id} running in {self.batch_job_name}"
```
davidsmejia, line 38:
```suggestion
        if self.batch_job_id:
            return f"Job {self.id} - {self.batch_job_id} - {self.state}"
            
        return f"Job {self.id} - {self.state}"
        
```

api/scpca_portal/models/job.py
```diff
@@ -0,0 +1,57 @@
...
+    batch_job_name = models.TextField(null=True)
+    batch_job_definition = models.TextField(null=True)
+    batch_job_queue = models.TextField(null=True)
+    batch_container_overrides = models.JSONField(default=dict)
+
+    # Job Information Defined from AWS (via Response)
+    batch_job_id = models.TextField(null=True)
+    batch_status = models.TextField(null=True)
+
+    dataaset = models.ForeignKey(Dataset, null=True, on_delete=models.SET_NULL, related_name="jobs")
```
davidsmejia, line 35:
```suggestion
    # datasets should never be deleted
    dataset = models.ForeignKey(Dataset, null=True, on_delete=models.SET_NULL, related_name="jobs")
```

I think that in the future we will want to make this non-nullable. For now, since Dataset is in development it will make it easier to iterate transitioning away from download configs.

davidsmejia (commented):
Looks mostly right to me just a couple comments

nozomione:
I've applied your feedback and this PR is ready for another look. Thank you, David!

davidsmejia (approved):
LGTM 🚀

api/scpca_portal/enums/job_state.py
avrohomgottlieb: With FileFormats and Modalities pluralized, maybe it makes sense to pluralize JobStates for consistency with the other enums.

avrohomgottlieb (approved):
One tiny fix, but this looks great 👍

Issue #1098, opened 2025-02-07
Update Cloud Data Management section in API Readme


avrohomgottlieb:
### Context

Since implementing AWS Batch, we've yet to update the Cloud Data Management section of the API readme. 

### Problem or idea

There should be instructions as to how to generate computed files on Batch (`sync_original_files` + `load_metadata` + `dispatch_to_batch`), in addition to updated instructions on how to generate computed files on the API (`load_data).

### Solution or next step

...

PR #1122, opened 2025-02-24
1098 - Update cloud data management in API readme

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1098

This PR updates the Cloud Data Management section of the API readme, taking recent changes into account. 
Among recent changes includes:
- The addition of the Batch workflow and the `dispatch_to_batch` command
- The more granular approach of loading data on the API via the `load_metadata` and `generate_computed_files` commands

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -187,22 +187,57 @@ sportal load-data --max-workers 10 --reload-all --update-s3 --s3-multipart-chunk
 
 ## Cloud Data Management
 
-The `load_data` and `purge_project` commands can also be run in the cloud.
-The one difference is that in the cloud `load_data` defaults to uploading data.
-This is to help prevent the S3 bucket data from accidentally becoming out of sync with the database.
+### Processing Options
+There are two options available for processing data in the Cloud:
+- Running `load_data` on the API instance (or a combination of `load_metadata` and `generate_computed_files`)
+- Running `dispatch_to_batch` on the API instance, which kicks off processing on AWS Batch resources
 
-To run a command in production, there is a run_command.sh script that is created on the API instance.
-It passes any arguments through to the `manage.py`, so `./run_command.sh load_data --reload-all` will work nicely.
+Due to the fact that processing on Batch is ~10x faster than processing on the API, we recommend using Batch for processing.
 
-The following code can be used to process projects one by one with a minimum disk space footprint:
-```
+### Commands in Production
+To run a command in production, there is a run_command.sh script that is created on the API instance. It passes any arguments through to the manage.py script, making the following acceptable `./run_command.sh load_data --reload-all`.
+
+### Processing on the API
+The following code can be used to process projects on the API, one by one, with a minimum disk space footprint:
+
+```bash
 for i in $(seq -f "%02g" 1 20); do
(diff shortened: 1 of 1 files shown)


api/README.md
```diff
@@ -187,22 +187,57 @@ sportal load-data --max-workers 10 --reload-all --update-s3 --s3-multipart-chunk
...
+The following code is used for processing projects via AWS Batch:
+```bash
+./run_command.sh dispatch_to_batch
+```
+
+`dispatch_to_batch` has two options that can be passed, `regenerate-all` and `project-id`, which change which set of computed files are generated on Batch:
+- With neither `regenerate-all` and `project-id` passed, jobs are submitted for all projects which do not have computed files associated with them.
+- With the `regenerate-all` flag passed (without the `project-id` flag), jobs will be submitted which process computed files for all projects, indifferent to whether or not they already have computed files associated with them.
+- With the `project-id` flag passed (without the `regenerate-all` flag), jobs will be submitted for a specific project if there are no computed files already associated with it.
+- With both the `regenerate-all` and `project-id` flags passed, a single project's computed files will be regenerated, regardless of whether or not it already has computed files associated with it.
```
davidsmejia, line 234:
```suggestion
By default the `dispatch_to_batch` command will look at all projects and filter out all projects that already have at least 1 computed file. AWS Batch jobs will be dispatched and create valid computed files for matching projects that were not filtered out.

You can override this filter by passing the `--regenerate-all` flag. This will dispatch jobs independent of existing computed files. Any existing computed files will be purged before new ones are generated to replace them.

You can limit the scope of this command to only apply to a specific project by passing the `--project-id <SCPCP999999>` flag. This can be used in conjunction with `--regenerate-all` if you want to ignore existing computed files for that project.
```

I appreciate how complete this is, but perhaps this might be a little bit clearer. I took a swipe at rewriting this to explain each of the flags' usage and also added some more context. If you prefer it the other way that would be alright as well.

davidsmejia (commented):
A couple of small comments and one bigger one. I think that we should also include a description of `sync_original_files`. This really should be the first step.

api/README.md
```diff
@@ -187,22 +187,63 @@ sportal load-data --max-workers 10 --reload-all --update-s3 --s3-multipart-chunk
...
-The following code can be used to process projects one by one with a minimum disk space footprint:
+### Commands in Production
+To run a command in production, there is a `run_command.sh` script that is created on the API instance. It passes any arguments through to the `manage.py` script, making the following acceptable `./run_command.sh load_data --reload-all`.
+
+### Syncing the OriginalFile table
+Before processing can be carried out, the `OriginalFile` table must be populated and synced via the `sync_original_files` command. This command builds a local representation of all objects available in the default (or passed) s3 input bucket, and is considered the single source of truth for inputs files throughout the codebase.
+
+Syncing is carried out as follows:
+```bash
+./run_command.sh sync_original_files
```
davidsmejia, line 205: Can we add a description about the args here?

davidsmejia (approved):
LGTM 🚀

PR #1123, opened 2025-02-25
1106 - Update classes in the enums folder to Enum subclasses

### LINKED ISSUE #1106: Update classes in the enums folder to Enum subclasses
### Context

Related issue #1103

In previous issues #1065 and #1066, the `FileFormats` and `Modalities` classes were added to the `enums` folder to improve maintainability and reusability. 

Additionally, we also added the `JobStatus`class for the`Job` model to `/enums`.

### Problem or idea

In this issue, we will implement the approach from the linked related issue by converting all classes in `/enums` into `models.TextChoices` subclasses.

### Solution or next step

1. Convert all classes in`/enums` into subclasses of `models.TextChoices`.
2. Update all the references accordingly

nozomione:
## Issue Number

Close #1106

I've updated all the classes in the `enums` folder to be subclasses of `Enum`.

Regarding the `NAME_MAPPING` constant, it was previously inside `Modalities`, but due to how `TextChoices` works, it was getting included in the `choices` property, which wasn’t ideal when referencing it. To fix this, I've moved it to the `Sample` model as an inner class, since it’s only relevant for that model. This approach keeps things simple, avoids adding extra helpers, and makes future cleanups easier if needed (as discussed in our meeting).

Let me know if you have any other suggestions or thoughts on this, thank you!

```py
In [1]: from scpca_portal.enums import Modalities

In [2]: Modalities.choices
Out[2]: 
[('BULK_RNA_SEQ', 'Bulk Rna Seq'),
 ('CITE_SEQ', 'Cite Seq'),
 ('MULTIPLEXED', 'Multiplexed'),
 ('SINGLE_CELL', 'Single Cell'),
 ('SPATIAL', 'Spatial'),
 ("{'BULK_RNA_SEQ': 'Bulk RNA-seq', 'CITE_SEQ': 'CITE-seq', 'MULTIPLEXED': 'Multiplexed', 'SPATIAL': 'Spatial Data'}",
  'Name Mapping')]
 ```

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/enums/file_formats.py
+++ b/api/scpca_portal/enums/file_formats.py
@@ -1,12 +1,8 @@
-class FileFormats:
+from django.db.models import TextChoices
+
+
+class FileFormats(TextChoices):
     ANN_DATA = "ANN_DATA"
     SINGLE_CELL_EXPERIMENT = "SINGLE_CELL_EXPERIMENT"
     METADATA = "METADATA"
     SUPPLEMENTARY = "SUPPLEMENTARY"
-
-    CHOICES = (
-        (ANN_DATA, "AnnData"),
-        (SINGLE_CELL_EXPERIMENT, "Single cell experiment"),
-        (METADATA, "Metadata"),
-        (SUPPLEMENTARY, "Supplementary"),
-    )
--- a/api/scpca_portal/enums/job_states.py
+++ b/api/scpca_portal/enums/job_states.py
@@ -1,12 +1,8 @@
-class JobStates:
+from django.db.models import TextChoices
+
+
+class JobStates(TextChoices):
     CREATED = "CREATED"
     SUBMITTED = "SUBMITTED"
     COMPLETED = "COMPLETED"
     TERMINATED = "TERMINATED"
-
-    CHOICES = (
-        (CREATED, "Created"),
-        (SUBMITTED, "Submitted"),
-        (COMPLETED, "Completed"),
-        (COMPLETED, "Terminated"),
-    )
--- a/api/scpca_portal/enums/modalities.py
+++ b/api/scpca_portal/enums/modalities.py
@@ -2,8 +2,8 @@ from django.db.models import TextChoices
 
 
 class Modalities(TextChoices):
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

**Comment by davidsmejia** [comment]
> Let me know if you have any other suggestions or thoughts on this, thank you!

I think that is a good point. What do you think about leveraging the "label" on the TextChoices subclass? It would work very similarly except you would access the Identity based on the `name` and then access the `label`.

If we did this, I think it would make sense to do all of the enums this way as a convention. By default name and value would be the same as that is how we want to reference (name), store (value), and then translate to human readable (label).

```python
In [1]: from django.db.models import TextChoices
   ...:
   ...:
   ...: class Modalities(TextChoices):
   ...:     BULK_RNA_SEQ = ("BULK_RNA_SEQ", "Bulk RNA-seq")
   ...:     CITE_SEQ = ("CITE_SEQ", "CITE-seq")
   ...:     MULTIPLEXED = ("MULTIPLEXED", "Multiplexed")
   ...:     SINGLE_CELL = ("SINGLE_CELL", "Single-cell")
   ...:     SPATIAL = ("SPATIAL", "Spatial")
   ...:
   ...:

In [2]: Modalities["BULK_RNA_SEQ"].label
Out[2]: 'Bulk RNA-seq'

In [3]: Modalities["BULK_RNA_SEQ"].name
Out[3]: 'BULK_RNA_SEQ'

In [4]: Modalities["BULK_RNA_SEQ"].value
Out[4]: 'BULK_RNA_SEQ'
```

nozomione:
> What do you think about leveraging the "label" on the TextChoices subclass? It would work very similarly except you would access the Identity based on the `name` and then access the `label`.

That's a great idea to use the `label` property. I've applied your suggestion and this PR is now ready for another look. Thank you, David!

davidsmejia (approved):
LGTM 🚀 

I think that we will eventually want to remove the `label` definitions as this is in practice for client side presentation only and we should handle that translation there. We really don't need to worry about that yet and can address it in the future.

Issue #1115, opened 2025-02-21
Add test_validate_data tests to test_dataset module


avrohomgottlieb:
### Context

In issue https://github.com/AlexsLemonade/scpca-portal/issues/1114, the test_dataset module was created. We'd now like to make sure that valid datasets are created.

### Problem or idea

We should begin with validating the data attribute, which is a JSON field that should follow the following structure:
```python
data: {
	SCPCP000001: {
		merge_single_cell: True,
		includes_bulk: True,
		SINGLE_CELL: ["SCPCS000001", "SCPCS000002"],
		SPATIAL: ["SCPCS000001"]
	}
}
```

### Solution or next step
- add function in the Dataset model called `validate_data` which takes a JSON field
- add tests in test_data_set which test the validate_data method
- iterate over all of the keys and values to ascertain that they as we expect them to be
  - first key should be project_id ("SCPCP" followed by 6 digits)
  - config keys should be a subset of 4 possible options (merge_single_cell, includes_bulk, SINGLE_CELL, SPATIAL)
  - merge_single_cell and includes_bulk values should be boolean
  - SINGLE_CELL and SPATIAL keys should have arrays with sample_ids ("SCPCS" followed by 6 digits)
- create two tests per field, one a valid entry for the field and one with an invalid entry

PR #1129, opened 2025-03-03
1115 - Add test_validate_data tests to test_dataset module

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1115

This PR introduced a `is_data_valid` property method, which leverages the `DataValidator` class to determine if the Dataset data attribute is properly build. The Dataset data attribute must contain a dictionary with keys being project_ids and values being config dictionaries which contain options and samples to draw from when querying for Original Files during Computed File generation.

A few tests were added to validate the project_id keys and config values confirm to the desired structure.

## Types of changes

## Functional tests

- test_dataset::test_is_data_valid_project_id
- test_dataset::test_is_data_valid_config

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/models/dataset.py
+++ b/api/scpca_portal/models/dataset.py
@@ -87,29 +87,22 @@ class DataValidator:
         ]
 
     def validate_project(self, project_id: str) -> bool:
-        is_valid = self._validate_project_id(project_id)
-        is_valid &= (
-            self._validate_merge_single_cell(project_id)
-            if self.data.get(project_id).get("merge_single_cell")
-            else True
-        )
-        is_valid &= (
-            self._validate_includes_bulk(project_id)
-            if self.data.get(project_id).get("includes_bulk")
-            else True
-        )
-        is_valid &= (
-            self._validate_single_cell(project_id)
-            if self.data.get(project_id).get(Modalities.SINGLE_CELL)
-            else True
-        )
-        is_valid &= (
-            self._validate_spatial(project_id)
-            if self.data.get(project_id).get(Modalities.SPATIAL)
-            else True
-        )
+        if not self._validate_project_id(project_id):
+            return False
 
-        return is_valid
+        if not self._validate_merge_single_cell(project_id):
+            return False
+
+        if not self._validate_includes_bulk(project_id):
+            return False
+
+        if not self._validate_single_cell(project_id):
+            return False
+
+        if not self._validate_spatial(project_id):
+            return False
+
+        return True
 
     def _validate_project_id(self, project_id):
         return self._validate_id(project_id, common.PROJECT_ID_PREFIX)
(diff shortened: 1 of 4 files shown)


api/scpca_portal/models/dataset.py
```diff
@@ -59,3 +61,69 @@ class Meta:
...
+        return (
+            bool(self.project_id)
+            and self._validate_project_id()
+            and bool(self.config)
+            and len(self.config.keys()) == 4
+            and self._validate_merge_single_cell()
+            and self._validate_includes_bulk()
+            and self._validate_single_cell()
+            and self._validate_spatial()
+        )
```
davidsmejia, line 94:
```suggestion
        @property
        def is_valid(self) -> bool:
            return all([self.validate_project(key for key in self.data.keys()])
            
        @property
        def valid_projects(): List[str]:
            return [key for key in self.data.keys() if self.validate_project(key) ]
            
        @property
        def invalid_projects(): List[str]
            pass #invert valid 
```

1. Let's change the `DataElement` to `DatasetData` and then create a property in dataset `dataset_data`.
2. Update validation to match above so that we have properties that can return actionable information and not just the boolean of is or isn't valid.
3. Let's internally throw errors inside of the individual checks but capture them inside of `validate_project`and for now we can always use the non-throwing interface.

api/scpca_portal/models/dataset.py
```diff
@@ -59,3 +61,87 @@ class Meta:
...
+            if self.data.get(project_id).get(Modalities.SINGLE_CELL)
+            else True
+        )
+        is_valid &= (
+            self._validate_spatial(project_id)
+            if self.data.get(project_id).get(Modalities.SPATIAL)
+            else True
+        )
+
+        return is_valid
```
davidsmejia, line 112: Instead of evaluating everything before exiting with True/False, let's rewrite this to return false on first failed validation check. Base case return should be `True`.

davidsmejia: Can we also move "merge_single_cell", "includes_bulk" and the modalities into a new Enum called `DatasetDataProjectAttrs`

api/scpca_portal/models/dataset.py
```diff
@@ -59,3 +61,100 @@ class Meta:
...
+        id_number = id.removeprefix(prefix)
+        return len(id_number) == 6 and id_number.isdigit()
+
+    def _validate_merge_single_cell(self, project_id) -> bool:
+        if DatasetDataProjectConfig.MERGE_SINGLE_CELL not in self.data.get(project_id, {}):
+            return True
+
+        return isinstance(
+            self.data.get(project_id).get(DatasetDataProjectConfig.MERGE_SINGLE_CELL, {}), bool
+        )
```
davidsmejia, line 126:
```suggestion
    def _validate_merge_single_cell(self, project_id) -> bool:
        if value := self.data.get(project_id, {}).get(DatasetDataProjectConfig.MERGE_SINGLE_CELL):
            return isinstance(value, bool)

        return True
```

Do you think something like this would be easier to read?

davidsmejia (approved):
LGTM

I made a suggestion for the individual checks, I think the fallback type should be removed either way. Otherwise, this looks good.

api/scpca_portal/models/dataset.py
avrohomgottlieb: Look this

Issue #1133, opened 2025-03-07
Add support for CCDL dataset flag on dataset model


avrohomgottlieb:
### Context

It's important that we be able to discern between user generated and internally generated datasets.

### Problem or idea

Add an attribute should be added to the dataset model, called `is_ccdl`, which is `True` if the dataset was internally generated and `False` if the dataset was user generated.

### Solution or next step

- Add an `is_ccdl` attribute to the Dataset model
- Add `is_ccdl=True` to the Dataset factory method
- Add tests which check that the `is_ccdl` attribute is set correctly

avrohomgottlieb:
Addressed in https://github.com/AlexsLemonade/scpca-portal/pull/1148

PR #1148, opened 2025-03-13
1133 - CCDL Dataset Flag

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1133.

This PR adds `is_ccdl` and `is_metadata` to the dataset model, for usage in converting pre-generated computed files to datasets.

## Types of changes

## Functional tests

- `test_dataset::test_is_ccdl_default_set`

## Checklist

## Screenshots

N/A


avrohomgottlieb:
A more interesting test for the `is_ccdl` attr will be when it's set to `True` while we're building the management command and the dataset getter and validator methods. I plan on removing the test created here for `is_ccdl` and checking how it's set when we build out the tests cases for those methods. 

For now, I think the current test is fine as a place holder, though we don't except a default value in a Django Model field to change its value if its not explicitly set like in this case.

davidsmejia (approved):
We should move `is_metadata` up with other user editable attributes. I think that we can support this functionality via the API even if it doesn't fit nicely in the portal client at the moment.

LGTM otherwise

Issue #1144, opened 2025-03-12
Add Move Samples Modal


avrohomgottlieb:
### Context

As we continue to develop the cart front end, we're continuing to develop components.

### Problem or idea

We need to build the Move Samples modal based on the below linked Figma design.
Figma: [Move Samples Modal](https://www.figma.com/design/AqRgSXil3E9bWihG2k1U3l/ScPCA-master?node-id=1284-15813&p=f&t=Prd5o37DV3t04lpu-0)

### Solution or next step

Implement the Move Samples Modal.

PR #1149, opened 2025-03-13
1144 - Add Move Samples Modal

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1144.

This PR introduces the `DatasetMoveSamplesModal` component, which allows users to move samples to their cart from previously generated datasets.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/client/src/components/DatasetMoveSamplesModal.js
+++ b/client/src/components/DatasetMoveSamplesModal.js
@@ -1,5 +1,5 @@
 import React, { useState } from 'react'
-import { Box, Grid, RadioButton, Text } from 'grommet'
+import { Box, Grid, RadioButtonGroup, Text } from 'grommet'
 import { config } from 'config'
 import { useResponsive } from 'hooks/useResponsive'
 import { Button } from 'components/Button'
@@ -7,6 +7,7 @@ import { Modal, ModalBody } from 'components/Modal'
 import { WarningText } from 'components/WarningText'
 import { InfoText } from 'components/InfoText'
 import { Link } from 'components/Link'
+import { FormField } from 'components/FormField'
 
 export const DatasetMoveSamplesModal = ({
   label = 'Move Samples',
@@ -15,23 +16,17 @@ export const DatasetMoveSamplesModal = ({
   appendDisabled = false
 }) => {
   const [showing, setShowing] = useState(false)
-  const [checkedAppendSamples, setCheckedAppendSamples] = useState(false)
-  const [checkedReplaceSamples, setCheckedReplaceSamples] = useState(false)
+  const radioOptions = [
+    { label: 'Append samples to My Dataset', value: 'append' },
+    { label: 'Replace samples in My Dataset', value: 'replace' }
+  ]
+  const [action, setAction] = useState(radioOptions[0].value)
 
   const { responsive } = useResponsive()
   const totalSamples = 34
 
   const handleClick = () => {
     setShowing(true)
-  }
-
-  const handleChangedCheckAppend = () => {
-    setCheckedAppendSamples(true)
-    setCheckedReplaceSamples(false)
-  }
-  const handleChangedCheckReplace = () => {
-    setCheckedReplaceSamples(true)
-    setCheckedAppendSamples(false)
   }
 
   return (
@@ -58,20 +53,14 @@ export const DatasetMoveSamplesModal = ({
                 iconMargin="0"
               />
             </Box>
-            <Box>
-              <RadioButton
-                name="radio"
-                checked={checkedAppendSamples}
-                label="Append samples to My Dataset"
-                onChange={handleChangedCheckAppend}
+            <FormField>
+              <RadioButtonGroup
+                name="action"
+                options={radioOptions}
+                value={action}
+                onChange={({ target: { value } }) => setAction(value)}
               />
-              <RadioButton
-                name="radio"
-                checked={checkedReplaceSamples}
-                label="Replace samples in My Dataset"
-                onChange={handleChangedCheckReplace}
-              />
-            </Box>
+            </FormField>
             {!appendDisabled && (
               <InfoText>
                 <Text>
--- a/client/src/components/DatasetMoveSamplesModal.js
+++ b/client/src/components/DatasetMoveSamplesModal.js
@@ -14,9 +14,12 @@ export const DatasetMoveSamplesModal = ({
   modalDisabled = false,
   appendDisabled = false
 }) => {
+  const { responsive } = useResponsive()
   const [showing, setShowing] = useState(false)
   const [checkedAppendSamples, setCheckedAppendSamples] = useState(false)
   const [checkedReplaceSamples, setCheckedReplaceSamples] = useState(false)
+
+  const totalSamples = 34
 
   const handleClick = () => {
     setShowing(true)
@@ -51,7 +54,7 @@ export const DatasetMoveSamplesModal = ({
             <Box>
               <WarningText
                 lineBreak={false}
-                text="There are 34 samples in My Dataset"
+                text={`There are ${totalSamples} samples in My Dataset`}
                 iconMargin="0"
               />
             </Box>
@@ -81,6 +84,14 @@ export const DatasetMoveSamplesModal = ({
                 </Text>
               </InfoText>
             )}
+            <Box
+              align="center"
+              direction={responsive('column', 'row')}
+              gap="xlarge"
+              justify="end"
+            >
+              <Button primary aria-label={label} label={label} />
+            </Box>
           </Grid>
         </ModalBody>
       </Modal>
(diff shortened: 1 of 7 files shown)


avrohomgottlieb:
Here's what I'm looking at currently. Anyone have any ideas about how to fix the spacing issues?

image

client/src/components/DatasetMoveSamplesModal.js
```diff
@@ -0,0 +1,100 @@
...
+                label="Append samples to My Dataset"
+                onChange={handleChangedCheckAppend}
+              />
+              <RadioButton
+                name="radio"
+                checked={checkedReplaceSamples}
+                label="Replace samples in My Dataset"
+                onChange={handleChangedCheckReplace}
+              />
+            </Box>
```
nozomione, line 74: Instead of using `Box`, we can use the custom `FormField` component, which is a shared component used specifically for wrapping form elements. 

Grommet also provides `RadioButtonGroup` for managing multiple radio buttons. You can define these radio button options as an array of objects and pass them to the [`options`](https://v2.grommet.io/radiobuttongroup#options) prop of `RadioButtonGroup`.

By using `RadioButtonGroup`, we can also simplify state management by handling `value` changes with a single state variable, rather than using two separate states for each `RadioButton` 👍 

```suggestion
            <FormField>
              <RadioButtonGroup
                name="action"
                options={radioOptions}
                value={action}
                onChange={({ target: { value } }) => setAction(value)}
              />
          </FormField>
```

client/src/components/DatasetMoveSamplesModal.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  const handleChangedCheckAppend = () => {
+    setCheckedAppendSamples(true)
+    setCheckedReplaceSamples(false)
+  }
+  const handleChangedCheckReplace = () => {
+    setCheckedReplaceSamples(true)
+    setCheckedAppendSamples(false)
+  }
```
nozomione, line 35: We can remove these handlers. 

```suggestion
```

client/src/components/DatasetMoveSamplesModal.js
```diff
@@ -0,0 +1,100 @@
...
+
+export const DatasetMoveSamplesModal = ({
+  label = 'Move Samples',
+  title = 'Move Samples to My Dataset',
+  modalDisabled = false,
+  appendDisabled = false
+}) => {
+  const [showing, setShowing] = useState(false)
+  const [checkedAppendSamples, setCheckedAppendSamples] = useState(false)
+  const [checkedReplaceSamples, setCheckedReplaceSamples] = useState(false)
```
nozomione, line 19: With `RadioButtonGroup`, we can define the radio button options and manage the `value` change using a single state.

For example:

```suggestion
 
  const radioOptions = [
    { label: 'Append samples to My Dataset', value: 'append' },
    { label: 'Replace samples in My Dataset', value: 'replace' }
  ]
  const [action, setAction] = useState(radioOptions[0].value) // set default as 'append' here  
```

client/src/components/DatasetMoveSamplesModal.js
```diff
@@ -0,0 +1,100 @@
...
+              <RadioButton
+                name="radio"
+                checked={checkedReplaceSamples}
+                label="Replace samples in My Dataset"
+                onChange={handleChangedCheckReplace}
+              />
+            </Box>
+            {!appendDisabled && (
+              <InfoText>
+                <Text>
```
nozomione, line 77: We can pass `margin` prop to `Text` to add left margin for spacing between the info icon and the copy.

```suggestion
                <Text margin={{ left: 'medium' }}>
```

client/src/components/DatasetMoveSamplesModal.js
```diff
@@ -0,0 +1,100 @@
...
+              />
+              <RadioButton
+                name="radio"
+                checked={checkedReplaceSamples}
+                label="Replace samples in My Dataset"
+                onChange={handleChangedCheckReplace}
+              />
+            </Box>
+            {!appendDisabled && (
+              <InfoText>
```
nozomione, line 76: Currently, the icon size is `16px`(a fixed value set by `InfoText`), but it needs to be `24px`. To fix this, we can modify `InfoText` to add a new prop (e.g., `iconSize`) and pass it to the `Icon` component's `size` prop.

For example: 

```suggestion
              <InfoText iconSize="medium">
```

And we can update `InfoText` as follows:

```jsx
export const InfoText = ({ 
     label, 
     iconSize = '16px',  // add a new prop
     children }) => (
  <Box direction="row" align="center" gap="small">
    <Box pad={{ top: '2px' }}>
      <Icon
        size={iconSize} // pass it here
```

nozomione (changes requested):
Great job, Avrohom 👍 I’ve left my initial feedback. I’ll provide the rest of the feedback (e.g., off-spacing) in my second review later!

client/src/components/DatasetMoveSamplesModal.js
```diff
@@ -0,0 +1,100 @@
...
+            margin={{ bottom: '0' }}
+            pad={{ bottom: '0' }}
+          >
+            <Box>
+              <WarningText
+                lineBreak={false}
+                text={`There are ${totalSamples} samples in My Dataset`}
+                iconMargin="0"
+              />
+            </Box>
```
nozomione, line 60: The `WarningText` component renders a block-level element, so `Box` can be removed here, unless additional styling for layout is required (e.g., if we want to add additional 10px bottom margin, `<Box  margin={{ bottom: '10px' }}>`). 

```suggestion
            <WarningText
                lineBreak={false}
                text={`There are ${totalSamples} samples in My Dataset`}
                iconMargin="0"
            />      
```

client/src/components/DatasetMoveSamplesModal.js
```diff
@@ -0,0 +1,100 @@
...
+            columns={['auto']}
+            margin={{ bottom: '0' }}
+            pad={{ bottom: '0' }}
+          >
+            <Box>
+              <WarningText
+                lineBreak={false}
+                text={`There are ${totalSamples} samples in My Dataset`}
+                iconMargin="0"
+              />
```
nozomione, line 59: Regarding the spacing issue, to resolve this, we can modify the `WarningText` to accept new props while keeping the default values to support the current UIs.

Also, the font size should be `21px` (currently `16px`, a fixed value set by `WarningText`), and the icon size should be `24px` which can be adjusted using the `iconSize` props. 

For example: 
```suggestion
             <WarningText
                lineBreak={false}
                text={`There are ${totalSamples} samples in My Dataset`}
                iconMargin="0"
                iconSize="24px" // set icon size
                textMargin="4px" // a new prop for text spacing
                textSize="21px" // a new prop for text size
              />
```

And we can update the `WarningText` as follows:

```jsx
eexport const WarningText = ({
  lineBreak = true,
  link = '',
  linkLabel = '',
  iconColor = 'status-warning',
  iconMargin = { right: 'medium' },
  iconName = 'Warning',
  iconPad = { right: 'small' },
  iconSize = '16px',
  text = '',
  textMargin = '8px', // a new prop
  textSize = '16px',  // a new prop
  newTab = false,
  children
}) => {
  return (
    <Box
      align="center"
      direction="row"
      margin={{ top: 'small', bottom: 'medium' }}
    >
      <Box margin={iconMargin} pad={iconPad}>
        <Icon color={iconColor} size={iconSize} name={iconName} />
      </Box>
      <Paragraph size={textSize} margin={textMargin}> // pass them here
```

nozomione (changes requested):
I've left my second round of feedback, including my suggestion for fixing the spacing issue. 

Feel free to ping me if you'd like any clarification or want to discuss any of my feedback items 👍

client/src/components/DatasetMoveSamplesModal.js
davidsmejia: Instead of trying to add a ton of props for every bit and bob in this component.  I would instead just leverage the use of children and not pass text, link, or newTab. Honestly, newTab should always be true and removed anyways but dont do it in this PR.

You can literally put whatever you want and style it correctly which is preferred to having many specific props for styling nested components. The same as how `InfoText` is used.

nozomione: This is a great point! As David suggested, instead of editing the `WarningText` component, let's leverage `children` to address the off-spacing issue with the text 👍 

For instance (we use either Grommet `Paragraph` or `Text` component to render text): 

```suggestion
             <WarningText
                lineBreak={false}
                iconMargin="0"
                iconSize="24px" // set the icon size here
              >
              <Text size="24px" margin="ADJUST_AS_NEEDED"> 
                   There are {totalSamples} samples in My Dataset
              </ Text>
```

davidsmejia (approved):
LGTM

nozomione (approved):
LGTM 👍

PR #1151, opened 2025-03-17
1145 - Implement Job::terminate using the existing Job model instance

### LINKED ISSUE #1145: Implement Job::terminate using the existing Job model instance
### Context

The `Job` model has the `terminate` method responsible for terminating (or canceling) a submitted job on AWS Batch.

In a previous PR #1132, we implemented  the `submit` method along with the corresponding test using mock data, as the development of the `Dataset` model is still in progress.

### Problem or idea

In this issue, we'll implement the `terminate` method and its corresponding test using the existing job model instance with mock data.

The method should:
- Terminate a job via `boto3`
  - If failed to terminate, throw an error  
- Updates the following job instance fields:
  -  `retry_on_termination` 
  - `terminated_at`
  - `state` to `TERMINATED`

The test should verify the following flow using mock data:
1. Create a job instance
2. Submit the job to AWS Batch
3. Terminate the job via AWS Batch
4. Make sure the job instance is updated with the appropriate field values

### Solution or next step

- Implement the `terminate` method using `boto3`
- Add the corresponding test using mocked methods and data

nozomione:
## Issue Number

Close #1145

I've implement  the `Job::terminate` method and added the corresponding test using mocks. 

Additionally, I've added the `JobFactory` class and incorporated it in the `TestJob`, as well as cleaned up the no longer used attributes in the `JobTest::setUp` method.

## Types of changes

## Functional tests

- `TestJob::test_terminate_job`
- `TestJob::test_terminate_job_failure`
- `JobFactory`

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/models/job.py
+++ b/api/scpca_portal/models/job.py
@@ -12,6 +12,8 @@ from scpca_portal.enums import JobStates
 from scpca_portal.models import Dataset
 from scpca_portal.models.base import TimestampedModel
 
+logger = get_and_configure_logger(__name__)
+
 
 class Job(TimestampedModel):
     class Meta:
@@ -114,13 +116,6 @@ class Job(TimestampedModel):
         """
         return boto3.client("batch", region_name=settings.AWS_REGION)
 
-    @property
-    def _logger(self):
-        """
-        Logger object for logging message.
-        """
-        return get_and_configure_logger(__name__)
-
     def submit(self) -> None:
         """
         Submit a job via boto3, update batch_job_id and state, and
@@ -148,7 +143,7 @@ class Job(TimestampedModel):
         """
 
         if self.state in [JobStates.COMPLETED, JobStates.TERMINATED]:
-            self._logger.info(
+            logger.info(
                 f"Job with the {self.state} state cannot be terminated.",
                 job_id=self.pk,
                 batch_job_id=self.batch_job_id,
@@ -163,7 +158,7 @@ class Job(TimestampedModel):
 
             self.save()
         except ClientError as error:
-            self._logger.exception(
+            logger.exception(
                 f"Failed to terminate the job due to a ClientException:\n\t{error}",
                 job_id=self.pk,
                 batch_job_id=self.batch_job_id,
@@ -172,7 +167,7 @@ class Job(TimestampedModel):
         except Exception as error:
             self.critical_error = True
             self.state = JobStates.TERMINATED
-            self._logger.exception(
+            logger.exception(
                 f"Failed to terminate the job due to a ServerException:\n\t{error}",
                 job_id=self.pk,
                 batch_job_id=self.batch_job_id,
@@ -181,7 +176,7 @@ class Job(TimestampedModel):
             self.save()
             return False
 
-        self._logger.info(
+        logger.info(
             "Job termination complete.",
             job_id=self.pk,
             batch_job_id=self.batch_job_id,
--- a/api/scpca_portal/models/job.py
+++ b/api/scpca_portal/models/job.py
@@ -143,8 +143,8 @@ class Job(TimestampedModel):
     def terminate(self, retry_on_termination=False):
         """
         Terminate the submitted and incompleted job via boto3, and update state.
-        Throws an error if failed to terminate the job and,
-        set critical_error to True if the job is irrecoverable.
+        Throw an error if failed to terminate the job and,
+        set critical_error to True if the job is irrecoverable (e.g., server error).
         """
 
         if self.state in [JobStates.COMPLETED, JobStates.TERMINATED]:
--- a/api/scpca_portal/models/job.py
+++ b/api/scpca_portal/models/job.py
@@ -134,7 +134,7 @@ class Job(TimestampedModel):
 
         self.save()
 
-    def terminate(self, retry_on_termination=False):
+    def terminate(self, retry_on_termination=False) -> bool:
         """
         Terminate the submitted and incompleted job via boto3, and update state.
         Return True if the job is successfully terminated or already terminated, otherwise False.
--- a/api/scpca_portal/models/job.py
+++ b/api/scpca_portal/models/job.py
@@ -5,7 +5,6 @@ from django.db import models
 from django.utils.timezone import make_aware
 
 import boto3
-from botocore.exceptions import ClientError
 
 from scpca_portal.config.logging import get_and_configure_logger
 from scpca_portal.enums import JobStates
@@ -152,16 +151,10 @@ class Job(TimestampedModel):
             self.terminated_at = make_aware(datetime.now())
 
             self.save()
-        except ClientError as error:
-            logger.exception(
-                f"Failed to terminate the job due to a ClientException:\n\t{error}",
-                job_id=self.pk,
-                batch_job_id=self.batch_job_id,
-            )
-            return False
+
         except Exception as error:
             logger.exception(
-                f"Failed to terminate the job due to a ServerException:\n\t{error}",
+                f"Failed to terminate the job due to: \n\t{error}",
                 job_id=self.pk,
                 batch_job_id=self.batch_job_id,
             )
--- a/api/scpca_portal/models/job.py
+++ b/api/scpca_portal/models/job.py
@@ -5,7 +5,9 @@ from django.db import models
 from django.utils.timezone import make_aware
 
 import boto3
+from botocore.exceptions import ClientError
 
+from scpca_portal.config.logging import get_and_configure_logger
 from scpca_portal.enums import JobStates
 from scpca_portal.models import Dataset
 from scpca_portal.models.base import TimestampedModel
@@ -112,9 +114,18 @@ class Job(TimestampedModel):
         """
         return boto3.client("batch", region_name=settings.AWS_REGION)
 
+    @property
+    def _logger(self):
+        """
+        Logger object for logging message.
+        """
+        return get_and_configure_logger(__name__)
+
     def submit(self) -> None:
-        """Submit a job via boto3, update batch_job_id and state, and
-        save the job object to the db"""
+        """
+        Submit a job via boto3, update batch_job_id and state, and
+        save the job object to the db.
+        """
 
         response = self._batch.submit_job(
             jobName=self.batch_job_name,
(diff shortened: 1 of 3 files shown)


api/scpca_portal/test/models/test_job.py
```diff
@@ -4,64 +4,102 @@
...
         self.assertIsNone(job.id)
 
         job.submit()
-
-        mock_batch_client.submit_job.assert_called_once_with(
-            jobName=self.batch_job_name,
-            jobQueue=settings.AWS_BATCH_JOB_QUEUE_NAME,
-            jobDefinition=settings.AWS_BATCH_JOB_DEFINITION_NAME,
-            containerOverrides=self.project_job_batch_container_overrides,
-        )
```
nozomione, line 49: Since we're checking the saved job instance state in the database, I've omitted the check before saving the instance, as they are identical. I've applied the same logic to the tests for `Job::terminate` as well.

api/scpca_portal/test/factories.py
nozomione, line 1: @davidsmejia since you suggested adding a job factory to optimize the `JobTest`, I went ahead and added it. Thank you!

api/scpca_portal/models/job.py
```diff
@@ -112,9 +114,18 @@ def _batch(self):
         """
         return boto3.client("batch", region_name=settings.AWS_REGION)
 
+    @property
+    def _logger(self):
```
davidsmejia, line 118: Handle logger the same way as `models/contact.py` no need to bring it in as a model property.

api/scpca_portal/models/job.py
```diff
@@ -131,10 +142,51 @@ def submit(self) -> None:
...
+            self.critical_error = True
+            self.state = JobStates.TERMINATED
+            self._logger.exception(
+                f"Failed to terminate the job due to a ServerException:\n\t{error}",
+                job_id=self.pk,
+                batch_job_id=self.batch_job_id,
+            )
+
+            self.save()
+            return False
```
davidsmejia, line 182: 1. all function returns should have a consistent return type. So if you exit early and return false you need to return true as the base case.
2. I don't think you need to log out if the job state was wrong when the terminate was attempted. Instead just exit. 
3. `critical_error` does not mean we were unable to terminate a job, that is reserved for an unrecoverable error that occurs during processing. You can remove both that line and the line that sets the state in the exception block.
4. If you ever set the value of state to be `TERMINATED` then you need to also set the timestamp for `terminated_at` as well.

I think you can rewrite this so that the return type is always the bool if the job was successful or not.
Arguably you could also return true if the job is already terminated.

api/scpca_portal/test/factories.py
```diff
@@ -204,3 +206,47 @@ class Meta:
...
+    # Test-only fields (not part of the model)
+    project_id = None
+    sample_id = None
+    download_config_name = "MOCK_DOWNLOAD_CONFIG"
+    resource_flag = factory.LazyAttribute(
+        lambda obj: "--project-id" if obj.project_id else "--sample-id"
+    )
+    resource_id = factory.LazyAttribute(
+        lambda obj: obj.project_id if obj.project_id else obj.sample_id
+    )
```
davidsmejia, line 231: Can you think of a way to remove these. I would strongly discourage changing the model we are meant to be testing in order to test. Especially since we define sample_id here then only reference project_id below.

davidsmejia (commented):
Great start, lets make sure we handle logging, return types, and exceptions better before we merge this in.

api/scpca_portal/models/job.py
nozomione: I’ve updated `Job::terminate` to handle the following cases in 40f6bba:
- Returns `True` if:
  - The job is successfully terminated
  - The job has already been terminated
- Returns `False` if:
  - The job’s state is `COMPLETED`
  - A `ClientException` occurs
  - A `ServerException` occurs

I've added info logger for successful termination, but let me know if it's unnecessary or if returning `True` alone is sufficient. Thank you, David!

api/scpca_portal/test/factories.py
nozomione: I've removed the test-only fields from `JobFactory` and updated the corresponding tests reflect these changes. Thank you for this insight!

nozomione:
I've applied the feedback and this PR is ready for another look. Thank you, David!

api/scpca_portal/models/job.py
```diff
@@ -131,10 +137,42 @@ def submit(self) -> None:
...
+            self.terminated_at = make_aware(datetime.now())
+
+            self.save()
+        except ClientError as error:
+            logger.exception(
+                f"Failed to terminate the job due to a ClientException:\n\t{error}",
+                job_id=self.pk,
+                batch_job_id=self.batch_job_id,
+            )
+            return False
```
davidsmejia, line 161: This looks better but let's also get rid of the `ClientError` exception as capturing that doesn't provide any actionable information vs a regular `Exception`.

api/scpca_portal/models/job.py
```diff
@@ -131,10 +137,42 @@ def submit(self) -> None:
 
     def terminate(self, retry_on_termination=False):
```
davidsmejia, line 138: you can add type hinting for the return type here as well.

nozomione:
I've applied your feedback and this PR is ready for another look. Thank you, David!

api/scpca_portal/test/models/test_job.py
```diff
@@ -4,64 +4,102 @@
...
+        mock_batch_client.submit_job.assert_called_once()
 
         # After submission, the job should be updated
-        self.assertEqual(job.batch_job_id, self.mock_job_id)
+        self.assertEqual(job.batch_job_id, self.mock_batch_job_id)
         self.assertEqual(job.state, JobStates.SUBMITTED)
         self.assertIsInstance(job.submitted_at, datetime)
```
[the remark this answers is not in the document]
davidsmejia: Yeah I think you could test that before or after submitting the job. This looks good to me.

api/scpca_portal/test/models/test_job.py
davidsmejia, line 1: instead if `response = job.terminate` can we change this to `success = `. This should go with the type hinting for the return type as well as indicating in the docstring to match the current implementation.

nozomione: Great catch 👍, thank you!

api/scpca_portal/test/models/test_job.py
```diff
@@ -6,62 +6,96 @@
...
+        # Set up mock for the TERMINATED job
+        terminated_job = JobFactory(
+            batch_job_name=self.mock_project_batch_job_name,
+            batch_job_id=self.mock_batch_job_id,
+            state=JobStates.TERMINATED,
+        )
+
+        # Should return Ture for previously terminated job
+        success = terminated_job.terminate(retry_on_termination=True)
+        mock_batch_client.terminate_job.assert_called()
```
davidsmejia, line 80: based on my read this should fail so lets take a closer look

api/scpca_portal/test/models/test_job.py
```diff
@@ -6,62 +6,96 @@
...
+        # Set up mock to raise an exception
+        mock_batch_client.terminate_job.side_effect = Exception("Exception")
+
+        success = job.terminate(retry_on_termination=True)
+        mock_batch_client.terminate_job.assert_called()
+        self.assertFalse(success)
+
+        # The job should not be updated
+        saved_job = Job.objects.first()
+        self.assertEqual(saved_job.state, JobStates.SUBMITTED)
```
davidsmejia, line 100:
```suggestion
        self.assertEqual(saved_job.state, job.state)
```

Let's assert that the value hasn't changed by checking against the un-refreshed value

api/scpca_portal/test/models/test_job.py
nozomione: I separated the tests for the submitted jobs (`test_terminate_job`) and terminated jobs (`test_terminate_job_on_terminated_job`) to prevent conflicting assertions.

nozomione: Alternatively, instead of separating the tests, we can assert that the mock method wasn't called twice for the terminated job by using the following assertion:
- `self.assertFalse(mock_batch_client.terminate_job.call_count, 2)` (this may be a better approach?) 
- `mock_batch_client.terminate_job.assert_called_once()` (make sure called only once within the test)

Let me know your thoughts on this.

nozomione: Alternatively, instead of separating the tests, we can assert that the mock method wasn't called twice for the terminated job by using one of the following assertions:
- `self.assertEqual(mock_batch_client.terminate_job.call_count, 1)` (this may be a better approach?) 
- `mock_batch_client.terminate_job.assert_called_once()`  

Let me know your thoughts on this.

api/scpca_portal/models/job.py
```diff
@@ -129,12 +134,38 @@ def submit(self) -> None:
...
-        Terminate the currently running job via boto3, and update state.
-        Set critical_error to True if the job is irrecoverable.
+        Terminate the submitted and incomplete job via boto3, and update state.
+        Return True if the job is successfully terminated or already terminated, otherwise False.
+        Throw an error if failed to terminate the job.
         """
-        pass
+
+        if self.state in [JobStates.COMPLETED, JobStates.TERMINATED]:
+            return self.state == JobStates.TERMINATED
```
avrohomgottlieb, line 145: General question - should we not be returning True if a Job is in a COMPLETED state as well?

api/scpca_portal/test/models/test_job.py
```diff
@@ -6,62 +6,95 @@
...
+            batch_job_id=self.mock_batch_job_id,
+            state=JobStates.SUBMITTED,
+        )
+
+        success = submitted_job.terminate(retry_on_termination=True)
+        mock_batch_client.terminate_job.assert_called_once()
+        self.assertTrue(success)
+
+        # After termination, the job should be saved with correct field values
+        saved_job = Job.objects.first()
```
avrohomgottlieb, line 65:
```suggestion
        saved_job = Job.objects.filter(pk=submitted_job.pk).first()
```
I feel like this is more correct.

api/scpca_portal/test/models/test_job.py
```diff
@@ -6,62 +6,95 @@
...
+
+        # Set up mock to raise an exception
+        mock_batch_client.terminate_job.side_effect = Exception("Exception")
+
+        success = job.terminate(retry_on_termination=True)
+        mock_batch_client.terminate_job.assert_called_once()
+        self.assertFalse(success)
+
+        # The job state should remain the same
+        saved_job = Job.objects.first()
```
avrohomgottlieb, line 99:
```suggestion
        saved_job = Job.objects.filter(pk=job.pk).first()
```
Same as above.

avrohomgottlieb (commented):
Looks great.  A couple of comments, but I think you did a great job here.

api/scpca_portal/models/job.py
nozomione: Great question -we've decided to leave this as is for now and revisit it later if necessary.

api/scpca_portal/test/models/test_job.py
davidsmejia: Just want to add some more input here:
1. for these assertions you could also do `refresh_from_db` and make you assertions because you know the values you want them to be and you aren't testing that they have changed.
2. If you want to detect if the job's value have changed it makes sense to fetch a new instance of the same job model as you do here
3. Since each test is a transaction you can also add some completeness by assert that the number of jobs in the DB is only 1. This is less beneficial in this case but would be helpful when you are testing functionality that creates jobs. However, here you are instatiating through the factory method so not that helpful.

nozomione:
I've applied your feedback and this PR is ready for another look. Thank you, David and Avrohom!

api/scpca_portal/test/models/test_job.py
nozomione: I've improved readability by swapping the order of assertions for `TERMINATED` and `SUBMITTED` jobs, and replaced `call_count` with `assert_not_called` for the `TERMINATED` job in d7089e2. Thank you!

davidsmejia (approved):
LGTM

api/scpca_portal/test/factories.py
```diff
@@ -204,3 +207,32 @@ class Meta:
...
+                "manage.py",
+                "generate_computed_file",
+                (
+                    "--project-id"
+                    if obj.batch_job_name.startswith(common.PROJECT_ID_PREFIX)
+                    else "--sample-id"
+                ),
+                (
+                    "SCPCP000000"
+                    if obj.batch_job_name.startswith(common.SAMPLE_ID_PREFIX)
```
avrohomgottlieb, line 231:
```suggestion
                    if obj.batch_job_name.startswith(common.PROJECT_ID_PREFIX)
```

avrohomgottlieb (approved):
One small edit to make, but looks ready to be merged in. Great job!

Issue #1142, opened 2025-03-12
Implement Notification context hook and component with Storybook


nozomione:
### Context

Figma: [Download Process](https://www.figma.com/board/RBhllWGPSTUFSYmMgomtuy/carty-b-design-discussions?node-id=464-25793&t=Z5JghO7jjX3BpU0T-0)(The notification UI is in the last frame of `Move to My Dataset` section)

We are working on the cart feature for the portal and preparing the UIs for the Dataset Add Samples components.

### Problem or idea

As part of the linked epic, we need to implement the Notification context along with its corresponding hook and component based on the Figma design. It will initially use mocked data via Storybook.

### Solution or next step

Implement the Notification context and its hook and component in Storybook using mock data.

PR #1152, opened 2025-03-17
1142 - Add Notification Component

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1142

This PR added the Notification context, hook, and component. 
The component was also added to storybook, and can be accessed [here](https://scpca-portal-l2w9demk8-ccdl.vercel.app/?path=/story/components-notification--default).

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


client/.storybook/stories/Notification.stories.js
```diff
@@ -0,0 +1,81 @@
...
+      <Box
+        align="center"
+        gap="large"
+        justify="center"
+        margin={{ top: 'large' }}
+        width='fill'>
+          <Success />
+          <Error />
+          <MoveSamplesSuccess />
+
```
nozomione, line 78: I noticed an extra line here.

```suggestion
```

nozomione (approved):
Great job, Avrohom 👍🚀 !

davidsmejia (commented):
Per our conversation during ticket time can we update the following:
- change `Notification` to `Notifications` and include at top of story layout, inside of context instead of rendering in place
- Notification components should be rendered inside of loop in `Notifications` component (can be in the same file)
- context should be an array of notifications that are looped over in `Notifications` component, each notification data structure can contain message, and type (success or error)

client/src/contexts/NotificationContext.js
```diff
@@ -0,0 +1,38 @@
...
+    setNotifications((prev) => [
+      ...prev,
+      {
+        [id]: {
+          id,
+          message,
+          type
+        }
+      }
+    ])
```
nozomione, line 26: We can simplify the `notifications` data structure by removing the outer `id` key:

```suggestion
    setNotifications((prev) => [
      ...prev,
      {
          id,
          message,
          type
      }
    ])
```

client/src/contexts/NotificationContext.js
```diff
@@ -0,0 +1,38 @@
...
+export const NotificationContextProvider = ({ children }) => {
+  const [notifications, setNotifications] = useState([
+    {
+      test: {
+        id: 'test',
+        type: 'success',
+        message: 'this is fun!'
+      }
+    }
+  ])
```
nozomione, line 14: Since we removed the nested object, we can also clean up the outer dynamic key (e.g., 'test') from the mock data:

```suggestion
  const [notifications, setNotifications] = useState([
    {
        id: 'test',
        type: 'success',
        message: 'this is fun!'
    }
  ])
```

client/src/components/Notifications.js
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <>
+      {notifications.map((n) => {
+        const notification = Object.values(n)[0]
+        return (
+          <Notification key={notification.id} notification={notification} />
+        )
+      })}
+    </>
+  )
```
nozomione, line 66: Now we can simply pass each element of `notifications` to the `Notification` component:
  
```suggestion
  return (
    <>
      {notifications.map((n) => (
        <Notification key={n.id} notification={n} />
      ))}
    </>
  )
```

nozomione (approved):
I've left a few suggestions. Otherwise, LGTM - Great job, Avrohom 🚀 !

client/.storybook/stories/Notifications.stories.js
```diff
@@ -0,0 +1,11 @@
...
+import { useNotification } from 'hooks/useNotification'
+import { Box } from 'grommet'
+import { Button } from 'components/Button'
+import { Notifications } from 'components/Notifications'
+
+export default {
+    title: 'Components/Notifications'
+}
+
+export const Default = () => <Notifications />
```
davidsmejia, line 11:
```suggestion
export const Default = () => {
  const { showNotification } = useNotification()
  return (
    <Notifications />
    <Button onClick={() => showNotification(id, type, message)}/>
  )
}
```

if you want to switch the order, i would make it message, id, type where id is auto set as `Date.now()` and type should be `success` in the function definition

client/src/contexts/NotificationContext.js
davidsmejia, line 26:
```suggestion
  const showNotification = (message, id, type = 'success') =>
    setNotifications((prev) => [
      ...prev,
     {
     id: id || Date.now()
      message,
      type
    }
    ])
```

I would recommend switching the order

davidsmejia (approved):
LGTM 🚀

Issue #1160, opened 2025-03-20
Tests for CCDL Dataset


avrohomgottlieb:
_Please label your issue with a label to specify its priority: breaking, high, medium, or low._

### Context

In PR https://github.com/AlexsLemonade/scpca-portal/pull/1154 we added building and validation methods for CCDL datasets, which included `get_ccdl_dataset`, `get_ccdl_data`, `existng_ccdl_datasets` and `should_process`.

### Problem or idea

We should add tests for these methods to validate intended inputs and outputs.

### Solution or next step

...

PR #1165, opened 2025-03-26
1160 - CCDL Dataset Tests

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1160

This PR adds a test for `Dataset::get_or_find_ccdl_dataset`. Portal and project ccdl datasets are checked, along with all ccdl_dataset types.

## Types of changes

## Functional tests

- `test_get_or_find_ccdl_dataset`

## Checklist

## Screenshots

N/A


davidsmejia (commented):
This looks good to me.

One question before we lock this in:
Should we change the convention for `get_or_find` to return `(thing, found)` where found is true when fetched from the DB and false when it is populated? If you prefer this let's update the test to use found as the variable name instead of `is_new_dataset`. Thoughts?

avrohomgottlieb:
> This looks good to me.
> 
> One question before we lock this in: Should we change the convention for `get_or_find` to return `(thing, found)` where found is true when fetched from the DB and false when it is populated? If you prefer this let's update the test to use found as the variable name instead of `is_new_dataset`. Thoughts?

I like this. I was nervous to use `created` as we had originally decided upon (in line with Django conventions) because it differs from our usage of the term `create` which we use to denote "added to the DB". I like `found`, which reflects the function name.

api/scpca_portal/models/dataset.py
```diff
@@ -87,12 +87,12 @@ def get_or_find_ccdl_dataset(
...
             is_ccdl=True, ccdl_name=ccdl_name, ccdl_project_id=project_id
         ).first():
-            return dataset, False
+            return dataset, True
 
         dataset = cls(is_ccdl=True, ccdl_name=ccdl_name, ccdl_project_id=project_id)
         dataset.format = dataset.ccdl_type["format"]
         dataset.data = dataset.get_ccdl_data()
-        return dataset, True
+        return dataset, False
```
nozomione, line 95: Suggestion: Can we utilize Django built-in [`.get_or_create`](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#get-or-create) shortcut method to optimize the logic by using the `created` boolean flag?

>  Returns a tuple of `(object, created),` where object is the retrieved or created object and created is a boolean specifying whether a new object was created.

nozomione (approved):
I've added a minor suggestion - otherwise 👍

api/scpca_portal/models/dataset.py
davidsmejia: We had discussed this a bit before landing on the current implementation. The difference is that we don't want to return a saved version of the dataset. Django doesn't allow for us to override that behavior with their `get_or_create` convenience method.

It would be nice if there was an existing django convention / convenience method that would allow for returning either a retrieved instance or a new unsaved instance. If a better option becomes available switching it over wouldn't be that difficult so for now I think we can keep it this way.

davidsmejia (approved):
Suggesting that we assert that the datasets we expect to be the same, in fact, are the same.

Otherwise LGTM

PR #1167, opened 2025-03-28
1146 - Implement Job::sync_state and Job::bulk_sync_state methods using the existing Job model instance

### LINKED ISSUE #1146: Implement Job::sync_state and Job::bulk_sync_state methods using the existing Job model instance
### Context

The `Job` model should implement synchronization methods to update the job instance's state with the corresponding remote AWS Batch job status, both individually and in bulk.

### Problem or idea

We'll implement a `sync_state` instance method and a `bulk_sync_state` class method  and their corresponding tests using mock data.

The status mapping between AWS Batch and a job instance is as follows:

| AWS Batch | `Job` Instance |
|:--|:--|
| `SUBMITTED` | `SUBMITTED` | 
|  `PENDING` | `SUBMITTED` |
| `RUNNABLE` | `SUBMITTED` |
| `STARTING` |`SUBMITTED` | 
|  `RUNNING` | `SUBMITTED` |
| `SUCCEEDED` | `COMPLETED`  | 
| `FAILED` | `COMPLETED` or `TERMINATED` | 

The `sync_state` method should:
- Fetch the current status from AWS Batch for a `SUBMITTED` job instance
- Updates the instance's `state` and`state_reason` fields to match AWS Batch response:
   - If a job status is `FAILED`,  update `failure_reason` if it hasn't been set already.
   - If a job status is `FAILED`  and `isCanceled` or `isTerminated` is `True` ,  update `terminated_at` if it hasn't been set already.
- Save the instance to the database

The `bulk_sync_state` method should: 
- Fetch the current status from AWS Batch for all `SUBMITTED` job instances
- Updates each job instance's `state` (the same logic applies for `FAILED` stauts)
- Save them to the database

The test for the `sync_state` method should verify the following flow using mock data:
1. Fetch the AWS Batch status for a `SUBMITTED` job instance
2. Update the fields based on the response
3. Make sure the job instance is saved with the appropriate field values

For test for the `bulk_sync_state` method, it should test against multiple job instances instead of one.

### Solution or next step
- Implement the `sync_state` and `bulk_sync_state`methods
- Implement the `get_jobs` method in the `batch` module
- Add the corresponding test using mocked methods and data

nozomione:
## Issue Number

Closes #1146

I've implemented the `Job::sync_state` instance method and `Job::bulk_sync_state` class method, and added the corresponding tests.

Other changes include:
- Added the `get_jobs` method to the `batch` module.
- Added `utils.get_chunk_list` for `boto3.describe_jobs` job ID size limit per request (I've used the solution from [Stack overflow: Split a list into equally-sized chunks](https://stackoverflow.com/a/312464), as using generators seems to be a commonly used approach. Please let me know if you have any other suggestions!)
- Defined `batch_job_id` using `factory.Sequence` and the default `batch_job_name` in `JobFactory`, and cleaned up `setUp` and no longer required arguments in `TestJob`

## Types of changes

## Functional tests

- `TestJob::test_sync_state`
- `TestJob::test_bulk_sync_state`
- `TestJob::test_bulk_sync_state_no_matching_batch_job_found`

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/batch.py
+++ b/api/scpca_portal/batch.py
@@ -1,6 +1,9 @@
+from typing import Dict, List
+
 import boto3
 from botocore.client import Config
 
+from scpca_portal import utils
 from scpca_portal.config.logging import get_and_configure_logger
 
 logger = get_and_configure_logger(__name__)
@@ -20,7 +23,6 @@ def submit_job(job) -> str | None:
             jobDefinition=job.batch_job_definition,
             containerOverrides=job.batch_container_overrides,
         )
-
     except Exception as error:
         logger.exception(
             f"Failed to terminate the job due to: \n\t{error}",
@@ -47,7 +49,6 @@ def terminate_job(job) -> bool:
     """
     try:
         aws_batch.terminate_job(jobId=job.batch_job_id, reason="Terminating job.")
-
     except Exception as error:
         logger.exception(
             f"Failed to terminate the job due to: \n\t{error}",
@@ -62,3 +63,42 @@ def terminate_job(job) -> bool:
         batch_job_id=job.batch_job_id,
     )
     return True
+
+
+def get_job(batch_job_id: str) -> Dict | None:
+    """
+    Fetch an AWS Batch job for the given job ID.
+    Return the fetched job on success, otherwise return None.
+    """
+    try:
+        response = aws_batch.describe_jobs(jobs=[batch_job_id])
+    except Exception as error:
+        logger.exception(
+            f"Failed to fetch AWS Batch job due to: \n\t{error}",
+            batch_job_id=batch_job_id,
+        )
+        return None
+
+    return response["jobs"][0]
+
+
+def get_jobs(batch_job_ids: List[str]) -> List[Dict] | None:
+    """
+    Bulk fetch AWS Batch jobs by the given job IDs.
+    Return a list of fetched jobs on success, otherwise return None.
+    """
+    max_limit = 100  # Limit of job IDs to send per request
+
+    result = []
+
+    try:
+        for chunk in utils.get_chunk_list(batch_job_ids, max_limit):
+            response = aws_batch.describe_jobs(jobs=chunk)
+            result.extend(response["jobs"])
+    except Exception as error:
+        logger.exception(
+            f"Failed to bulk fetch AWS Batch jobs due to: \n\t{error}",
+        )
+        return None
+
+    return result
--- a/api/scpca_portal/batch.py
+++ b/api/scpca_portal/batch.py
@@ -1,5 +1,7 @@
 from typing import Dict, List
 
+from django.template.defaultfilters import pluralize
+
 import boto3
 from botocore.client import Config
 
@@ -65,40 +67,24 @@ def terminate_job(job) -> bool:
     return True
 
 
-def get_job(batch_job_id: str) -> Dict | None:
-    """
-    Fetch an AWS Batch job for the given job ID.
-    Return the fetched job on success, otherwise return None.
-    """
-    try:
-        response = aws_batch.describe_jobs(jobs=[batch_job_id])
-    except Exception as error:
-        logger.exception(
-            f"Failed to fetch AWS Batch job due to: \n\t{error}",
-            batch_job_id=batch_job_id,
-        )
-        return None
-
-    return response["jobs"][0]
-
-
 def get_jobs(batch_job_ids: List[str]) -> List[Dict] | None:
     """
-    Bulk fetch AWS Batch jobs by the given job IDs.
+    Fetch AWS Batch job(s) for the given one or more job ID(s) in bulk.
     Return a list of fetched jobs on success, otherwise return None.
     """
     max_limit = 100  # Limit of job IDs to send per request
 
-    result = []
+    jobs = []
 
     try:
         for chunk in utils.get_chunk_list(batch_job_ids, max_limit):
             response = aws_batch.describe_jobs(jobs=chunk)
-            result.extend(response["jobs"])
+            jobs.extend(response["jobs"])
     except Exception as error:
         logger.exception(
-            f"Failed to bulk fetch AWS Batch jobs due to: \n\t{error}",
+            f"Failed to bulk fetch AWS Batch job{pluralize(len(batch_job_ids))} "
+            f"for job IDs: {', '.join(batch_job_ids)} due to: \n\t{error}"
         )
         return None
 
-    return result
+    return jobs
--- a/api/scpca_portal/batch.py
+++ b/api/scpca_portal/batch.py
@@ -1,4 +1,4 @@
-from typing import Dict, List
+from typing import Dict, Iterable, List
 
 from django.template.defaultfilters import pluralize
 
@@ -67,24 +67,24 @@ def terminate_job(job) -> bool:
     return True
 
 
-def get_jobs(batch_job_ids: List[str]) -> List[Dict] | None:
+def get_jobs(batch_jobs: Iterable["Job"]) -> List[Dict] | None:  # noqa: F821
     """
-    Fetch AWS Batch job(s) for the given one or more job ID(s) in bulk.
+    Fetch AWS Batch job(s) for the given one or more job(s) in bulk.
     Return a list of fetched jobs on success, otherwise return None.
     """
     max_limit = 100  # Limit of job IDs to send per request
-
     jobs = []
 
-    try:
-        for chunk in utils.get_chunk_list(batch_job_ids, max_limit):
-            response = aws_batch.describe_jobs(jobs=chunk)
-            jobs.extend(response["jobs"])
-    except Exception as error:
-        logger.exception(
-            f"Failed to bulk fetch AWS Batch job{pluralize(len(batch_job_ids))} "
-            f"for job IDs: {', '.join(batch_job_ids)} due to: \n\t{error}"
-        )
-        return None
+    if batch_job_ids := [job.batch_job_id for job in batch_jobs]:
+        try:
+            for chunk in utils.get_chunk_list(batch_job_ids, max_limit):
+                response = aws_batch.describe_jobs(jobs=chunk)
+                jobs.extend(response["jobs"])
+        except Exception as error:
+            logger.exception(
+                f"Failed to bulk fetch AWS Batch job{pluralize(len(batch_job_ids))} "
+                f"for job IDs: {', '.join(batch_job_ids)} due to: \n\t{error}"
+            )
+            return None
 
     return jobs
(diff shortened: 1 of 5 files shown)


api/scpca_portal/test/models/test_job.py
```diff
@@ -66,6 +95,129 @@ def test_submit_job_failure(self, mock_batch_submit_job):
...
+        success = Job.bulk_sync_state()
+        mock_batch_get_jobs.assert_called()
+        self.assertTrue(success)
+
+        # Job with state change should be updated and saved with correct field values
+        saved_job = Job.objects.filter(batch_job_id=f"{self.mock_batch_job_id}-{1}").first()
+        self.assertEqual(saved_job.state, JobStates.COMPLETED)
+        self.assertEqual(saved_job.failure_reason, "Job FAILED")
+        self.assertIsInstance(saved_job.completed_at, datetime)
+
```
nozomione, line 220: I've temporarily included the test for for the missing AWS batch job.

api/scpca_portal/utils/helpers.py
```diff
@@ -42,6 +42,14 @@ def join_workflow_versions(workflow_versions: Set) -> str:
...
 
 
+def get_chunk_list(list: List[Any], size: int) -> Generator[List[Any], None, None]:
+    """
+    Yields chunks of the given list with the specified size.
+    """
+    for i in range(0, len(list), size):
+        yield list[i : i + size]
+
+
```
nozomione, line 52: I've used the solution from [Stack overflow: Split a list into equally-sized chunks](https://stackoverflow.com/a/312464) for list chunking, as using generators seems to be a commonly used approach. Please let me know if you have any other suggestions!

api/scpca_portal/models/job.py
```diff
@@ -104,6 +122,50 @@ def get_sample_job(
...
+                            if aws_job_status == "FAILED" and not job.failure_reason:
+                                job.failure_reason = aws_job["statusReason"]
+
+                            job.state = mapped_job_state
+                            job.completed_at = make_aware(datetime.now())
+                            synced_jobs.append(job)
+                    else:
+                        # TODO: How should we handle when no matched AWS job returned?
+                        # Currently, it continues without interruption - log, or raise an error?
+                        continue
```
nozomione, line 158: While implementing the tests, I encountered an error due to a missing matched batch job (thanks to my typo 😅). This made me consider the scenario where no matching AWS job is found via the API call. Although this may be unlikely, if it does happen, it could halt the execution of the bulk syncing process. Just to be safe, should we handle this case?

api/scpca_portal/test/models/test_job.py
```diff
@@ -18,6 +18,40 @@ def setUp(self):
...
+        return [
+            {
+                "jobId": job["jobId"],
+                "status": job["status"],
+                "statusReason": f"Job {job['status']}",
+                "isTerminated": terminated_job and job["jobId"] == terminated_job["jobId"],
+            }
+            for job in list_of_jobs
+        ]
```
nozomione, line 54: I've added these helpers within `TestJob`. Should we move them to `JobFactory`?

api/scpca_portal/models/job.py
```diff
@@ -104,6 +121,59 @@ def get_sample_job(
...
+                            if is_terminated:
+                                job.terminated_at = make_aware(datetime.now())
+
+                            job.state = mapped_job_state
+                            job.completed_at = make_aware(datetime.now())
+                            synced_jobs.append(job)
+                    else:
+                        # TODO: How should we handle when no matched AWS job returned?
+                        # Currently, it continues without interruption - log, or raise an error?
+                        continue
```
nozomione, line 165: While implementing the tests, I encountered an error due to a missing matched batch job (thanks to my typo 😅). This made me consider the scenario where no matching AWS job is found via the API call. Although this may be unlikely, if it does happen, it could halt the execution of the bulk syncing process. Just to be safe, should we handle this case?

api/scpca_portal/batch.py
```diff
@@ -62,3 +65,26 @@ def terminate_job(job) -> bool:
         batch_job_id=job.batch_job_id,
     )
     return True
+
+
+def get_jobs(batch_job_ids: List[str]) -> List[Dict] | None:
```
davidsmejia, line 70: Can we update this method to take in an `Iterable` of `Job` instances?

api/scpca_portal/test/models/test_job.py
```diff
@@ -18,6 +18,40 @@ def setUp(self):
...
+            "SUBMITTED",
+            "PENDING",
+            "RUNNABLE",
+            "STARTING",
+            "RUNNING",
+            "SUCCEEDED",
+            "FAILED",
+        ]
+
+    def bulk_create_mock_jobs(self, list_of_jobs):
```
davidsmejia, line 31:
```suggestion
    def create_mock_jobs(self, list_of_jobs):
```

api/scpca_portal/test/models/test_job.py
```diff
@@ -18,6 +18,40 @@ def setUp(self):
...
+        """Helper to create job instances using JobFactory."""
+        for job in list_of_jobs:
+            JobFactory(
+                batch_job_name=self.mock_project_batch_job_name,
+                batch_job_id=job["batch_job_id"],
+                state=job["state"],
+            )
+
+    def bulk_create_mock_aws_batch_jobs(self, list_of_jobs, terminated_job=None):
```
davidsmejia, line 40:
```suggestion
    def create_mock_aws_batch_jobs(self, list_of_jobs, terminated_job=None):
```

Dont really need to worry about "bulk" in this context as the plurality of "jobs" covers that.

api/scpca_portal/test/models/test_job.py
davidsmejia: I think a JobFactory would be an improvement. It would end up being less logic based than the `create_mock...` implementations which we should avoid when writing tests.

api/scpca_portal/test/models/test_job.py
```diff
@@ -66,6 +100,157 @@ def test_submit_job_failure(self, mock_batch_submit_job):
...
+                self.assertIsInstance(saved_job.terminated_at, datetime)
+            elif saved_job.state == JobStates.COMPLETED.name:
+                self.assertEqual(saved_job.state, JobStates.COMPLETED.name)
+                self.assertIn(saved_job.failure_reason, [None, "Job FAILED"])
+                self.assertIsInstance(saved_job.completed_at, datetime)
+            else:
+                self.assertEqual(saved_job.state, JobStates.SUBMITTED.name)
+                self.assertIsNone(saved_job.failure_reason)
+                self.assertIsNone(saved_job.completed_at)
+                self.assertIsNone(saved_job.terminated_at)
```
davidsmejia, line 225: This would probably be better if we pulled it outside of the same loop.

Instead of doing one query for all of the jobs at the same time you can query them individually and then do a different loop for each expected job.state. You can remove the assertEqual since it will be part of the query.

The last query instead of doing `Job.objects.filter(state=JobStates.SUBMITTED.name)` you could do `Job.objects.exclude(terminated_jobs + completed_jobs)`. This assumes you cached the new queryset for terminated jobs in `terminated_jobs` variable.

davidsmejia (commented):
I agree with needing to use the `JobFactory`. Once you make that change I think this will be a lot cleaner.

api/scpca_portal/batch.py
nozomione: I've updated the type hint to `Iterable["Job"]` (using a string reference to avoid a circular import with the `Job` model). But, this results in a Flake8 error (`undefined name 'Job'`). To resolve this, I've used Flake8's [In-line error ignoring](https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html) (i.e., # noqa: F821) for this line temporarily. Let me know your thoughts on this. Thank you!

nozomione:
I've applied your feedback, and this PR is ready for another look. Thank you, David!

api/scpca_portal/test/models/test_job.py
```diff
@@ -66,6 +75,163 @@ def test_submit_job_failure(self, mock_batch_submit_job):
...
+                )
+            ],
+            terminated_job={"jobId": f"{self.mock_batch_job_id}-7", "status": "FAILED"},
+        )
+        JobFactory.get_mock_jobs(
+            [
+                {"batch_job_id": f"{self.mock_batch_job_id}-{i}", "state": JobStates.SUBMITTED.name}
+                for i in range(0, 8)
+            ]
+        )
```
davidsmejia, line 181:
```suggestion
        jobs = [JobFactory() for _ in range(7)]
                
        mock_response = [{"jobId": job.aws_batch_id} for job in jobs]
        
        mock_response[0]["status"] = "PENDING"
        # .... all other jobs
        mock_response[4]["status"] = "TERMINATED"
        
        mock_batch_get_jobs.return_value = mock_response
        
```

You should be able to use `factory.Sequence(lambda n: f"batch-id{n}")` to dynamically define the `aws_batch_job_id` attributes

api/scpca_portal/test/models/test_job.py
```diff
@@ -66,6 +75,163 @@ def test_submit_job_failure(self, mock_batch_submit_job):
...
+                {"jobId": f"{self.mock_batch_job_id}-{i}", "status": status}
+                for i, status in enumerate(["PENDING", "FAILED"])
+            ]
+        )
+        JobFactory.get_mock_jobs(
+            [
+                {"batch_job_id": f"{self.mock_batch_job_id}-{i}", "state": JobStates.SUBMITTED.name}
+                for i in range(1, 4)
+            ]
+        )
```
davidsmejia, line 222: Same comment as above.

nozomione:
I've updated the `JobFactory` as follows:
- Defined the `batch_job_name` using `factor.Sequence`
- Removed the helpers for bulk generating mock data

Additionally, I've add a default `batch_job_name` in `JobFactory` and adjusted the `TestJob` (e.g., cleaned up `setUp()` and passed arguments in each tests).

This PR is ready for another look. Thank you, David!

nozomione:
Based on your feedback, I've updated the `Job::get_mapped_job_state` method and refactored `Job::sync_state` (adjusted the test to reflect the return value when no changes were made) and `Job::bulk_sync_state`. This PR is ready for your review. Thank you, David!

api/scpca_portal/models/job.py
```diff
@@ -45,6 +47,38 @@ def __str__(self):
             return f"Job {self.id} - {self.batch_job_id} - {self.state}"
         return f"Job {self.id} - {self.state}"
 
+    @staticmethod
+    def update_job_state(job, aws_job) -> tuple[Self, bool]:
```
davidsmejia, line 51:
```suggestion
    def update_job_state(job: Self, aws_job: dict) -> tuple[Self, bool]:
```

davidsmejia (approved):
LGTM just wanted to add the typing back to `update_job_state`

avrohomgottlieb (approved):
LGTM 🚀 
Awesome PR Nozomi

Issue #1108, opened 2025-02-14
Dataset Original Files Query


avrohomgottlieb:
### Context

PRs https://github.com/AlexsLemonade/scpca-portal/pull/1060 and https://github.com/AlexsLemonade/scpca-portal/pull/1105 introduced the S3KeyInfo construct as a means of clarifying the setting of OriginalFile attributes. This has prepared the OriginalFile model for its desired usage of optimized querying for computed file generation. Ultimately, the query for OriginalFiles will occur on the Dataset model, which will end up being the center piece of Computed File generation.

### Problem or idea

For the Dataset model, we'd like to develop a direct, clear and efficient query for OriginalFiles.

### Solution or next step

```python
  class Dataset:
    ...

    @property
    def original_files(self):
        files = OriginalFile.objects.none()
        for project_id, config in self.data.items():
            
            # single cell
            single_cell = OriginalFile.downloadable_objects.filter(
                project_id=project_id, is_single_cell=True
            )
            
            spatial = OriginalFile.downloadable_objects.filter(
                project_id=project_id, is_spatial=True
            )

            single_cell_format = single_cell.filter(format=FORMAT_EXTENSIONS[self.format])
            # grab single cell files
            if config["include_single_cell_merged"]:
                files |= single_cell_format.filter(is_merged=True)
            else:
                files |= single_cell_format.filter(is_merged=False, sample_id__in=config["SINGLE_CELL"]) 
            
            # grab single-cell supplementary
            files |= single_cell.filter(
                project_id=project_id, is_supplementary=True, sample_id__in=config["SINGLE_CELL"]
            )
                
            # grab all spatial files
            files |= spatial.filter(sample_id__in=config["SPATIAL"])
            
            if config["include_bulk"]:
                files |= OriginalFiles.objects.filter(project_id=project_id, is_bulk=True)
            
        return files

    @property
    def original_file_paths(self):
        return [Path(of.s3_key) for of in self.original_files]
                
```

PR #1168, opened 2025-03-31
1108 - Dataset OriginalFile Query

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1108.

This PR introduces an OriginalFile query on the Dataset model. This query is able to grab all of a Dataset's OriginalFiles based on its data and format attributes. A `test_original_files_property` test was also added, which verifies through different configurations that the behavior of the method is as expected.

## Types of changes

## Functional tests

- `test_original_files_property`

## Checklist

## Screenshots

N/A


davidsmejia (approved):
LGTM 🚀

Issue #1147, opened 2025-03-13
Add hashing fields for Datasets


avrohomgottlieb:
### Context

In order to determine whether or not an existing ccdl dataset needs to be regenerated, hash values must be checked.

### Problem or idea

We should add 3 attributes to the Dataset models - hash_readme, hash_metadata, and hash_data.

hash_readme should be calculated by hashing the contents of the readme.
hash_metadata should be calculated by hashing the contents of the metadata file.
hash_data should be calculated by hashing the concatenated string of all original file hashes.

### Solution or next step
- Add 3 attributes
- Add get_hash_readme function
- Add get_hash_metadata function
- Add get_hash_data function
- Integrating into management command

PR #1172, opened 2025-04-03
1147 - Add Dataset Hashes

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1147.

This PR adds to the Dataset model three hashing attributes:
- hash_data
- hash_metadata
- hash_readme

Getter methods for generating these attributes were added as well, in addition to tests which validate that the getter methods work as expected.

## Types of changes

## Functional tests

- test_get_hash_data
- test_get_hash_metadata
- test_get_hash_readme

## Checklist

## Screenshots

N/A


api/scpca_portal/models/dataset.py
```diff
@@ -188,6 +215,22 @@ def original_files(self) -> Iterable[OriginalFile]:
     def original_file_paths(self) -> Set[Path]:
         return {Path(of.s3_key) for of in self.original_files}
 
+    def get_hash_data(self) -> int:
+        original_files_sorted = sorted(self.original_files, key=lambda of: of.s3_key)
```
davidsmejia, line 219: Since `self.original_files` is a queryset could we just get a sorted `values_list` instead of doing this after executing the query?

api/scpca_portal/models/dataset.py
```diff
@@ -188,6 +215,22 @@ def original_files(self) -> Iterable[OriginalFile]:
...
+        concat_hash = "".join(of.hash for of in original_files_sorted)
+        return hash(concat_hash.strip())
+
+    def get_hash_metadata(self) -> int:
+        libraries_metadata = Library.get_libraries_metadata(self.libraries)
+        metadata_file_contents = metadata_file.get_file_contents(libraries_metadata)
+        return hash(metadata_file_contents)
+
+    def get_hash_readme(self) -> int:
```
davidsmejia, line 228: This looks right but we need to account for the `date` in the readme. That first line is going to make this not work day to day. For now I think we should just pop off the first line before calculating the hash.

api/scpca_portal/test/factories.py
```diff
@@ -195,14 +200,14 @@ class Meta:
         "SCPCP999990": {
             "merge_single_cell": False,
             "includes_bulk": True,
-            Modalities.SINGLE_CELL: ["SCPCS999990", "SCPCS999991"],
-            Modalities.SPATIAL: ["SCPCS999992"],
+            Modalities.SINGLE_CELL.name: ["SCPCS999990", "SCPCS999997"],
```
davidsmejia, line 203: This should still be `value`.

api/scpca_portal/models/dataset.py
```diff
@@ -188,6 +215,25 @@ def original_files(self) -> Iterable[OriginalFile]:
...
+        readme_file_contents = readme_file.get_file_contents(
+            self.ccdl_type, Project.objects.filter(scpca_id__in=self.data.keys())
+        )
+        readme_file_contents_no_date = readme_file_contents.split("\n", 1)[1].strip()
```
davidsmejia, line 234:
```suggestion
        # remove first line which contains date
        readme_file_contents_no_date = readme_file_contents.split("\n", 1)[1].strip()
```

davidsmejia (approved):
LGTM

Issue #1140, opened 2025-03-07
Add management command which submits CCDL datasets


avrohomgottlieb:
### Context

Issue https://github.com/AlexsLemonade/scpca-portal/issues/1139 added the ability to create CCDL datasets and validate that they have libraries and samples associated with them.

### Problem or idea

A management command should be added, which iterates over all project and ccdl_dataset_type combinations, as well as portal and ccdl_dataset_type combinations, and queues jobs for submission to Batch.

### Solution or next step

- Add create_ccdl_datasets method, which iterates over all ccdl_dataset_types for portal wide and additionally for each project in DB
- Add has_libraries method on the Dataset model, which checks if a dataset has libraries
- Update necessary attributes on datasets that need to be regenerated
- Saves new and updated datasets
- Add get_dataset_job method on the Job model
- Submit the Job to Batch
- Add test module - test_create_ccdl_datasets

PR #1180, opened 2025-04-07
1140 - Add create_ccdl_dataset Management Command

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1140.

This PR creates valid CCDL Datasets and submits them to Batch.

## Types of changes

## Functional tests

...

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/management/commands/create_ccdl_datasets.py
+++ b/api/scpca_portal/management/commands/create_ccdl_datasets.py
@@ -1,0 +1,44 @@
+import logging
+
+from django.core.management.base import BaseCommand
+
+from scpca_portal import ccdl_datasets
+from scpca_portal.models import Dataset, Job, Project
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+logger.addHandler(logging.StreamHandler())
+
+
+class Command(BaseCommand):
+    help = """
+    Create all ccdl datasets and dispatch them as jobs to AWS Batch.
+    """
+
+    def handle(self, *args, **kwargs):
+        self.create_ccdl_datasets(**kwargs)
+
+    def process_dataset(self, ccdl_name, project_id: str | None = None) -> bool:
+        dataset, found = Dataset.get_or_find_ccdl_dataset(ccdl_name, project_id)
+        if not found and not dataset.libraries:
+            return False
+        if found and not dataset.should_process:
+            return False
+
+        dataset.save()
+        job = Job.get_dataset_job(dataset.id)
+        job.submit()
+
+        logger.info(f"{dataset} job submitted successfully.")
+
+        return True
+
+    def create_ccdl_datasets(self, **kwargs) -> None:
+        for ccdl_name in ccdl_datasets.TYPES:
+            # Project Datasets
+            for project in Project.objects.all():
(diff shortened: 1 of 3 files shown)


api/scpca_portal/management/commands/create_ccdl_datasets.py
```diff
@@ -0,0 +1,44 @@
...
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+logger.addHandler(logging.StreamHandler())
+
+
+class Command(BaseCommand):
+    help = """
+    Create all ccdl datasets and dispatch them as jobs to AWS Batch.
+    """
+
```
avrohomgottlieb, line 17: Do we want to add the `regenerate-all`,  `project-id` and `notify` flags like we do in `dispatch_to_batch`

api/scpca_portal/management/commands/create_ccdl_datasets.py
```diff
@@ -0,0 +1,44 @@
...
+        dataset.save()
+        job = Job.get_dataset_job(dataset.id)
+        job.submit()
+
+        logger.info(f"{dataset} job submitted successfully.")
+
+        return True
+
+    def create_ccdl_datasets(self, **kwargs) -> None:
+        for ccdl_name in ccdl_datasets.TYPES:
```
avrohomgottlieb, line 37: Is it better to iterate over the keys of `ccdl_dataset.TYPES` or over `enums/CCDLDatasetNames` directly?

api/scpca_portal/models/job.py
```diff
@@ -49,6 +49,30 @@ def __str__(self):
...
+            batch_container_overrides={
+                "command": [
+                    "python",
+                    "manage.py",
+                    "generate_computed_file",
+                    "--dataset-id",
+                    dataset_id,
+                    notify_flag,
+                ],
+            },
```
nozomione, line 73: Should we add a job ID to `batch_container_overrides` as well?

api/scpca_portal/models/job.py
```diff
@@ -49,6 +49,30 @@ def __str__(self):
             return f"Job {self.id} - {self.batch_job_id} - {self.state}"
         return f"Job {self.id} - {self.state}"
 
+    @classmethod
+    def get_dataset_job(cls, dataset_id: models.UUIDField, notify: bool = False):
```
nozomione, line 53: Add a return type `Self`?

nozomione (approved):
I've left a few minor suggestions, but otherwise, I'm approving this PR 👍!

api/scpca_portal/models/job.py
avrohomgottlieb: 100% - we definitely want to do this. At this point, the challenge lies in a Job instance not receiving a Job ID until it's saved. The solution would be to refactor this in a way where `batch_container_overrides` is assigned after the Job has been saved, while the other attributes can be populated here.

api/scpca_portal/models/job.py
avrohomgottlieb: Added

Issue #1205, opened 2025-04-21
Implement Dataset Detail Serializer


avrohomgottlieb:
### Context

In Issue https://github.com/AlexsLemonade/scpca-portal/issues/1203, `DatasetViewSet` and `DatasetSerializer` were created.

### Problem or idea

We now want to create a Detail Serializer, which will be called in `DatasetViewSet::get_serializer_class` when the `self.action` is `retrieve`. 

### Solution or next step
- Add a `DatasetDetailSerializer` class to `views/dataset.py` (inheriting from `serializers::DatasetSerializer`), which includes `Projects`, `ComputedFiles` and `Samples`
- Add tests which load the data, make an API request, and verify that the right data and relationships are returned

PR #1227, opened 2025-05-01
1205 - Handle dataset viewset retrieve action

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1205.

This PR adds the ability to send `get` requests to the `/datasets/<id>` endpoint to retrieve individual datasets.

## Types of changes

## Functional tests

## Checklist

## Screenshots

N/A


api/scpca_portal/serializers.py
```diff
@@ -218,6 +218,10 @@ class Meta:
         )
 
 
+class DatasetDetailSerializer(DatasetSerializer):
+    computed_file = ComputedFileSerializer(read_only=True, many=False)
+
+
```
avrohomgottlieb, line 224: @davidsmejia We wanted to add the ability to query projects and samples into this serializer. Samples I understand, as they're essentially the base unit of object represented in the `data` attribute, but I'm having trouble seeing why we would want full project objects, as we only want a subset of all of their associated samples.

davidsmejia: Not worried about samples here. We just want to be able to list related ccdl datasets from the project serializer via a `ccdl_datasets` attribute defined in the project serializer.

davidsmejia (approved):

davidsmejia:
LGTM

PR #1252, opened 2025-05-08
Production Deploy

nozomione:
# Production Deploy
### API
- None

### Client

- Fixes accessing `map` on undefined table `instance.page`
- Adds missing `key` prop to mapped renders
- Fixes `p > p` double nesting
- Fixes misalignment in the `Multiplexed with` column of the samples table


davidsmejia (approved):
LGTM

PR #1253, opened 2025-05-09
fix exit code on test failure

davidsmejia:
## Issue Number

N/A

@avrohomgottlieb discovered that the exit code was zero when tests were failing.

This PR removes the `-e` option from the  hashbang and explicitly sets it at the top of the script which seems to actually set the option.

This PR sets 

## Types of changes

## Functional tests

updated a test to explicitly fail and checked the return code

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/run_tests.sh
+++ b/api/run_tests.sh
@@ -1,4 +1,6 @@
-#!/bin/bash -e
+#!/bin/bash
+
+set -e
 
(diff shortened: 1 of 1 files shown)


avrohomgottlieb (approved):
LGTM

PR #1254, opened 2025-05-09
1235 - Include missing bulk-only samples in metadata-only downloads

### LINKED ISSUE #1235: Include missing bulk-only samples in metadata-only downloads
### Context

Parent issue: #1191

The following four **bulk-only** samples are missing from the `metadata-only` downloads, including both the portal-wide metadata and each project's `single_cell_metadata.tsv` files:

- SCPCP000009;**SCPCS000129**;SCPCL000166
- SCPCP000006;**SCPCS000210**;SCPCL000271
- SCPCP000006;**SCPCS000211**;SCPCL000284
- SCPCP000017;**SCPCS000606**;SCPCL001182

Although these samples are included in dedicated bulk metadata TSV files, they should also be represented in the `metadata-only` download files.

### Problem or idea

The root cause is that we currently handle bulk libraries for bulk-only samples differently during the metadata loading process. During this process, we create `Library` objects in the database using `*_metadata.json` files stored in the S3 input bucket. However, libraries for bulk-only samples do not have associated `*_metadata.json` files, which results in no `Library` objects being created for these samples.

When generating `metadata-only` TSV files, we rely on querying libraries from the database. Since libraries for bulk-only samples do not exist in the database, they are excluded from the output files.

We can resolve this by creating bulk `Library` objects during metadata loading to ensure that all libraries, regardless of modality, are represented in the `metadata-only` files.

### Solution or next step

1. Implement a new method `Project::load_bulk_libraries` that creates bulk library objects during metadata loading usig the downloaded `bulk_metadata.tsv` file from the S3 input bucket with`csv.DictReader`

2. Update `Project.get_libraries` to include bulk libraries when the `metadata_only` flag is set to `True`, or if the project includes bulk data for `single_cell_metadata.tsv`

3. Add tests to verify the implementation work as expected

nozomione:
## Issue Number

Closes #1235

> [!note]
> As a fist step, we've addressed the missing samples in `metadata-only` downloads (portal-wide and project-level metadata files). 
> 
>  For `single_cell_metadata.tsv` changes, we'll handle them in a separate issue. Please see [this comment](https://github.com/AlexsLemonade/scpca-portal/issues/1235#issuecomment-2867882672) for an update  `single_cell_metadata.tsv`.  Thank you!

Previously, 4 bulk-only samples were missing from the generated metadata download files.

This PR ensures that all bulk samples are now included in the following `metadata-only` downloads:
- Portal-wide metadata (i.e., `PORTAL_ALL_METADATA_<accessed-date>.zip`)
- Project-level metadata (i.e., `<project-accession-id>_ALL_METADATA_<accessed-date>.zip`)

Changes made:

- Created bulk library objects in the database during metadata loading
- Assigned the modality `BULK_RNA_SEQ` to bulk libraries
- Updated `Project::get_libraries` to return all libraries for metadata-only downloads

Other change:
- Removed a `TODO` comment and a line that manually added the project ID, since the project ID is now included in metadata JSON file. This prevents the `project_id` column to from being added to the generated metadata TSV files.

For the UI ([comment](https://github.com/AlexsLemonade/scpca-portal/pull/1254#pullrequestreview-2834204379)):
- Excluded bulk libraries from the project summaries in `Project::update_project_aggregate_properties`
- **FE:** Added a check `hasSingleCellData` to render the `"Single-cell"` string in the `ProjectSamplesTable` component

## Types of changes

## Functional tests

The following tests have been updated:

- `test_loader`
- `test_create_portal_metadata`
- `test_dataset`
- `test/expected_values/`

## Checklist

## Screenshots

N/A


davidsmejia (commented):
This looks good to me for now. Just want to confirm with other people that the updated summaries are not going to be an issue. Will approve or request changes based on the outcome of that discussion.

davidsmejia:
Ok for this PR we should remove the bulk libraries when we generate the aggregate values. This might be more involved than just filtering out bulk libraries in `Project.py:357` since we want to keep the changes in the samples table.

I think the best approach would just be to filter out bulk seq_unit in the loop at `Project.py:480`

davidsmejia:
Before merging this we should decide if we can address bulk only samples having `single-cell` in modalities here or open a new issue to address that since it is not a side effect of this change.

nozomione:
> Before merging this we should decide if we can address bulk only samples having `single-cell` in modalities here or open a new issue to address that since it is not a side effect of this change.

Per our discussion, we included the fix here since it was a small FE update (I just added a simple check), and also updated the `Project::update_project_aggregate_properties` method and the corresponding test.

This PR is now ready for your review. Thank you, David!

api/scpca_portal/test/expected_values/project_SCPCP999990.py
```diff
@@ -41,8 +41,8 @@ class Project_SCPCP999990:
         "s3_input_bucket": settings.AWS_S3_INPUT_BUCKET_NAME,
         "sample_count": 4,
         "scpca_id": SCPCA_ID,
-        "seq_units": "cell, spot",
-        "technologies": "10Xv3, visium",
+        "seq_units": "bulk, cell, spot",
+        "technologies": "10Xv3, paired_end, visium",
```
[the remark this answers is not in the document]
nozomione: 🗒️  Based on the team's discussion in Slack:

We'll filter out bulk libraries from the `summaries` in the Project Details but retain them in the Sample Details view. Additionally, update the FE to display the` Single-cell` string in the `Modalities` column for each row only if the sample includes single-cell data.

client/src/components/ProjectSamplesTable.js
```diff
@@ -172,7 +172,8 @@ export const ProjectSamplesTable = ({
     { Header: 'Technology', accessor: 'technologies' },
     {
       Header: 'Modalities',
-      accessor: ({ modalities }) => ['Single-cell', ...modalities].join(', ')
+      accessor: ({ has_single_cell_data: hasSingleCellData, modalities }) =>
+        [hasSingleCellData && 'Single-cell', ...modalities].join(', ')
```
davidsmejia: Do you think we could iterate over modalities and call getReadable? You might need to make an entry for `BULK_RNA_SEQ`

api/scpca_portal/models/project.py
```diff
@@ -433,16 +450,28 @@ def update_project_aggregate_properties(self):
         summaries_counts = Counter()
         technologies = set()
 
+        # Get the seq_unit and technology values of bulk libraries to exclude them later
+        bulk_libraries = Library.objects.filter(modality=Modalities.BULK_RNA_SEQ)
+        bulk_seq_units = {library.metadata.get("seq_unit") for library in bulk_libraries}
+        bulk_technologies = {library.metadata.get("technology") for library in bulk_libraries}
```
davidsmejia, line 456:
```suggestion
        # Get the seq_unit and technology values of bulk libraries to exclude them later
        bulk_libraries = Library.objects.filter(modality=Modalities.BULK_RNA_SEQ)
        bulk_seq_units = set(bulk_libraries.values_list('metadata__seq_unit', flat=True))
        bulk_technologies = set(bulk_libraries.values_list('metadata__technology', flat=True))
```

This should work as well as if these were not defined for a bulk library we would have a problem.

api/scpca_portal/models/project.py
```diff
@@ -484,6 +513,8 @@ def update_project_aggregate_properties(self):
             project_summary.sample_count = count
             project_summary.save(update_fields=("sample_count",))
 
+        return self.summaries.all()
```
davidsmejia, line 516:
```suggestion
```
Was this just added for testing? If we don't use it we should remove.

davidsmejia (approved):
LGTM some minor feedback to look at before merging though.

api/scpca_portal/models/project.py
nozomione: Yes, it was added for testing - now removed 👍

client/src/components/ProjectSamplesTable.js
nozomione: It looks the values are already returned in a human-readable format from the API. We've prefixed the `"Single-cell" as a default value since it's not included in the API response.

e.g. ) https://api.scpca.alexslemonade.org/v1/samples/SCPCS000050

```js
"modalities": [
        "Bulk RNA-seq",
        "CITE-seq"
    ],
```

Alternatively, I can add the `Single Cell` to the `modalities` property in the `Sample` model so it's rendered automatically without needing to manually insert the value.
```py

    @property
    def modalities(self):
        attr_name_modality_mapping = {
            "has_single_cell_data": Modalities.SINGLE_CELL, # Add here
            "has_bulk_rna_seq": Modalities.BULK_RNA_SEQ,
            "has_cite_seq_data": Modalities.CITE_SEQ,
            "has_multiplexed_data": Modalities.MULTIPLEXED,
            "has_spatial_data": Modalities.SPATIAL,
        }

        return sorted(
            [
                modality_name.label
                for attr_name, modality_name in attr_name_modality_mapping.items()
                if getattr(self, attr_name)
            ]
        )
```

Let me know which you prefer. Thank you, David!

api/scpca_portal/models/project.py
```diff
@@ -433,16 +450,28 @@ def update_project_aggregate_properties(self):
...
                 organisms.add(sample.additional_metadata["organism"])
 
-            sample_seq_units = sample.seq_units.split(", ")
-            sample_technologies = sample.technologies.split(", ")
+            # Exclude bulk values from the sample's metadata
+            sample_seq_units = set(sample.seq_units.split(", ")) - bulk_seq_units
+            sample_technologies = set(sample.technologies.split(", ")) - bulk_technologies
+
+            if not sample_seq_units or not sample_technologies:
+                continue
```
davidsmejia, line 472: we can remove this

client/src/components/ProjectSamplesTable.js
nozomione: 🗒️ We'll refactor the BE to convert these human-readable attributes into tokens and return a list of tokens to the FE. On the FE, we'll use `getReadable` to convert the tokens back into a human-readable format, which we'll handle in separate issues.

For the time being, we'll leave the implementation as it is.

avrohomgottlieb (approved):
This is an awesome PR! Everything looks great.

The one comment that I would make is that now that we're treating the bulk tsv files as input metadata files, you should add `".tsv"` to the `METADATA_EXTENTIONS` set in common.py. As well, you should check that all of the tests pass for `OriginalFile`, as this change will activate the `is_metadata=True` attribute in `OriginalFile` for bulk tsv files.  Bulk tsv files were actually classified as `is_metadata=False` (this happens in `InputBucketS3KeyInfo`) due to the `.tsv` extension not being included inside of `METADATA_EXTENSIONS`, which we only used for input metadata files.

nozomione:
> This is an awesome PR! Everything looks great.
> 
> The one comment that I would make is that now that we're treating the bulk tsv files as input metadata files, you should add `".tsv"` to the `METADATA_EXTENTIONS` set in common.py. As well, you should check that all of the tests pass for `OriginalFile`, as this change will activate the `is_metadata=True` attribute in `OriginalFile` for bulk tsv files. Bulk tsv files were actually classified as `is_metadata=False` (this happens in `InputBucketS3KeyInfo`) due to the `.tsv` extension not being included inside of `METADATA_EXTENSIONS`, which we only used for input metadata files.

Thank you for your insight, Avrohom! I’ve added the `.tsv` file extension to `METADATA_EXTENSIONS` as you suggested 👍 Additionally, I’ve excluded the bulk files from the `InputBucketS3KeyInfo._is_metadata` flag, as they are treated as input metadata files but still need to be included in downloads - this ensures that all tests pass.

Issue #1237, opened 2025-05-05
Consolidate Dataset serializers


avrohomgottlieb:
### Context

At this point, there are 4 Dataset serializers - DatasetSerializer, DatasetDetailSerializer, DatasetCreateSerializer, and DatasetUpdateSerializer.

### Problem or idea

Currently, all fields are manually added to DatasetSerializer, DatasetCreateSerializer and DatasetUpdateSerializer. Both DatasetCreate and DatasetUpdate should inherit from DatasetSerializer, and both Create and Update should overwrite the `read_only_fields` attribute with the unique combination that each need.

### Solution or next step
- Create should inherit from DatasetSerializer, and overwrite read_only fields to include all but format, data, email and start
- Update should inherit from DatasetSerializer, and overwrite read_only fields to include all but data, email and start

PR #1276, opened 2025-05-22
1237 - Consolidate Dataset Serializers

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1237.

This PR consolidates `serializers/dataset.py` by creating inheritance relationships between the different custom serializers.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/serializers/dataset.py
+++ b/api/scpca_portal/serializers/dataset.py
@@ -69,64 +69,15 @@ class DatasetDetailSerializer(DatasetSerializer):
     computed_file = ComputedFileSerializer(read_only=True, many=False)
 
 
-class DatasetCreateSerializer(serializers.ModelSerializer):
-    class Meta:
-        model = Dataset
-        fields = (
-            "data",
-            "email",
-            "format",
-            "start",
-            "id",
-            "stats",
-            "data_hash",
-            "metadata_hash",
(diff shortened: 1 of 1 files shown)


api/scpca_portal/serializers/dataset.py
```diff
@@ -129,28 +63,23 @@ class Meta:
             "terminated_reason",
             "computed_file",
         )
```
nozomione, line 65: Just curious: If the `fields` and `read_only_fields` are almost identical, would it be possible to reference `fields` in `read_only_fields`?

e.g. )
```py
read_only_fields = fields
```


 This might improve readability, but I understand there could be a reason to keep them separate!

nozomione (approved):
I've left a small comment for curiosity regarding the assignment of `fields` and `read_only_fields`. Otherwise, this PR looks great - LGTM 👍

davidsmejia (approved):

api/scpca_portal/serializers/dataset.py
avrohomgottlieb: Sounds good. As long we're aware that it's passed by reference and that shouldn't be a problem.

Issue #1306, opened 2025-06-16
Create initial lock file query support


avrohomgottlieb:
### Context

In order to determine when input projects have been modified and their metadata needs to be reloaded, a lock file named `projects.lock` has been added to the input bucket. This file enables us to know which projects are inaccessible when building datasets, either to send an error to the requester (be it a custom dataset made by a user or a ccdl dataset generated internally), or to queue the job for processing after its project's metadata has been reloaded.

### Problem or idea

Add a mechanism for parsing the `projects.lock` lockfile, and making its contents available for datasets to query who wish to generate computed files.

### Solution or next step
- Add a `lockfile` module
- Add `check_file_empty` function to `s3` module which uses the `s3api head-object` subcommand and checks to see if `ContentLength` is 0
- Add `get_is_lockfile_empty` function to `lockfile` module, which calls `s3::is_file_empty` on the `projects.lock` lockfile
- Add `tainted` BooleanField attribute to `Project` model, which is set to `True` if a project is present in the lockfile
- Add `taint_projects` function to the `lockfile` module, which calls `get_is_lockfile_empty`, and if it's not empty, downloads the lockfile and sets the `tainted` attributes on all the corresponding projects
- Add tests which check that
  - The lockfile empty check works as anticipated
  - The lockfile is being parsed correctly, and the correct projects are being tainted

PR #1316, opened 2025-06-18
1306 - Create lockfile query support

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1306.

This PR adds a `lockfile` module, and adds support for locking projects in the `sync_original_files` management command. As well, locked projects are loadable in the `load_metadata` command by passing a `--reload-locked` flag.

## Types of changes

## Functional tests

- `test_load_metadata::test_reload_locked`
- `test/models/test_project::test_lock_projects`
- `test_lockfile::test_get_lockfile_project_ids`
- `test_s3::test_check_mocked_output`
- `test_s3::test_check_test_inputs`

## Checklist

## Screenshots

N/A


api/scpca_portal/test/management/commands/test_load_metadata.py
```diff
@@ -111,17 +114,42 @@ def test_reload_existing(self):
...
+            self.submitter_whitelist,
+            self.input_bucket_name,
+            self.reload_existing,
+            self.reload_locked,
+            self.update_s3,
+        )
 
+        reload_locked = True
+        reload_existing = True
+        self.load_metadata(reload_locked=reload_locked, reload_existing=reload_existing)
```
nozomione, line 135: I've noticed that `load_metadata` calls `lockfile.get_lockfile_project_ids` for filtering, which then calls `s3.download_files` that uses `subprocess`. This causes `subprocess.CalledProcessError` when running this test.

To avoid this, should we patch `lockfile.get_lockfile_project_ids` in `setUp`?

api/scpca_portal/test/management/commands/test_load_metadata.py
```diff
@@ -12,11 +12,13 @@
 
 class TestLoadMetadata(TestCase):
     def setUp(self):
-        self.load_metadata = partial(call_command, "load_metadata")
+        with patch("scpca_portal.lockfile.get_lockfile_project_ids", return_value=[]):
```
nozomione, line 15: **Suggestion:** Instead of using the `with` context manager, we could store the patched method as follows: 
```suggestion
       get_lockfile_project_ids_patch = patch(
            "scpca_portal.lockfile.get_lockfile_project_ids", return_value=[]
        )    
      self.mock_get_lockfile_project_ids = get_lockfile_project_ids_patch.start()
  ```
  
Then, add `get_lockfile_project_ids_patch` to `self.patches` for `tearDown`?

**Review by nozomione — CHANGES_REQUESTED** [review]
Great PR, Avrohom 🚀! I've left a few suggestions - please take a look and let me know your thoughts.

File: api/scpca_portal/test/management/commands/test_load_metadata.py (hunk 1 above)
Comment (avrohomgottlieb) [reply api/scpca_portal/test/management/commands/test_load_metadata.py:135]: Good call.

File: api/scpca_portal/models/project.py [hunk 3]
```diff
@@ -77,6 +80,17 @@ def get_from_dict(cls, data: Dict):
...
+    def lock_projects(cls, locked_project_ids: List[str]) -> List[Self]:
+        locked_projects = []
+        for project_id in locked_project_ids:
+            project = cls.objects.filter(scpca_id=project_id).first()
+            project.is_locked = True
+            locked_projects.append(project)
+        cls.objects.bulk_update(locked_projects, ["is_locked"])
+
+        return locked_projects
+
```
nozomione, line 93: **Suggestion:** We could reduce query lookups by using a mapper dict if necessary: 
```suggestion
    @classmethod
    def lock_projects(cls, locked_project_ids: List[str]) -> List[Self]:
        locked_projects = []      
        project_by_id = { 
             p.scpca_id: p 
             for p in cls.objects.filter(
                  scpca_id__in=locked_project_ids
             )
        }
        
        for project_id in locked_project_ids:
            if project := project_by_id.get(project_id):
               project.is_locked = True
               locked_projects.append(project)
        cls.objects.bulk_update(locked_projects, ["is_locked"])

        return locked_projects
```

nozomione (approved):
I've left a small suggestion - otherwise this looks great! LGTM 👍

api/scpca_portal/models/project.py
avrohomgottlieb: Good thinking about reducing queries. In this case, I'm not so worried about it, because we're not expecting many project ids to be found in the lockfile at the same time. In any case, it is still best practice. Taking inspiration from your suggestion, I feel like iterating over the following would be an improvement:
```python
for project in cls.objects.filter(scpca_id__in=locked_project_ids):
```

Regarding mapper dicts - mapper dicts are a bit tricky because they can be kind of hard to debug. We typically try to stay away from ad-hoc data structures on the API if we can, preferring to create and instantiate custom classes with defined data types.

nozomione: Thanks for the insight, Avrohom! In my PR #1317, I actually used an ad-hoc mapper dict (`sample_by_id`) to look up the first object from the project samples (e.g., `Library.load_metadata`). 

Here's what I implemented:
```py
sample_by_id = {sample.scpca_id: sample for sample in project.samples.all()}
# Within the loop
   if sample := sample_by_id.get(sample_id):
```

This replaced the original approach, which queried the samples on each iteration:
```py
if sample := self.samples.filter(scpca_id=sample_id).first():
```

Do you think I should revert to the original version? What do you think?

davidsmejia: @nozomione your implementation on 1317 is fine as is. In that context we iterate over every sample and library necessarily so fetching them at once in batches is more efficient.

While it is true we generally avoid adhoc data structures. Creating a "map" or dict is really no different then just a named array in that use case. The type of adhoc data structures we try to avoid are things that would be better served as a class or logic that should be extracted into another module.

api/scpca_portal/s3.py
```diff
@@ -32,12 +32,29 @@
...
-def remove_listed_directories(listed_objects):
+def _exclude_objects_with_key_substrings(
+    bucket_objects: List[Dict], substrings: List[str]
+) -> List[Dict]:
+    """
+    Return filtered version of passed bucket objects,
+    removing all objects whose s3 keys include any of the passed substrings.
+    """
+    if not substrings:
+        return bucket_objects
```
davidsmejia, line 43: can we move this check to the caller

Issue #1282, opened 2025-05-30
Implement 'Download Ready' Email Template Header, Footer and Link


avrohomgottlieb:
### Context
The design for the email template can be found [here](https://www.figma.com/design/AqRgSXil3E9bWihG2k1U3l/ScPCA-master?node-id=1521-20279&p=f&t=GzcQUFZCA4RRvBc6-0).

### Problem or idea
In this issue we should begin constructing the email template. In a different issue, we'll handle the Dataset Summary and Download File Summary tables.

### Solution or next step
- Add header
- Add footer
- Add link

PR #1318, opened 2025-06-26
1282 - Start implementation of dataset_ready email template

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1282.

In this PR, a `dataset_ready` email template was added. Included in this template thusfar are the header, the link button, and the footer. The main content, which includes a Dataset Summary table and a Download File Summary table, will be added in a future PR.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/email_templates/dataset_ready.html
+++ b/api/scpca_portal/email_templates/dataset_ready.html
@@ -1,0 +1,140 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Your dataset is ready!</title>
+    <style>
+      body {
+        font-family: sans-serif;
+        background-color: #f3f3f3;
+        margin: 0;
+        padding: 0;
+      }
+
+      @media only screen {
+        html {
+          min-height: 100%;
+          background: #f3f3f3;
+        }
+      }
+
+      table {
+        border-collapse: collapse;
+        width: 100%;
+      }
+
+      .container {
+        max-width: 600px;
+        margin: 0 auto;
+        background-color: #ffffff;
+        border-radius: 4px;
+        overflow: hidden;
+      }
+
+      .banner {
+        background-color: #145ECC;
+        color: #ffffff;
+        padding: 20px;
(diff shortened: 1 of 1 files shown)


api/scpca_portal/email_templates/dataset_ready.html
```diff
@@ -0,0 +1,140 @@
...
+          <h2>Your dataset is ready for download!</h2>
+          <p>Uncompressed Size: {{ uncompressed_size }} GB</p>
+          <a href="{{ download_link }}" class="button">Get Dataset</a>
+        </td>
+      </tr>
+
+      <!-- Divider line -->
+      <tr>
+        <td>
+          <div class="divider"></div>
```
nozomione, line 116: [Can I email](https://www.caniemail.com/features/html-div/) recommends using `<td>` instead of `<div>` if we'll need to support Outlooks (2007-2019) on Windows.

api/scpca_portal/email_templates/dataset_ready.html
```diff
@@ -0,0 +1,140 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Your dataset is ready!</title>
+    <style>
```
nozomione, line 6: According to [Can I email](https://www.caniemail.com/features/html-style/), some mail clients (e.g., Gmail Mobile) remove the first `<head>`, so it's recommended to place `<style>` inside a second `<head>`. To be safe, using inline CSS is the safest approach.

nozomione (approved):
Nice work, Avrohom - I left a few comments but I approve this PR  👍

**Suggestion:** To ensure styles are supported across all mail clients, we could leverage a package like [django-inlinecss](https://pypi.org/project/django-inlinecss/). It automatically converts external style rules into inline CSS and is fairly easy to use. It also seems to be a well-used and popular choice for email templates.

davidsmejia (approved):
LGTM

PR #1334, opened 2025-07-01
1320 - Move the loader methods to the appropriate files

### LINKED ISSUE #1320: Move the loader methods to the appropriate files
### Context

The current `loader` module handles several distinct tasks, including `data` directory preparation, metadata loading, project creation, and computed file generation. As we transition computed file generation from the `Project`/`Sample` level to the `Dataset` level, most of the methods in this module will become obsolete and eventually be removed.

This transitioning phase provide a good opportunity to reorganize the module by splitting it into smaller, more focused components. This refactoring will improve code maintainability, scalability, and testability.

### Problem or idea

The following methods should be moved from the `loader` module to `utils` since they are used across the codebase:
- `prep_data_dirs`
- `remove_project_input_files`

The remaining methods will stay in the `loader` module. For those planned for removal after the dataset release (marked as "slot to remove"), we'll add `TODO` comments for easy lookup later.

Remaining `loader` methods:
- `get_projects_metadata`
- `create_project `
- `_can_process_project` (slot to remove)
- `_can_purge_project` (slot to remove)
- `_create_computed_file` (slot to remove)
- `_create_computed_file_callback` (slot to remove)
- `generate_computed_file` (slot to remove)
- `generate_computed_files` (slot to remove)

### Solution or next step

- Move `prep_data_dirs` and `remove_project_input_files`methods from `loader` to `utils`
- Add a `TODO` comment (`# TODO: Remove after dataset release`) to methods marked as "slot to remove"
  - Define a temporary constant for this `TODO` comment in `common`, so it can be referenced consistently across the codebase and related issues.
- Update all references and the corresponding tests to reflect these changes

nozomione:
## Issue Number

Closes #1320

I've broken up the loader module into appropriate files and updated the corresponding tests to reflect these changes.

## Types of changes

## Functional tests

- Updated all references in the corresponding tests to reflect the changes

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/common.py
+++ b/api/scpca_portal/common.py
@@ -1,5 +1,7 @@
 from scpca_portal.enums import JobStates, Modalities
 
+TODO_AFTER_DATASET_RELEASE = "Remove after the dataset release"
+
 CSV_MULTI_VALUE_DELIMITER = ";"
 
 TAB = "\t"
--- a/api/scpca_portal/common.py
+++ b/api/scpca_portal/common.py
@@ -1,7 +1,5 @@
 from scpca_portal.enums import JobStates, Modalities
 
-TODO_AFTER_DATASET_RELEASE = "Remove after the dataset release"
-
 CSV_MULTI_VALUE_DELIMITER = ";"
 
 TAB = "\t"
--- a/api/scpca_portal/utils/helpers.py
+++ b/api/scpca_portal/utils/helpers.py
@@ -2,15 +2,44 @@
 
 import inspect
 import math
+import shutil
 from collections import namedtuple
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Callable, Dict, Generator, Iterable, List, Set, Tuple
 
+from django.conf import settings
+
 from scpca_portal import common
 from scpca_portal.config.logging import get_and_configure_logger
 
 logger = get_and_configure_logger(__name__)
+
+
+def prep_data_dirs(wipe_input_dir: bool = False, wipe_output_dir: bool = True) -> None:
+    """
+    Create the input and output data dirs, if they do not yet exist.
+    Allow for options to be passed to wipe these dirs if they do exist.
+        - wipe_input_dir defaults to False because we typically want to keep input data files
+        between testing rounds to speed up our tests.
+        - wipe_output_dir defaults to True because we typically don't want to keep around
+        computed files after execution.
+    The options are given to the caller for to customize behavior for different use cases.
+    """
+    # Prepare data input directory.
+    if wipe_input_dir:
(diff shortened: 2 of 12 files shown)


api/scpca_portal/common.py
```diff
@@ -1,5 +1,7 @@
 from scpca_portal.enums import JobStates, Modalities
 
+TODO_AFTER_DATASET_RELEASE = "Remove after the dataset release"
```
davidsmejia, line 3: Interesting idea but let's just put this inline. The point is that we will be deleting it anyway.

api/scpca_portal/utils/helpers.py
```diff
@@ -2,17 +2,46 @@
...
+
+    # Prepare data output directory.
+    if wipe_output_dir:
+        shutil.rmtree(settings.OUTPUT_DATA_PATH, ignore_errors=True)
+    settings.OUTPUT_DATA_PATH.mkdir(exist_ok=True, parents=True)
+
+
+def remove_project_input_files(project_id: str) -> None:
+    """Remove the input files located at the project_id's input directory."""
+    shutil.rmtree(settings.INPUT_DATA_PATH / project_id, ignore_errors=True)
```
davidsmejia, line 42: Since this is in `utils` now I think we should make the following changes:
- rename to `remove_input_dir(dir)`
- raise exception if no dir is passed. Since we shouldn't be calling this method to wipe the entire input directory.

api/scpca_portal/common.py
nozomione: Sure!

api/scpca_portal/utils/helpers.py
avrohomgottlieb: I am also inclined towards this. The simplest move would probably be to break this up, to add a static method to the Project model called `get_input_dir`, and then pass the returned Path object to `remove_input_dir(dir)`, which would do the `shutil.rmtree` operation.

```python
def get_input_dir(project_id) -> Path:
  return settings.INPUT_DATA_PATH / project_id
```

If you would add this, it would be good to remove the current property method `input_data_path`, which would no longer be necessary. The reason that method is less valuable than the one being proposed is because it's dependent on a project object already having been instantiated, something which wouldn't help us here.

avrohomgottlieb (commented):
All looks good with this PR, just had a comment about `remove_project_input_files`

api/scpca_portal/utils/helpers.py
avrohomgottlieb: I see that this was resolved in PR https://github.com/AlexsLemonade/scpca-portal/pull/1335 👍 .

davidsmejia (approved):
LGTM

avrohomgottlieb (approved):
LGTM

PR #1340, opened 2025-07-02
1329 - Defer processing for locked jobs

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1329.

## Purpose/Implementation Notes

Describe the big picture of your changes.

## Types of changes

What types of changes does your code introduce?

- Refactor (addresses code organization and design mentioned in corresponding issue)
- New feature (non-breaking change which adds functionality)

## Functional tests

List out the functional tests you've completed to verify your changes work locally.

## Checklist

- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works

## Screenshots

N/A

PR #1341, opened 2025-07-03
1323 - Use Job.apply_state in Job bulk class methods for bulk-level state sync

### LINKED ISSUE #1323: Use Job.apply_state in Job bulk class methods for bulk-level state sync
### Context

During the dataset job processing lifecycle, we support two types of state propagation from `Job` to `Dataset`:

- Bulk-level synchronization
- Instance-level synchronization

In #1322, we introduced the `Job.apply_state` instance method, which allows dynamic state mapping. We also refactored the `Job` and `Dataset` models to support instance-level synchronization using this new method.

### Problem or idea

This issue focuses on refactoring **Bulk-level synchronization** logic using `apply_state`. 

All existing `Job` class methods related to bulk state propagation will be updated to use `apply_state`. Additionally, obsolete state-specific methods (e.g., `on_job_<STATE>` event handlers) will be removed, as dynamic state mapping now replaces them.

e.g. ) `Job.terminate_processing` using `apply_state`:

```py
# Job model

@classmethod
 def terminate_processing(cls, reason: str | None = "Terminated processing jobs"):
  """
  Terminates all processing, incomplete jobs on AWS Batch.
  Updates each job's state and terminated_at with the given terminated reason.
  Returns all the terminated jobs.
  """
  terminated_jobs = []
  terminated_datasets = []

  jobs = cls.objects.filter(state=JobStates.PROCESSING)

  if not jobs:
    return []

  for job in jobs:
    if batch.terminate_job(job):
      if changed := job.apply_state(JobStates.TERMINATED, reason)
        terminated_jobs.append(job)
        # TODO: remove after dataset release
        if job.dataset:
          terminated_datasets.append(job.dataset)

    cls.objects.bulk_update(terminated_jobs, STATE_UPDATE_ATTRS)

    # If any updated objects
    if terminated_datasets:
      Dataset.bulk_update_state(job_datasets)

  return terminated_jobs
```

> [!note]
> The `if self.dataset` check is temporary and will be removed once the dataset feature is fully released.

### Solution or next step

- In `Job`:
  - Update all relevant `Job` class methods (e.g., `submit_pending`, `terminate_processing`) to use `Job.apply_state` 
  - Define a new constant `STATE_UPDATE_ATTRS` for bulk update attributes 
  - Add a `TODO` comment (`# TODO: remove after dataset release`) for temporary dataset checks

- In `Dataset`:
  - Rename `update_from_last_job` to `bulk_update_state` for clarity 
  - Remove the `update_from_last_job` instance method and all `on_job_<STATE>` event handlers

nozomione:
## Issue Number

Closes #1323 

Changes include:
- Refactored `Job` class method for bulk operations to use `apply_state`
- Adjusted `bulk_update_state` for both `Job` and `Dataset`
- Removed all obsoleted methods from the `Dataset` model
- Updated the corresponding tests accordingly

## Types of changes

## Functional tests

- Adjusted the existing tests to reflect the changes

## Checklist

## Screenshots

N/A


api/scpca_portal/models/job.py
```diff
@@ -174,117 +185,126 @@ def get_sample_job(
...
+            return []
+
         retry_jobs = []
+        retry_datasets = []
 
         for job in jobs:
-            retry_jobs.append(job.get_retry_job(save=False))
+            if retry_job := job.get_retry_job(save=False):
+                retry_jobs.append(retry_job)
+                if job.dataset:  # TODO: common.TODO_AFTER_DATASET_RELEASE
```
davidsmejia, line 201: can you update these comments again to match the others from earlier

api/scpca_portal/models/job.py
```diff
@@ -174,70 +174,88 @@ def get_sample_job(
     @classmethod
     def create_retry_jobs(cls, jobs: List[Self]) -> List[Self]:
         """
-        Creates new jobs to retry the given jobs.
+        Creates new PENDING jobs to retry the given jobs.
+        Calls the datasets' method to sync the jobs' state.
         Returns the newly created retry jobs.
         """
+        if jobs is None:
+            return []
```
avrohomgottlieb, line 182:
```suggestion
        if jobs is None:
            return []
```
```suggestion
        if not jobs:
            return []
```

avrohomgottlieb, line 182:
```suggestion
        if not jobs:
            return []
```

avrohomgottlieb (approved):
LGTM. One small comment, but all looks great.

nozomione:
Hey @davidsmejia & @avrohomgottlieb - regarding the behavior of `Job.submit_pending` and `Job.submit`:

PR #1340 added a `try/except` block to `submit_pending` (for bulk operation), which now calls the `submit` instance method directly (previously `batch.submit_job(job)`). Since `submit` saves both the job and its associated dataset, `submit_pending` no longer performs bulk saves.

To restore the intended bulk-saving behavior,  what do you think about adding a `save` kwarg to `submit` (default to `True`) ?  This would allow `submit_pending` to skip individual saves and instead perform a single bulk update for jobs and datasets after the `for` loop.

Also, since we introduced a similar `save` kwarg pattern that led to renaming of `get_retry_job` to `create_retry_job` for clarity, do you think it might make sense to revisit the name `submit` as well?

Let me know your thoughts. Thank you!

davidsmejia:
> Hey @davidsmejia & @avrohomgottlieb - regarding the behavior of `Job.submit_pending` and `Job.submit`:
> 
> PR #1340 added a `try/except` block to `submit_pending` (for bulk operation), which now calls the `submit` instance method directly (previously `batch.submit_job(job)`). Since `submit` saves both the job and its associated dataset, `submit_pending` no longer performs bulk saves.
> 
> To restore the intended bulk-saving behavior, what do you think about adding a `save` kwarg to `submit` (default to `True`) ? This would allow `submit_pending` to skip individual saves and instead perform a single bulk update for jobs and datasets after the `for` loop.
> 
> Also, since we introduced a similar `save` kwarg pattern that led to renaming of `get_retry_job` to `create_retry_job` for clarity, do you think it might make sense to revisit the name `submit` as well?
> 
> Let me know your thoughts. Thank you!

For now lets just resolve conflicts and then handle the bulk saving action later. Too many PRs are touching the same area so let's fine tune this later on.

PR #1342, opened 2025-07-03
1324 - Use Job.apply_state in process_dataset management command

### LINKED ISSUE #1324: Use Job.apply_state in process_dataset management command
### Context

In #1322, we introduced a new `Job.apply_state` instance method for dynamic state mapping and refactored the `Job` and `Dataset` models to support instance-level synchronization utilizing this new method. As part of that refactor, several methods related state sync were no longer used and will be removed as they became obsolete.

These changes should be applied throughout the codebase to ensure consistency and keep everything up to date.

### Problem or idea

The current implementation of the `process_dataset` management sets the `Job` instance state directly and calls the will-be-removed  `Job.update_state_at` method. This should be updated to use `Job.apply_state` instead.

```py
#  The process_dataset management command
...

try:
    job.process_dataset_job()
    job.state = JobStates.SUCCEEDED # Directly set state
except Exception:
    job.state = JobStates.FAILED # Directory set state

job.update_state_at() # Calls the obsolete method
job.save()
```
**NOTE:** The `changed` check can be omitted when saving objects.

### Solution or next step

- Refactor the `process_dataset` management command to use `Job.apply_state` within the `try/except` block
- Clean up obsolete methods from the codebase

nozomione:
## Issue Number

Closes #1324 
Base branch PR: #1323 (which includes the necessary changes)

Changes include:
- Refactored the `process_dataset` management command to use `Job.apply_state`
- Removed the obsolete `Job.update_state_at` method 

## Types of changes

## Functional tests

N/A

## Checklist

- [c] Lint and unit tests pass locally with my changes

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/management/commands/process_dataset.py
+++ b/api/scpca_portal/management/commands/process_dataset.py
@@ -46,12 +46,12 @@ class Command(BaseCommand):
 
         try:
             job.process_dataset_job()
-            job.state = JobStates.SUCCEEDED
+            job.apply_state(JobStates.SUCCEEDED)
         except Exception:
-            job.state = JobStates.FAILED
-
-        job.update_state_at()
-        job.save()
-
-        if job.dataset.email:
-            notifications.send_dataset_file_completed_email(job)
+            job.apply_state(JobStates.FAILED)
+        finally:
+            job.save()
+            if job.dataset:  # TODO: common.TODO_AFTER_DATASET_RELEASE
+                job.dataset.save()
+                if job.dataset.email:
+                    notifications.send_dataset_file_completed_email(job)


api/scpca_portal/management/commands/process_dataset.py
```diff
@@ -46,12 +46,12 @@ def process_dataset(self, job_id: str, **kwargs) -> None:
...
-
-        job.update_state_at()
-        job.save()
-
-        if job.dataset.email:
-            notifications.send_dataset_file_completed_email(job)
+            job.apply_state(JobStates.FAILED)
+        finally:
+            job.save()
+            if job.dataset:  # TODO: common.TODO_AFTER_DATASET_RELEASE
```
davidsmejia, line 54: Can we update this comment

davidsmejia (approved):
LGTM after the comment

api/scpca_portal/management/commands/process_dataset.py
```diff
@@ -46,12 +46,12 @@ def process_dataset(self, job_id: str, **kwargs) -> None:
...
-
-        if job.dataset.email:
-            notifications.send_dataset_file_completed_email(job)
+            job.apply_state(JobStates.FAILED)
+        finally:
+            job.save()
+            if job.dataset:  # TODO: Remove after the dataset release
+                job.dataset.save()
+                if job.dataset.email:
+                    notifications.send_dataset_file_completed_email(job)
```
avrohomgottlieb, line 57: With this implementation, because we're catching all errors (as all errors inherit from `Exception`), there's no need to use `finally`. In the PR https://github.com/AlexsLemonade/scpca-portal/pull/1348, where this code is amended to only catch `DatasetError`, `finally` would be appropriate there, to handle the case where we wouldn't catch an error and would want to run the saving code before the exception would be raised.

avrohomgottlieb (approved):
Looks good to me. If you're able to merge this into dev before PR https://github.com/AlexsLemonade/scpca-portal/pull/1348 goes in, I think that would be preferable.

Issue #1328, opened 2025-06-30
Requeue locked dataset jobs during processing


avrohomgottlieb:
### Context

When dataset jobs end up on Batch, we have to make sure that at no point during processing were any of their projects added to the lockfile.

### Problem or idea

Add logic that determines if a job needs to be requeued at the time of processing.

### Solution or next step
- Add calls to `Dataset::has_lockfile_projects` and `Dataset::has_locked_projects` throughout `ComputedFile::get_dataset_file`
  - First check should be at the top of `get_dataset_file`
  - Further checks should be during the downloading of files: files should be downloaded by project, and after each project's files have concluded downloading, the locked project checks should be done
- For now, a generic exception should be thrown when locked projects are found
- The exception should be caught in `process_dataset`, the job's `JobState` should be set to `FAILED`, the job should be cloned with `get_retry_job` and consequently saved
- Test that the exceptions are being thrown correctly

PR #1346, opened 2025-07-07
1328 - Requeue locked jobs during processing

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1328.

In this PR, checks are made for lockfile and locked projects during processing, and should lockfile or locked projects be found, jobs are failed and cloned. The retry job then enters into the pending queue, to be dispatched at a later point.

## Types of changes

## Functional tests

- Additions and updates made to current tests to account for changes

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/management/commands/process_dataset.py
+++ b/api/scpca_portal/management/commands/process_dataset.py
@@ -47,7 +47,10 @@ class Command(BaseCommand):
         try:
             job.process_dataset_job()
             job.state = JobStates.SUCCEEDED
-        except Exception:
+        except Exception as e:
+            if str(e) == "Dataset has a locked project.":
+                retry_job = job.get_retry_job()
+                retry_job.save()
             job.state = JobStates.FAILED
 
         job.update_state_at()
--- a/api/scpca_portal/management/commands/process_dataset.py
+++ b/api/scpca_portal/management/commands/process_dataset.py
@@ -49,8 +49,7 @@ class Command(BaseCommand):
             job.state = JobStates.SUCCEEDED
         except Exception as e:
             if str(e) == "Dataset has a locked project.":
-                retry_job = job.get_retry_job()
-                retry_job.save()
+                job.get_retry_job(save=True)
             job.state = JobStates.FAILED
 
         job.update_state_at()
--- a/api/scpca_portal/management/commands/process_dataset.py
+++ b/api/scpca_portal/management/commands/process_dataset.py
@@ -49,7 +49,7 @@ class Command(BaseCommand):
             job.state = JobStates.SUCCEEDED
         except Exception as e:
             if str(e) == "Dataset has a locked project.":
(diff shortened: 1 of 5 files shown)


api/scpca_portal/management/commands/process_dataset.py
```diff
@@ -47,7 +47,10 @@ def process_dataset(self, job_id: str, **kwargs) -> None:
         try:
             job.process_dataset_job()
             job.state = JobStates.SUCCEEDED
-        except Exception:
+        except Exception as e:
+            if str(e) == "Dataset has a locked project.":
+                retry_job = job.get_retry_job()
+                retry_job.save()
```
nozomione, line 53: Once the latest `dev` updates are merged,  we can safely remove the line that saves the job, as `get_retry_job()` now saves the instance by default:

https://github.com/AlexsLemonade/scpca-portal/blob/d6b4202f87a334bd5d31d769722f5338322118bd/api/scpca_portal/models/job.py#L424

```suggestion
```

avrohomgottlieb: Thanks for bringing this up, Nozomi. @davidsmejia is the saving of an object in a `getter` function not a deviation from our conventions? Maybe we should consider renaming `get_retry_job` to `create_retry_job`? I must have missed this on https://github.com/AlexsLemonade/scpca-portal/pull/, I apologize.

To bring the point home, the above code is left with the following, which seems strange (to `get` an object but not do anything with it). 
```python
try:
    ...
except Exception as e:
    if str(e) == "Dataset has a locked project.":
        job.get_retry_job()
...
```

avrohomgottlieb: After looking at PR https://github.com/AlexsLemonade/scpca-portal/pull/1341, I'm actually a bit torn about it. 

The ambiguity lies in the fact that saving is optional:
- If it doesn't save, then it's very much a `getter`, like it's used in `create_retry_jobs`.
- If it does save, then it's a `creater`, like how it's used here. 

In any case, it should return the object, but the whole point of calling it in `create_retry_job` is that it _doesn't_ save within the method, so I'm not sure what the default behavior should be which should dictate the method's name.

nozomione: How about `get_or_create_retry_job`? @avrohomgottlieb @davidsmejia

davidsmejia: Since the default behavior is saving we should rename it to `create_retry_job` and then allow for the optional `save=False` which will allow us to more efficiently persist to the database

davidsmejia: I wouldn't rename it in this PR though, we can address it holistically after the dust settles on this sprint to prevent confusion.

PR #1367, opened 2025-07-15
1349 - Add save kwarg to Job instance methods for bulk and instance-level updates

### LINKED ISSUE #1349: Add save kwarg to Job instance methods for bulk and instance-level updates
### Context

We've optimized state propagation from `Job` to `Dataset` in PRs #1339 and #1341 by introducing `Job.apply_state`, which is now used in `Job` instance methods (e.g., `submit`). These instance methods are called by the corresponding `Job` class methods (e.g., `submit_pending`) that handle bulk operations, and thus need to support both individual and bulk-level updates. 

- For bulk-level: allow skipping individual saves for performance optimization
- For instance-level: support immediate saving

### Problem or idea

To achieve this, we should add an optional `save` kwarg (`True` by default) to the following instance methods:
- `get_retry_job` →` create_retry_job` (rename)
- `submit`
- `sync_state`
- `terminate`

### Solution or next step

- Update the instance methods listed above to accept a `save` kwarg with a default value of `True`
- Update the corresponding class methods to reflect these changes:
   - `create_retry_jobs`
   - `submit_ pending`
   - `bulk_sync_state`
   - `terminate_processing`

nozomione:
## Issue Number

Closes #1349

Changes include:

- In `Job` model:
  - Renamed `get_retry_job` to `create_retry_job`
  - Refactored instance methods and adjusted corresponding class methods for bulk operations:
    - `submit` and `submit_pending`
    - `terminate` and `terminate_processing`
    - `sync_state` and `bulk_sync_state`

- In `exceptions`:
  - Added new `JobError` subclasses: 
     -  `JobSyncNotProcessingError`
     -  `JobInvalidTerminateStateError`
     -  `JobSyncStateFailedError`
     -  `JobTerminationFailedError`
  - Added `BatchError` and its subclass `BatchGetJobsFailedError`

- In `batch`:
  - Refactor  `batch.get_jobs` to use `BatchGetJobsFailedError`

- Updated existing tests and added new one for exceptions in `TestJob`

## Types of changes

## Functional tests

Updated: 
- `TestJob.test_sync_state`
- `TestJob.test_terminate` 

Added:
- `TestJob.test_sync_state_handle_exception`
- `TestJob.test_terminate_handle_exception`

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/batch.py
+++ b/api/scpca_portal/batch.py
@@ -8,6 +8,7 @@ from botocore.client import Config
 
 from scpca_portal import utils
 from scpca_portal.config.logging import get_and_configure_logger
+from scpca_portal.exceptions import BatchGetJobsFailedError
 
 logger = get_and_configure_logger(__name__)
 aws_batch = boto3.client(
@@ -73,7 +74,8 @@ def terminate_job(job) -> bool:
 def get_jobs(batch_jobs: Iterable["Job"]) -> List[Dict] | None:  # noqa: F821
     """
     Fetch AWS Batch job(s) for the given one or more job(s) in bulk.
-    Return a list of fetched jobs on success, otherwise return None.
+    Raises BatchGetJobsFailedError on failure.
+    Return a list of fetched jobs on success.
     """
     max_limit = 100  # Limit of job IDs to send per request
     jobs = []
@@ -86,8 +88,11 @@ def get_jobs(batch_jobs: Iterable["Job"]) -> List[Dict] | None:  # noqa: F821
         except Exception as error:
             logger.exception(
                 f"Failed to bulk fetch AWS Batch job{pluralize(len(batch_job_ids))} "
-                f"for job IDs: {', '.join(batch_job_ids)} due to: \n\t{error}"
+                f"for job ID{pluralize(len(batch_job_ids))}: "
+                f"{', '.join(batch_job_ids)} due to: \n\t{error}"
             )
-            return None
+            raise BatchGetJobsFailedError(job_ids=batch_job_ids) from error
+
+    logger.info("AWS Job fetch complete.", batch_job_ids=batch_job_ids)
 
     return jobs
--- a/api/scpca_portal/batch.py
+++ b/api/scpca_portal/batch.py
@@ -81,17 +81,16 @@ def get_jobs(batch_jobs: Iterable["Job"]) -> List[Dict] | None:  # noqa: F821
     jobs = []
 
     if batch_job_ids := [job.batch_job_id for job in batch_jobs]:
-        try:
-            for chunk in utils.get_chunk_list(batch_job_ids, max_limit):
+        for chunk in utils.get_chunk_list(batch_job_ids, max_limit):
+            try:
                 response = aws_batch.describe_jobs(jobs=chunk)
                 jobs.extend(response["jobs"])
-        except Exception as error:
-            logger.exception(
-                f"Failed to bulk fetch AWS Batch job{pluralize(len(batch_job_ids))} "
-                f"for job ID{pluralize(len(batch_job_ids))}: "
-                f"{', '.join(batch_job_ids)} due to: \n\t{error}"
-            )
-            raise BatchGetJobsFailedError(job_ids=batch_job_ids) from error
+            except Exception as error:
+                logger.exception(
+                    f"Failed to bulk fetch AWS Batch job{pluralize(len(chunk))} "
+                    f"for job IDs: {', '.join(chunk)} due to: \n\t{error}"
+                )
+                raise BatchGetJobsFailedError(job_ids=batch_job_ids) from error
 
     logger.info("AWS Job fetch complete.", batch_job_ids=batch_job_ids)
 
--- a/api/scpca_portal/exceptions.py
+++ b/api/scpca_portal/exceptions.py
@@ -46,3 +46,15 @@ class JobInvalidRetryStateError(JobError):
     def __init__(self, job=None):
         message = "Jobs in final states cannot be retried."
         super().__init__(message, job)
+
+
+class JobTerminateNotProcessingError(JobError):
+    def __init__(self, job=None):
+        message = "Jobs in final states cannot be terminated."
+        super().__init__(message, job)
+
+
+class JobTerminationFailedError(JobError):
+    def __init__(self, job=None):
+        message = "Error terminating job in Batch."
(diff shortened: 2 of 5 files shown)


nozomione:
@davidsmejia  and @avrohomgottlieb

I've refactored the `submit` and `terminate` instance methods to accept a `save` optional kwarg, which allows saving of the job for bulk operations.

To align with the behavior of `submit`, I added new exception subclasses for `terminate`. I also removed return statements from both methods, since they now raise exceptions if the action (submit or terminate) fails. I've already discussed this new pattern with David.

For bulk operations, I updated `terminate_processing` to follow the same structure as `submit_pending` by adding `try`/`except` blocks and logging job counts for successful actions and exceptions (`JobTerminateNotProcessingError` and `JobError` which includes `JobTerminationFailedError`).

`Job.sync_state` and its corresponding class method `Job.bulk_sync_state` still need to be refactored. Before proceeding with that and applying the same pattern, I wanted to share this draft PR to get your input on the changes so far.

Let me know what you think. Thank you, everyone!

nozomione:
> Looks good so far, just some comments on semantics

Thank you for the feedback - applied ✅ !

nozomione:
@davidsmejia and @avrohomgottlieb 

Another quick question:

Do you think we should enforce the use of keyword arguments for optional parameters (e.g., `save`, `reason`) to improve clarity?

```py
def submit(self, *, save=True):
    ...

def terminate(self, *, reason: str | None = "Terminated processing job") -> bool:
    ...
```

This way, when these options are used, it’s immediately clear what they refer to:

```py
job.submit(save=False)
job.terminate(reason="Timeout during processing")
```

Curious to hear your thoughts - thank you, everyone!

davidsmejia:
> @davidsmejia and @avrohomgottlieb
> 
> Another quick question:
> 
> Do you think we should enforce the use of keyword arguments for optional parameters (e.g., `save`, `reason`) to improve clarity?
> 
> ```python
> def submit(self, *, save=True):
>     ...
> 
> def terminate(self, *, reason: str | None = "Terminated processing job") -> bool:
>     ...
> ```
> 
> This way, when these options are used, it’s immediately clear what they refer to:
> 
> ```python
> job.submit(save=False)
> job.terminate(reason="Timeout during processing")
> ```
> 
> Curious to hear your thoughts - thank you, everyone!

No, saving should be done by default, and the option to delay saving is the exception so I would expect it to have a sensible default of `True`

nozomione:
@davidsmejia and @avrohomgottlieb 

For the `sync_state` and `bulk_sync_state` methods, since `boto3`'s Batch client [`describe_jobs`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/batch/client/describe_jobs.html) accepts up to **100** jobs per request, I introduced a custom exception, `BatchGetJobsFailedError`, in the `get_jobs` method of our `batch` module to handle fetch failures.

Both `sync_state` and `bulk_sync_state` internally `get_jobs`, but handle `BatchGetJobsFailedError` differently. Each method still returns a boolean indicating whether any job states were updated.

Via the `sync_state` instance method:
- Raises exceptions (`JobSyncNotProcessingError`, `JobSyncStateFailedError`) if syncing fails, which is consistent with other instance methods (e.g., `submit`, `terminate`)
- Saves the job instance (does not accept an optional `save` kwarg - same as before)
- Returns a boolean indicating whether the job state were changed during sync  (same as before)

Via the  `bulk_sync_state` class method for bulk operation:
- Does not `sync_state` and directly calls `batch.get_jobs` for bulk job fetching (same as before)
- Logs synced jobs, unchanged jobs, and failed jobs, which is consistent with the logging patterns used in other bulk operation class methods
- Returns a boolean indicating whether any job states were updated during sync (same as before)

For the failed jobs, I can access the specific job IDs from the custom `BatchGetJobsFailedError` exception. I considered logging them again in `bulk_sync_state`, but since they're already logged in `batch.get_jobs`, I just logged the count instead.

Please let me know your thoughts. Thank you!

api/scpca_portal/batch.py
```diff
@@ -73,7 +74,8 @@ def terminate_job(job) -> bool:
 def get_jobs(batch_jobs: Iterable["Job"]) -> List[Dict] | None:  # noqa: F821
```
davidsmejia, line 74: Let's move the try / catch inside and pass the chunk to the logger.exception.

api/scpca_portal/exceptions.py
```diff
@@ -1,3 +1,30 @@
+from django.template.defaultfilters import pluralize
+
+
+# scpca_portal.batch
+class BatchError(Exception):
+    def __init__(self, message: str | None = None, job=None, job_ids=None):
+        default_message = "A boto3 batch client error occurred."
```
davidsmejia, line 7:
```suggestion
        default_message = "An error occurred while communicating with AWS Batch API."
```

api/scpca_portal/models/job.py
```diff
@@ -291,37 +310,51 @@ def bulk_sync_state(cls) -> bool:
         Syncs all PROCESSING jobs' states with the corresponding AWS Batch jobs.
         Saves the jobs (state, timestamp, reason).
         Calls the datasets' method to sync the jobs' state.
-        Returns a boolean indicating if the jobs and datasets were updated and saved.
+        Returns a boolean indicating if the jobs and datasets were updated during sync.
         """
         processing_jobs = cls.objects.filter(state=JobStates.PROCESSING)
         if not processing_jobs.exists():
+            logger.info("No processing jobs to sync.")
```
davidsmejia, line 317:
```suggestion
```

api/scpca_portal/models/job.py
```diff
@@ -291,37 +310,51 @@ def bulk_sync_state(cls) -> bool:
...
+            logger.info("No jobs were updated during sync.")
             return False
 
+        logger.info(f"Synced {len(synced_jobs)} jobs with AWS.")
         cls.bulk_update_state(synced_jobs)
         if synced_datasets:  # TODO: Remove after the dataset release
             Dataset.bulk_update_state(synced_datasets)
 
+        if unchanged_jobs:
+            logger.info(f"{len(unchanged_jobs)} jobs remain unchanged and are still processing.")
```
davidsmejia, line 354:
```suggestion
```

nozomione:
I've updated the logging, and this PR is ready for another look. Thank you, David!

avrohomgottlieb (approved):

PR #1368, opened 2025-07-15
Enable staging datasets with ENABLE_DATASETS envar

### LINKED ISSUE #1345: Enable User Dataset Endpoints on staging
### Context

We removed dataset endpoints from the API in production as we roll the feature out. Now we will need to open them back up on staging to allow for development and testing.

### Problem or idea

We should just add a new envar that can be set with terraform to determine if we should allow the feature to be accessed.

Add check on create/save endpoint to make sure token is associated with a CCDL email address.

### Solution or next step

- [ ] Add `IS_STAGING` envar to check to against which should be when ses domain is staging.
- [ ] Add additional restriction on the dataset endpoint to prevent opening the feature fully on staging.

davidsmejia:
## Issue Number

Closes #1345

- Changes use of `settings.PRODUCTION` to specifically use `settings.CLEAN_UP_DATA`
- Adds `settings.ENABLE_FEATURE_PREVIEW` to more generically allow for new features to be enabled on local and staging but not production.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/.github/workflows/deploy_staging_backend.yml
+++ b/.github/workflows/deploy_staging_backend.yml
@@ -40,3 +40,4 @@ jobs:
         run: cd infrastructure && python3 deploy.py -s staging -u deployer -a ccdlstaging -v $(git rev-parse HEAD) --project
         env:
           SENTRY_ENV: staging-api
+          ENABLE_DATASETS: "true"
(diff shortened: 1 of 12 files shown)


avrohomgottlieb (approved):
LGTM

PR #1373, opened 2025-07-18
Add bulk samples and libraries to dataset

### LINKED ISSUE #1366: Add bulk to Dataset::files_summary count.
### Context

While working out how the `Dataset::files_summary` should work we realized that bulk samples are not part of `Dataset::samples`. We should review how this impacts the current dataset computed file creation and informational properties to see if they can be added without any unintended side effects. Otherwise they will be added specifically to the `Dataset::files_summary` method. 

### Problem or idea

Ideally this will work the same and we can add bulk samples and libraries to the existing properties, otherwise we will add a new mechanism for grabbing bulk when we need them.

### Solution or next step

- Review Dataset::samples and Dataset::libraries usage
- One of:
  - Add bulk samples and libraries to already existing `samples` and `libraries` properties
  - add additional methods to support `Dataset::files_summary`

davidsmejia:
## Issue Number

Closes #1366 

- Adds bulk samples to Dataset.samples, which in turn adds Bulk libraries to dataset.libraries.

Also based on requirements discussion from yesterday I added in the project summaries counts since implementation was trivial. Added test to confirm the structure and counting.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/test/views/test_dataset.py
+++ b/api/scpca_portal/test/views/test_dataset.py
@@ -112,23 +112,6 @@ class DatasetsTestCase(APITestCase):
         response = self.client.delete(url)
         self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)
 
-    def test_stats_property_keys(self):
-        url = reverse("datasets-detail", args=[self.ccdl_dataset.id])
-        response = self.client.get(url)
-        stats_property = response.json().get("stats")
-        stats_property_fields = {
-            "current_data_hash",
-            "current_readme_hash",
-            "current_metadata_hash",
-            "is_hash_changed",
-            "uncompressed_size",
-            "diagnoses_summary",
-            "files_summary",
-            "project_diagnoses",
-        }
-        for field in stats_property_fields:
-            self.assertIn(field, stats_property)
-
     @patch("scpca_portal.models.Job.submit")
     def test_create_submit_job(self, mock_submit_job):
         url = reverse("datasets-list", args=[])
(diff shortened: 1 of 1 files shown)


api/scpca_portal/models/dataset.py
```diff
@@ -179,36 +179,53 @@ def files_summary(self) -> list[dict]:
...
+
+            if not library_ids:
+                continue
 
-                seen_libraries.extend(library_ids)
+            if samples_ids := (
+                dataset_samples.filter(libraries__scpca_id__in=library_ids)
+                .exclude(scpca_id__in=seen_samples)
+                .values_list("scpca_id", flat=True)
+            ):
```
nozomione, line 203: I noticed that in the sample filtering logic, we might be getting duplicate sample IDs when a sample is linked to multiple matching libraries.

For example, in the `files_summary` for `SCPCP000004`, the sample counts seem to include one extra sample:
```py
[{'samples_count': 33, 'name': 'Single-nuclei samples', 'format': '.rds'},  # should be 32?
 {'samples_count': 9, 'name': 'Single-cell samples', 'format': '.rds'}]  # should be 8?
 ```

It would be helpful to append `.distinct()` to ensure only unique samples are counted:
```suggestion

           if samples_ids := (
                dataset_samples.filter(libraries__scpca_id__in=library_ids)
                .exclude(scpca_id__in=seen_samples)
                .distinct()
                .values_list("scpca_id", flat=True)
            ):
 ```

Let me know what you think!

nozomione (approved):
Nice implementation 💯 🚀! I've left a suggestion regarding `samples_ids`. Please take a look. Otherwise, I'm approving this PR - LGTM!

PR #1401, opened 2025-08-07
attach metadata dataset to project serializer

davidsmejia:
## Issue Number

N/A

- To reduce the load time of /projects endpoint I want to test just serializing the metadata dataset when listing projects.

## Types of changes

- Adds `Project.metadata_dataset` to the project serializer.

Note: This is not a complete change, we will want to also list out the Ids of the datasets and then fully serialize them on the detail of the project or just always show the ids.

## Functional tests

N/A

## Checklist

## Screenshots

N/A


nozomione (approved):
👍

Issue #1436, opened 2025-08-25
Cache pre-defined list of dynamic properties


avrohomgottlieb:
### Context

In the Dataset model, stats and file items attributes recompute their values every time a dataset is fetched. This consistent recomputation is expensive and slow, and should be optimized.

### Problem or idea

There are certain attributes that may be cache-able that could be computed once upon object creation without being recomputed each time a dataset is fetched.

### Solution or next step
- Investigate which attributes may be computed upon creation and cached
- Implement the caching of these attributes

PR #1457, opened 2025-09-04
Cache dynamic Dataset properties

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1436.

- Convert cache-able Dataset properties to attributes, re-computed on save
- Add custom Dataset::save method
- Add migration

> [!NOTE]
> Bug fix: Hashes were not being re-computed when datasets were being reprocessed. This PR addresses that by hashes now being re-computed on every save.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


avrohomgottlieb:
It could be that every save doesn't need to have the attributes recalculated. To handle such a case, we could add a `no_update_cached` argument to the custom `save` method in order to bypass the caching and go straight to saving.

nozomione (approved):

api/scpca_portal/models/dataset.py
```diff
@@ -60,6 +61,22 @@ class Meta:
     data_hash = models.CharField(max_length=32, null=True)
     metadata_hash = models.CharField(max_length=32, null=True)
     readme_hash = models.CharField(max_length=32, null=True)
+    combined_hash = models.CharField(max_length=32, null=True)
```
davidsmejia, line 64: Let's rework combined hash to be a function that returns a tuple:

```python
data_hash, metadata_hash, readme_hash, combined_hash = get_hashes()
```

this should reduce the duplication of calculating the separate hashes as well and simplify usage

api/scpca_portal/models/dataset.py
```diff
@@ -106,6 +123,35 @@ def __str__(self):
...
+
+        # stats property attributes
+        self.estimated_size_in_bytes = self.get_estimated_size_in_bytes()
+        self.diagnoses_summary = self.get_diagnoses_summary()
+        self.files_summary = self.get_files_summary()
+        self.project_diagnoses = self.get_project_diagnoses()
+        self.project_modality_counts = self.get_project_modality_counts()
+        self.modality_count_mismatch_projects = self.get_modality_count_mismatch_projects()
+        self.project_sample_counts = self.get_project_sample_counts()
+        self.project_titles = self.get_project_titles()
```
davidsmejia, line 151: Let's move these attributes outside of stats now that they are immediately on the model.

@nozomione this will impact you so heads up

Issue #1445, opened 2025-08-26
Setup Cellbrowser


davidsmejia:
### Context

As per request we will be adding cellbrowser to the portal.

### Problem or idea

We will be serving the portal through an iframe that only loads after the user has accepted terms. 

### Solution or next step

- [x] Add bucket to infrastructure configured with secret header
- [x] Allow permissions for @allyhawkins and @jashapiro to copy files into the bucket
- [ ] Add page to client that loads cellbrowser through iframe after accepting terms
- [x] upload initial cellbrowser to new bucket and identify which files need to be preserved between uploads

PR #1489, opened 2025-09-15
remove lifecycle ignore cors_rule

davidsmejia:
## Issue Number

#1445 

This is a relic from the old approach where cors_rule was defined inside of the s3 bucket resource. Removing as it shouldn't affect the bucket in the same way with the new related cors resource.

## Types of changes

## Functional tests

N/A ran a terraform plan

## Checklist

## Screenshots

N/A


nozomione (approved):
👍

Issue #1543, opened 2025-10-21
Support project specific lockfiles


davidsmejia:
_Please label your issue with a label to specify its priority: breaking, high, medium, or low._

### Context
We want to update how the lockfile works so that we can more easily check project status more often.

Our previous implementation of the lockfile expected a persistent, singular lockfile that would contain a newline separated list of project ids that were currently locked. This worked but the drawback was that you would have to download the lockfile in order to read the contents. In order to alleviate that concern we have switched over to using one ephemeral lockfile per project. When the project's lockfile exists (ec: `/SCPCP000001.lock`) then we know that the project will be unavailable.

### Problem or idea

We should make the following changes to support this.

- Update interface for lockfile module to two methods:
  - `get_is_locked_project`
  - `get_locked_projects`

- Update s3 to support the following actions
  - check if file exists
  - list files by suffix, ideally by directory

- Update tests

### Solution or next step

- [ ] Add new s3 functionality
- [ ] Update `lockfile.py` to new methods
- [ ] Update tests and clean up dead code

PR #1554, opened 2025-10-30
Support project specific lockfiles

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1543.

- adds `list_files_by_suffix` and `check_file_exists` functions the s3 module
- adds `get_is_locked_project` and `get_locked_project_ids` functions to the lockfile module
- adds tests
- removes no longer used s3 and lockfile module functions
- cleans up tests and old references

## Types of changes

## Functional tests

- `test_lockfile::test_get_is_locked_project`
- `test_lockfile::test_get_locked_project_ids`

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/s3.py
+++ b/api/scpca_portal/s3.py
@@ -52,6 +52,9 @@ def _remove_listed_directories(listed_objects):
     return [obj for obj in listed_objects if not obj["s3_key"].endswith("/")]
 
 
+# TODO: list_bucket_objects and list_files_by_suffix should be combined.
+# See linked comment below for details on how to accomplish this:
+# https://github.com/AlexsLemonade/scpca-portal/pull/1554#issuecomment-3482924420
 def list_bucket_objects(bucket: str, *, excluded_key_substrings: List[str] = []) -> List[Dict]:
     """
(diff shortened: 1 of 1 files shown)


avrohomgottlieb:
@davidsmejia a few open questions as it pertains to this PR:
- What did you have in mind with the `lockfile::get_is_locked_project` function? The approach that we've implemented in the past has been to grab all locked project ids and then to see if the dataset has any projects contained in that list, as opposed to checking whether or not a single project is locked. We could update the `ComputedFile::get_dataset_file`, when we download files on a project-by-project basis, to check whether or not the individual project is locked by calling `lockfile::get_is_locked_project`, though currently what we do is that we check if _any_ of the projects are locked during each iteration.
- What should happen to the `is_locked` `Project` model attr at this point? Do we still want to rely on it as a cached way to check if a certain project is locked, or do we prefer to just check s3 each time? The s3 calls are definitely less expensive with the new implementation, but will still be slower than checking a field in the DB.

nozomione (approved):
Great PR 👍

PR #1583, opened 2025-11-13
Fix Dataset.get_diagnoses_summary by adding order_by to fix annotation

davidsmejia:
## Issue Number

N/A

Jackie pointed this out when looking at the portal. We were listing all 1's for number of samples in the Dataset Summary section.

This PR corrects the count by adding an order_by to allow the ORM to correctly group and count for the annotations. I also removed distinct for the samples count because the queryset should be unique against ID anyway and this should make it clearer.

## Types of changes

## Functional tests

tested on the staging server but in a terminal

## Checklist

## Screenshots

N/A


avrohomgottlieb (approved):
👍

Issue #1643, opened 2025-12-01
Ensure buttons that need loading state render correctly


avrohomgottlieb:
### Context

In continuation of Issue https://github.com/AlexsLemonade/scpca-portal/issues/1642, there may be other confirmation buttons in the dataset flow that trigger async actions but async actions are not send to them on the onClick callback.

### Problem or idea

We should look through the dataset flow and update them accordingly, as was done in https://github.com/AlexsLemonade/scpca-portal/issues/1642.

PR #1655, opened 2025-12-05
Add spinners to dataset loader buttons

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1643.

This PR adds the UI spinner element to buttons in the Dataset workflow that make requests and hold until receiving a response.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/client/src/theme/button.js
+++ b/client/src/theme/button.js
@@ -71,10 +71,10 @@ export default {
       > div {
         margin: 0 auto;
       }
-      background: ${theme.global.colors['alexs-deep-blue-tint-40']};
-      border: ${theme.global.colors['alexs-deep-blue-tint-40']} 1px solid;
+      background: ${theme.global.colors['black-tint-90']};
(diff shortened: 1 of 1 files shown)


client/src/components/DatasetProjectCard.js
```diff
@@ -74,7 +81,8 @@ export const DatasetProjectCard = ({
             danger
             label="Remove"
             alignSelf={responsive('stretch', 'start')}
-            onClick={() => removeProjectById(projectId)}
+            onClick={handleRemoveProject}
+            loading={removeProjectLoading}
```
avrohomgottlieb: While this is a positive UI improvement, I had trouble with the Grommet button, which defaults to `primary` when an icon is present (i.e. when the spinner is invoked). This has to be addressed.

Issue #1678, opened 2025-12-12
Debug Portal Wide Files Initial Download Error


avrohomgottlieb:
### Context

In PR https://github.com/AlexsLemonade/scpca-portal/pull/1638, the `useEffect` setting the `selectedDataset` was updated to use progressive filtering, letting `modality`, `format`, `includes_merged` and `excludes_multiplexed` trickle down from one to the next to enable dataset selection.

In light of these changes, there is a new bug that has emerged where the initially selected dataset on portal-wide files is not functioning properly, leading the UI to hang after a token is requested and returned to the user. When refreshed, the download button works as expected.

### Problem or idea

This issue should be debugged and fixed. It is likely due to an incorrect initial dataset being selected in the above mentioned useEffect. That should be the place to begin.

### Solution or next step
...

PR #1681, opened 2025-12-15
Fix Portal Wide Files Initial Download Bug

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1678.

After careful debugging, it was determined that the portal wide initial file bug was a result of the `downloadDataset` state variable being set a few render cycles before the `selectedDataset` was being set. This resulted in the `asyncFetch` function attempting to request a download link from the API while the selectedDataset was undefined, yielding an undesirable API response. This PR resolves this issue by coupling the setting of the `downloadDataset` state variable with `selectedDataset` for portal wide files (where `datasets.length === 1`).

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/client/src/contexts/CCDLDatasetDownloadModalContext.js
+++ b/client/src/contexts/CCDLDatasetDownloadModalContext.js
@@ -46,9 +46,6 @@ export const CCDLDatasetDownloadModalContextProvider = ({
       setIncludesMerged(null)
       setExcludeMultiplexed(null)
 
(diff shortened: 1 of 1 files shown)


davidsmejia (commented):
also, whenever datasets change, in that useEffect we should also do the same clean up as we do on closing the modal. this should run independent of dataset count

PR #1690, opened 2025-12-17
Fix the loading prop invalid data type

nozomione:
## Issue Number

N/A

Target branch: `feature/dataset-integration`

(This PR currently targets the dataset feature branch since the error is present. However, it does not affect QA testing and can be merged after the Dataset launch.)

This PR addresses the following console error message generated on `localhost` for a better development experience:

> Received `false` for a non-boolean attribute `loading`.
> If you want to write it to the DOM, pass a string instead: loading="false" or loading={value.toString()}.
> If you used to conditionally omit it with loading={condition && value}, pass loading={condition ? value : undefined} instead.
> button

**Change includes:**
- Renamed the `loading` prop defined for the shared `Button` component to `awaiting` to indicate that the API request is in process (awaiting promise resolution) and to prevent conflicts with Grommet.
- Updated the prop reference and local state name in components that specifically use this `awaiting` prop.

 *The prop name is subject to change: I didn't choose `requesting` since that name was already used for a state in token-related components. If you have any suggestions, let me know!

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

Screenshot 2025-12-17 at 1 20 12 PM

### DIFF
--- a/client/src/components/DatasetAddProjectModal.js
+++ b/client/src/components/DatasetAddProjectModal.js
@@ -33,7 +33,7 @@ export const DatasetAddProjectModal = ({
 
   // Modal toggle
   const [showing, setShowing] = useState(false)
-  const [loading, setLoading] = useState(false)
+  const [awaiting, setAwaiting] = useState(false)
 
   const [modalities, setModalities] = useState([])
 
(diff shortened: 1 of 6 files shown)


nozomione:
Closing this PR in favor of PR #1698, which removes the styles related to the `loading` prop in Grommet's `theme/button` as well as the `loading` prop from `GrommetButton` to prevent the invalid data type error.

PR #1707, opened 2026-01-02
test 1705 lockfile not locking

davidsmejia:
This is a test PR to investigate why the tests are failing on GHA while passing locally.

I have already deleted all locally cached test data.


davidsmejia:
OK everything is configured fine but, we will need to update the test to pass `--no-sign-request` or the request to check if an object exists (specifically for the public test bucket) will fail otherwise

Issue #1742, opened 2026-01-13
Restore "the merged objects are not integrated" warning


sjspielman:
I noticed the new deploy seems to be missing an important warning for merged objects download. Previously, when downloading a merged object standalone, we had a warning that the samples are not integrated (sorry for grainy screenshot - this was taken from our scpca-nf manuscript response to reviewers!)

<img width="266" height="286" alt="Image" src="https://github.com/user-attachments/assets/c47db473-e23a-4af0-a22c-9e255f830238" />

<br><br>
However, looking at the same view now, we have lost this warning:

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/ecdf7e79-17a5-4feb-8cfa-6724e8cc9d3d" />

I'm pretty sure this needs to be restored. Tagging @davidsmejia to get engineers in the loop as well as @jaclyn-taroni to weigh in on priority.

PR #1743, opened 2026-01-15
add multiplexed merged warning

davidsmejia:
## Issue Number

#1742 

Changes
- adds merged and multiplexed warnings to CCDL dataset select and started screens
- updates layout of download started screen to match designs
- remove "as format" from the portal-wide modal title
- when we list files in download modal (everywhere) and show spatial data add "as spaceranger" (lowercase hardcoded because it is different than the format dropdown or where we generally use format)
- `includes_multiplexed_files` is true on portal wide merged datasets this should be false (wont be visible in preview since it is an API change)
- Warns users that multiplexed is gone on single-cell portal-wide download.

## Types of changes

## Functional tests

N/A looked at each example locally

## Checklist

## Screenshots

Merged

---

Multiplexed


nozomione (approved):
LGTM 👍

Issue #1792, opened 2026-02-03
'Remove All' button visible on dataset processing/ready page


dvenprasad:
### Context

When a dataset contains more than one project, the `Remove All` button is still visible after the user has kicked off processing. 
Example: https://scpca.alexslemonade.org/datasets/9a0148f6-7cf4-4435-90ec-8be404e62aa3

<img width="1248" height="821" alt="Image" src="https://github.com/user-attachments/assets/969b5570-eb0f-445c-b3a1-da16f2f2f1dc" />

### Problem or idea

We should remove that button there. At this point that dataset is read-only.
Currently, the button doesn't currently do anything but it's confusing.

### Solution or next step

Tagging @davidsmejia for next steps

PR #1794, opened 2026-02-03
hide Remove All button when read only

davidsmejia:
## Issue Number

#1792

Previously we only checked if the dataset had keys when determining if we should show or hide the `Remove All` button. This PR adds the check to see if we have set the component to be read only and then also hides it in that circumstance.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


davidsmejia:
Preview: https://scpca-portal-z2ajisnyr-ccdl.vercel.app/datasets/f6d9afeb-9cc3-4d8e-91fa-6f954b2da9ff

nozomione (approved):
👍

Issue #1790, opened 2026-02-02
Upgrade to React v19


avrohomgottlieb:
### Context

After completing the NextJS v15 upgrade, now we can make our way up to NextJS v16.

### Problem or idea

In order to successfully upgrade to NextJS v16, we need to upgrade to React v19.

### Solution or next step
- Upgrade to React v19

Issue #1791, opened 2026-02-02
Upgrade to NextJS v16


avrohomgottlieb:
### Context

In Issue https://github.com/AlexsLemonade/scpca-portal/issues/1790, React was upgraded to v19.

### Problem or idea

With React upgraded to v19, we can now upgrade to NextJS 16.

### Solution or next step
- Upgrade to NextJS v16

PR #1865, opened 2026-03-05
Upgrade to React v19 and Next v16

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1790 and https://github.com/AlexsLemonade/scpca-portal/issues/1791.

This PR upgrades React from v18 to v19 and NextJS from v15 to v16.

Originally, we had planned on splitting up these two upgrades into two separate PRs, but as a result of an error loading svgs (explained below), and the need to explicitly configure webpack for nextjs 16 (as the default for nextjs 16 is turbopack), it made sense to do both upgrades in the same PR.

> [!NOTE]
> **Changes to SVG Loading**
> As explained in the [React 19 Upgrade Guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide), `defaultProps` was removed from function components, which caused errors with how SVGs were being imported. This seemed to stem from how SVGs were being loaded inside of babel, via `inline-react-svg` and the `svgo` plugins. A secondary hydration issue then occurred, with the client SVG rendering not matching what was being rendered on the server. To resolve this issue, we moved the SVG loading to webpack (configured in `next.config.js`), utilizing `@svgr/webpack` (while disabling cleanupIds and utilizing prefixIds to avoid ID collisions with cases of multiple SVGs on the same page).

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/client/src/components/Icon.js
+++ b/client/src/components/Icon.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import { Text } from 'grommet'
 import { Blank } from 'grommet-icons'
+import styled from 'styled-components'
 import Check from '../images/check.svg'
 import ChevronDown from '../images/chevron-down.svg'
 import ChevronLeft from '../images/chevron-left.svg'
@@ -29,6 +30,13 @@ import Instagram from '../images/instagram.svg'
 import Github from '../images/github.svg'
 import Copy from '../images/copy.svg'
 
+// force the exclamation point to fill black
+const WarningBlackFill = styled(Warning)`
+  > :last-child {
+    fill: black !important;
+  }
+`
+
 export const SVGs = {
   Check,
   ChevronDown,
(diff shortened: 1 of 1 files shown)


client/next.config.js
```diff
@@ -58,8 +58,50 @@ module.exports = () => {
...
         use: ['raw-loader', 'template-literal-loader']
       })
+
+      const fileLoaderRule = config.module.rules.find((r) =>
+        r.test?.test?.('.svg')
+      )
+      config.module.rules.push(
+        // keep ?url imports working
+        { ...fileLoaderRule, test: /\.svg$/i, resourceQuery: /url/ },
+        // everything else → React component
```
davidsmejia:
```suggestion
```

PR #1925, opened 2026-04-16
Add Custom Filterset to CCDL Dataset View

### LINKED ISSUE #1917: Add Custom Filterset for CCDL Datasets
### Context

In PR https://github.com/AlexsLemonade/scpca-portal/pull/1868, custom fitlersets were added to the Prooject and Sample Views.

### Problem or idea

A custom filterset should now be added to the CCDL Dataset View.

### Solution or next step
- Add custom filterset to `views/ccdl_dataset.py`
- Add test cases to verify change

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1917.

In this PR:
- A custom filterset is added to the CCDL Dataset view
- A null property check is added to build_auto_filter function to enable passing of isnull filter

## Types of changes

## Functional tests

- `test_filter::FilterTest::test_nullable_auto_fields`

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/filter.py
+++ b/api/scpca_portal/filter.py
@@ -38,8 +38,9 @@ class ArrayFieldContainsFilter(django_filters.BaseInFilter, django_filters.CharF
 def build_auto_filterset(
     model,
     auto_fields: list[str] = [],
-    extra_fields: dict[str, list[str]] = None,
-    extra_filters: dict = None,
+    extended_auto_field_lookups: dict[str, list[str]] = {},
+    extra_fields: dict[str, list[str]] = {},
+    extra_filters: dict = {},
 ):
     """
     Introspects a model and builds a FilterSet with sensible lookup expressions
@@ -48,6 +49,9 @@ def build_auto_filterset(
         model:          The Django model class to build a FilterSet for.
         auto_fields:    Optional allowlist of field names. Use this
                         to keep your public API surface intentional.
+        extended_auto_field_lookups:
+                        Optional dict of filters for specific auto fields
+                        as an extension from the per data type defaults
         extra_fields:   Additional model fields included in the public API
                         e.g. {"project__scpca_id": ["exact"]}.
         extra_filters:  Optional dict of additional filter instances to mix in,
@@ -78,6 +82,8 @@ def build_auto_filterset(
         for field_type, lookups in FILTER_LOOKUPS.items():
             if isinstance(field, field_type):
                 meta_fields[field.name] = lookups
+                if extended_lookup := extended_auto_field_lookups.get(field.name):
+                    meta_fields[field.name] += extended_lookup
                 break
 
     if extra_fields:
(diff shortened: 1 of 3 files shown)


api/scpca_portal/test/test_filter.py
```diff
@@ -27,6 +27,16 @@ def test_no_auto_fields(self):
...
 
+    def test_extended_auto_field_lookups(self):
+        SampleFilterSet = filter.build_auto_filterset(
+            Sample,
+            auto_fields=["scpca_id", "diagnosis"],  # TextFields
+            extended_auto_field_lookups={"diagnosis": ["isnull"]},
+        )
+        actual_filters = SampleFilterSet.base_filters.keys()
+        self.assertIn("diagnosis__isnull", actual_filters)
+        self.assertNotIn("scpca_id__isnull", actual_filters)
```
avrohomgottlieb, line 38: > [!WARNING]
> The `self.assertNotIn("scpca_id__isnull", actual_filters)` assertion is currently failing and needs to be investigated.

In filter.py, the `LOOKUP_FILTERS` data structure includes the following for `TextField`:
```py
FILTER_LOOKUPS = {
    ...
    models.TextField: ["exact", "icontains"],
}
```

It appears that `isnull` is being added by default in `SampleFilterSet.base_filters`, causing this test to fail. I'm not exactly sure why, because querying on staging for `isnull` on the Sample endpoint (which was recently merged into dev with a custom filterset) does not seem to work.

avrohomgottlieb: It's no longer relevant as the implementation has changed, but this was actually the result of `metadata_fields[field.name]` being set with `meta_fields[field.name] += ["isnull"]`  as a pass-by-reference. Essentially, once the list was mutated, all other references pointed to the list with the `isnull` filter.

api/scpca_portal/filter.py
```diff
@@ -77,7 +77,10 @@ def build_auto_filterset(
         # Standard field types: use dict-style meta fields for multi-lookup support
         for field_type, lookups in FILTER_LOOKUPS.items():
             if isinstance(field, field_type):
-                meta_fields[field.name] = lookups
+                # create deep copy to accommodate mutability of the list
```
davidsmejia: This does a shallow copy, which I think is fine for our use case.

davidsmejia (approved):
LGTM but the comment should be updated

api/scpca_portal/filter.py
avrohomgottlieb: Thanks for the feedback. As per the above comment, the lookup dict is passed into memory when it's declared, and if a shallow copy is made where only the reference is passed, all other attributes of the same lookup data type point to the extended list.

I verified this in `test_nullable_auto_fields`, where `scpca_id` (a non-nullable field) kept receiving the `isnull` filter. This was a result of `diagnosis` (which is a nullable field) extending the referenced list pointed by `models.TextField`, causing all attrs of type `models.TextField` to receive the `isnull` filter.

PR #1927, opened 2026-04-16
add hotfix for legacy local storage values

davidsmejia:
## Issue Number

N/A

We eagerly pushed a change that modifies the default dataset value that is saved in local storage and then make assumptions about that variables shape. This PR is just a temp hotfiix that can be removed at some point in the future.

It just enforces that the data attribute is defined on datasets. Worked locally, on load the dataset is updated in local storage and becomes compliant with the new session. We can extract this in to a migration step close to the entry point on application load in the future.

## Types of changes

## Functional tests

ran locally with legacy dataset shape and watched it update

## Checklist

## Screenshots

N/A


nozomione (approved):
Thank you for fixing this! LGTM 👍

avrohomgottlieb (approved):

PR #1946, opened 2026-05-05
Establish Automatic Cert Refresh

### LINKED ISSUE #1243: Automatic Cert Refresh
### Context

We currently cache our cert in a bucket to prevent being rate limited in situations where we need to do multiple production deploys in quick succession. However the API cert expired without being refreshed because it had been a good while since the last production deploy which cycles the cert.

### Problem or idea

We should update the API certificate configuration to automatically request a new cert when the current one is near expiration. This should be possible by updating the block inside of the API's user data script that handles making the request.

### Solution or next step

- [ ] setup automatic cert refresh

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1243.

- Updates the user data script to conditionally add a cron entry which looks to renew the ssl cert each day at midnight
- Adds a deploy hook for the renew command which is responsible for updating s3 and reloading nginx 
- Adds a cloudwatch log stream for the called renew command to monitor output
- Refactors pieces of the user-data script to support this new functionality
- Abstracts s3 syncing and cleanup logic to sub script called by cert creation and renewal flows

## Types of changes

## Functional tests

N/A

## Checklist

- [x] I have updated any associated Storybook stories (if applicable)

## Screenshots

N/A


avrohomgottlieb:
Once this PR is merged in, we're going to open up a new ticket to verify that everything functions as expected.

As will be detailed in that issue, we want to check three different conditions:
- That there's been no regression in using the existing cert on deploy (making sure it still pulls from s3)
- That there's been no regression on creating a new cert on deploy (making sure it correctly provisions a new cert if there is none in s3)
- That the renewal process (via the command called by the cron job) functions as expected

infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
```diff
@@ -55,32 +57,36 @@ if [[ ${stage} == "staging" || ${stage} == "prod" ]]; then
...
+	# Write scripts necessary for certbot cert renewal to instance
+	cat <<"EOF" > certbot_renew_deploy_hook.sh
+${certbot_renew_deploy_hook_script}
+EOF
+	chmod +x ./certbot_renew_deploy_hook.sh
+
+	cat <<"EOF" > certbot_sync_cert_to_s3_and_cleanup.sh
+${certbot_sync_cert_to_s3_and_cleanup_script}
+EOF
+	chmod +x ./certbot_sync_cert_to_s3_and_cleanup.sh
```
davidsmejia: This should be written at the top so the above conditional can execute it.

infrastructure/api-configuration/certbot/certbot_renew_deploy_hook.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+LOG_FILE="/var/log/cron/certbot_renew.log"
+
+echo "$(date): Cert renewed. Syncing cert with S3 and reloading Ngninx." >> "$LOG_FILE"
+
+./certbot_sync_cert_to_s3_and_cleanup.sh --scpca-portal-cert-bucket ${scpca_portal_cert_bucket}
```
davidsmejia: We can remove the argument to this script and just bake it into the sync script since we would never want to dynamically touch prod vs. staging from the other environment.

infrastructure/api-configuration/certbot/certbot_crontab_entry.txt
```diff
@@ -0,0 +1 @@
+0 0 * * * cd /home/ubuntu && certbot renew --quiet --deploy-hook ./certbot_renew_deploy_hook.sh >> /var/log/cron/certbot_renew.log 2>&1
```
davidsmejia: We should vary the specific time that we actually hit renew to be polite netizens.

infrastructure/api-configuration/certbot/certbot_sync_cert_to_s3_and_cleanup.sh
```diff
@@ -0,0 +1,41 @@
...
+		exit 1
+		;;
+esac
+
+# Validate that value isn't empty
+if [[ -z "$SCPCA_PORTAL_CERT_BUCKET" ]]; then
+	echo "Error: No value provided for --scpca-portal-cert-bucket"
+	exit 1
+fi
+
```
davidsmejia: If you want to log the full cycle we should also log when we upload to s3. This script calls from user-data but that shouldn't matter the immediate log above will inform us if it is a deploy or renew.

infrastructure/api-configuration/certbot/certbot_sync_cert_to_s3_and_cleanup.sh
davidsmejia, line 1: we can rename this to `infrastructure/api-configuration/certbot/certbot_s3_sync.sh`

infrastructure/api-configuration/certbot/cerbot_s3_sync.sh
```diff
@@ -0,0 +1,17 @@
...
+sudo zip -r ../letsencryptdir.zip ../letsencrypt/
+
+# Cleanup the extra copy of nginx.conf added to zip archive
+rm /etc/letsencrypt/nginx.conf
+
+# Sync with S3
+cd /home/ubuntu
+mv /etc/letsencryptdir.zip .
+aws s3 cp letsencryptdir.zip "s3://${scpca_portal_cert_bucket}/"
+rm letsencryptdir.zip
```
davidsmejia:
```suggestion

cd /home/ubuntu

zip -r letsencryptdir.zip /etc/letsencrypt/ /etc/nginx/nginx.conf

# Sync with S3
aws s3 cp letsencryptdir.zip "s3://${scpca_portal_cert_bucket}/"
rm letsencryptdir.zip
```

I think this is all we actually need. Go to the ubunu home dir, recursively zip the letsencrypt dir with the addtional nginx.conf file, then sync it up, and finally clean up.

infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
```diff
@@ -39,14 +40,21 @@ apt install docker-ce docker-ce-cli -y
...
+	apt install certbot python3-certbot-nginx -y
+
+	# Write script which syncs cert with s3 remote store
+	cat <<"EOF" > certbot_s3_sync.sh
+${certbot_s3_sync_script}
+EOF
+	chmod +x ./certbot_s3_sync.sh
+
     # Check here for the cert in S3, if present install, if not run certbot.
     if [[ $(aws s3 ls "${scpca_portal_cert_bucket}" | wc -l) == "0" ]]; then
```
davidsmejia:
```suggestion
    if aws s3api head-object --bucket "${scpca_portal_cert_bucket}" --key letsencrypt.zip > /dev/null 2>&1; then
```

We should be explicit with which file we are pulling or writing to on s3. I dont see any benefit to handling multiple files with arbitrary names when we dont guard against a non-zip file being copied down from s3.

infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
```diff
@@ -55,32 +63,30 @@ if [[ ${stage} == "staging" || ${stage} == "prod" ]]; then
...
-        mv /etc/letsencryptdir.zip .
-        aws s3 cp letsencryptdir.zip "s3://${scpca_portal_cert_bucket}/"
-        rm letsencryptdir.zip
+			PREFIX="$PREFIX.staging"
+		fi
+        certbot --nginx -d $PREFIX.$BASE_URL -n --agree-tos --redirect -m g3w4k4t5n3s7p7v8@alexslemonade.slack.com
+		./certbot_s3_sync.sh
     else
         zip_filename=$(aws s3 ls "${scpca_portal_cert_bucket}" | head -1 | awk '{print $4}')
         aws s3 cp "s3://${scpca_portal_cert_bucket}/$zip_filename" letsencryptdir.zip
```
davidsmejia: at your discretion maybe we also just call the zip_filename `letsencrypt.zip` since it seems like an unnecessary rename

infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
```diff
@@ -55,32 +63,29 @@ if [[ ${stage} == "staging" || ${stage} == "prod" ]]; then
...
     fi
+
+	# Add certbot cert auto renewal entry to cron
+	(crontab -l 2>/dev/null; echo "${certbot_crontab_entry}") | crontab -
+
+	# Write script responsible for certbot cert renewal
+	cat <<"EOF" > certbot_renew_deploy_hook.sh
+${certbot_renew_deploy_hook_script}
+EOF
+	chmod +x ./certbot_renew_deploy_hook.sh
```
davidsmejia: This should be moved up to where we install the other deps.

davidsmejia (approved):
LGTM with one nit

nozomione (approved):
👍

PR #1960, opened 2026-05-14
Set expires_at timestamp and tag output files on S3

### LINKED ISSUE #1951: Set expires_at timestamp and tag output files on S3
### Context

Related Issue: #1910

Feature branch: `feature/user-dataset-expiration` 

The portal features two dataset types:

- `CCDLDatasets`: per-processed datasets for all CCDL projects and metadata available for download (no expiration)
- `UserDatasets`: custom datasets created and processed by users (expire 7 days after processing)

Once `UserDatasets` expire, their corresponding computed files in S3 will be deleted, and the portal provides users an option to regenerate (re-process) expired datasets for download.

### Problem or idea

Currently, there is no 7-day expiration set up for `UserDatasets` resulting in generated computed files remaining on S3 indefinitely. This affects the portal UI as the regeneration option will not be available (see the linked related issue for more context). 

To address this, we need to implement the expiration feature using AWS's tag-based [S3 Lifecycle Policy Rule](https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-set-lifecycle-configuration-intro.html), which enables automated deletion of expired datasets after 7 days. We also need to set up a cron job that marks `UserDatasets` to enable dataset regeneration on the portal after 7 day of processing.

As part of configuring the tag-based S3 Lifecycle Policy Rule for object expiration, we'll implement the following steps:

1. Populate `Dataset.expires_at` upon successful job completion: In the `DatasetJobProcessor.on_run_done` hook method, set the dataset `expires_at` timestamp (now + 7 days) before saving it. The `expires_at` should not populated for `CCDLDatasets`.

```py
# DatasetJobProcessor

"""
on_run_done: Invoked after all steps have completed successfully.
"""

def on_run_done(self):
    self.job.save()
    self.job.dataset.save() 
    logger.info("Job completed.")
```

2. Tagging the generated computed file on S3: In the `DatasetJobProcessor.upload_new_computed_file` workflow step, after successful upload, add object tagging with a key-value pair of `dataset-type=user`. We'll add a new method to handle tagging in the `s3` module (e.g., `tag_user_dataset_output_file`). This method should be skipped for `CCDLDatasets`, and raise an exception on tagging failure.

```py
# DatasetJobProcessor.upload_new_computed_file step:

def upload_new_computed_file(self):
    s3.upload_output_file(
        self.job.dataset.computed_file.s3_key, self.job.dataset.computed_file.s3_bucket
    )
```

3. Add a test to verify the implementation works as expected:
- Assert `Dataset.expires_at` for `UserDatasets` when `DatasetJobProcessor.on_run_done` (mocked) runs
- Ensure the tagging is skipped for `CCDLDatasets` 

Additionally, as part of validation, we'll spin up the `dev` stack and create one `CCDLDataset` and two `UserDatasetes` to locally confirm behavior.

### Solution or next step

- Set the `expires_at` timestamp in `DatasetJobProcessor.on_run_done` 
- Tag the generated computed files upon uploading to S3 during the `DatasetJobProcessor.upload_new_computed_file` step
- Add a new test for the `Dataset.expires_at` field and mock if `DatasetJobProcessor.on_run_done` is called
- Verify the implementation using the `dev` stack

nozomione:
## Issue Number

Closes #1951

Feature branch: `feature/user-dataset-expiration`

Changes include: 
- Dataset state management:
   - Set the expires_at timestamp in `DatasetABC.apply_job_state`
   - Added the new property `expiration_delta` to the Dataset base class
- Dataset Tagging:
   - Added the new step `tag_new_computed_file` to `DatasetJobProcessor` for tagging `UserDataset` computed files 
   -  Added the new property `tags` to the Dataset base class
   - Added a new generic method `s3.tag_output_file` for tagging objects
   - Added new exceptions for S3 operations when uploading or tagging failure (🗒️ we'll refactor the `s3` modules later)
      - Sent Slack alert on tag failure so that the team can manually tag the S3 object (rarely occurs, e.g., AWS error).

## Types of changes

## Functional tests

- Added a new test `test_dataset_job_processor`

## Checklist

## Screenshots
N/A

### DIFF
--- a/api/scpca_portal/models/datasets/base.py
+++ b/api/scpca_portal/models/datasets/base.py
@@ -563,7 +563,7 @@ class DatasetABC(TimestampedModel, models.Model):
         Updates state attributes of the given datasets in bulk.
         """
         STATE_UPDATE_ATTRS = [
-            "is_expired",
+            "expires_at",
             "is_pending",
             "pending_at",
             "is_processing",
--- a/api/scpca_portal/models/datasets/base.py
+++ b/api/scpca_portal/models/datasets/base.py
@@ -118,6 +118,11 @@ class DatasetABC(TimestampedModel, models.Model):
         return self._meta.model
 
     @property
+    def expiration_delta(self) -> datetime | None:
+        # Expiration for processed datasets
+        return None
+
+    @property
     def tags(self) -> dict[str, str] | None:
         # Tagging for S3 Lifecycle Policy Rule
         return None
@@ -192,7 +197,7 @@ class DatasetABC(TimestampedModel, models.Model):
     def get_metadata_file_contents(self) -> List[tuple[str | None, Modalities | None, str]]:
         """
         Return a list of three element tuples which includes the project_id, modality,
-        and their associatied metadata file contents as a string.
+        and their associated metadata file contents as a string.
         """
         # We only return one metadata file for all metadata datasets
         # TODO: if we need to put project metadata files in a project folder,
@@ -544,6 +549,11 @@ class DatasetABC(TimestampedModel, models.Model):
 
         setattr(self, f"is_{state_str}", True)
         setattr(self, f"{state_str}_at", make_aware(datetime.now()))
+
+        # Populate the expiration date if required
+        if state == JobStates.SUCCEEDED and self.expiration_delta:
+            self.expires_at = self.expiration_delta
+
         if hasattr(self, f"{state_str}_reason"):
             setattr(self, f"{state_str}_reason", getattr(job, reason_attr))
 
@@ -553,6 +563,7 @@ class DatasetABC(TimestampedModel, models.Model):
         Updates state attributes of the given datasets in bulk.
         """
         STATE_UPDATE_ATTRS = [
+            "is_expired",
             "is_pending",
             "pending_at",
             "is_processing",
--- a/api/scpca_portal/models/datasets/base.py
+++ b/api/scpca_portal/models/datasets/base.py
@@ -123,8 +123,7 @@ class DatasetABC(TimestampedModel, models.Model):
         return None
 
     @property
-    def tags(self) -> dict[str, str] | None:
-        # Tagging for S3 Lifecycle Policy Rule
+    def s3_upload_tags(self) -> dict[str, str] | None:
         return None
 
     # HASHING AND CACHED ATTR LOGIC
--- a/api/scpca_portal/models/datasets/base.py
+++ b/api/scpca_portal/models/datasets/base.py
@@ -117,6 +117,11 @@ class DatasetABC(TimestampedModel, models.Model):
     def get_class(self) -> models.Model:
         return self._meta.model
 
+    @property
+    def tags(self) -> dict[str, str] | None:
+        # Tagging for S3 Lifecycle Policy Rule
+        return None
+
     # HASHING AND CACHED ATTR LOGIC
     def get_hashes(self) -> tuple[str, str, str, str]:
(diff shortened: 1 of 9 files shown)


nozomione:
I've applied the 1:1 feedback, and this PR is ready for another look. Thank you, David!

api/scpca_portal/models/datasets/base.py
```diff
@@ -117,6 +117,16 @@ def save(self, *args, **kwargs) -> None:
     def get_class(self) -> models.Model:
         return self._meta.model
 
+    @property
+    def expiration_delta(self) -> datetime | None:
+        # Expiration for processed datasets
+        return None
+
+    @property
+    def tags(self) -> dict[str, str] | None:
```
davidsmejia: I think `s3_upload_tags` would probably make more sense as a name, the s3 lifecycle policy rule is decoupled from this.

davidsmejia (approved):
LGTM once we rename the tags attr. Honestly, the tags probably don't belong on the dataset since that is more of a computed_file concern. However, for now we can leave it as is and rework it once computed file has been cleaned up.

api/scpca_portal/s3.py
```diff
@@ -225,6 +225,25 @@ def upload_output_file(key: str, bucket_name: str) -> bool:
...
+
+    tagging = {"TagSet": [{"Key": k, "Value": v} for k, v in tags.items()]}
+
+    try:
+        aws_s3.put_object_tagging(Bucket=bucket_name, Key=key, Tagging=tagging)
+    except Exception as error:
+        logger.error(f"Failed to tag computed file {key} due to the following error:\n\t{error}")
+        return False
+
+    return True
```
avrohomgottlieb, line 244: It's unfortunate for us in this instance that AWS uses the word "key" to refer to the name of the file and the word "key" to refer to a key in a key-value tagging pair. It's all built on key-value pairs, even the filename, but it's a little confusing here.

There's an argument to say that it would be beneficial to rename the parameter to s3_key or s3_key_filename or something of the like to differentiate it from the tag keys, but that would deviate from the function signatures of all the other ones in this module which define the parameter as `key: str`.

api/scpca_portal/models/datasets/ccdl_dataset.py
```diff
@@ -45,6 +46,14 @@ def get_or_find(
...
         return dataset, False
 
+    @property
+    def expiration_delta(self) -> datetime | None:
+        return None
+
+    @property
+    def s3_upload_tags(self) -> dict[str, str] | None:
+        return None
+
```
avrohomgottlieb:
```suggestion

```

This isn't necessary because it's already defined in the base class.

avrohomgottlieb (approved):
Great PR. One cleanup fix and a thought provoker, but this is good to go. Great job!

api/scpca_portal/s3.py
nozomione: Great point! Renaming the `key` parameter to something more explicit (e.g., `s3_key` or `s3_key_filename` as you suggested) would definitely enhance clarity. Indeed I matched the parameter names to align with others in the `s3` module for consistency in this PR. We can create a separate issue to apply your suggestion to reduce ambiguity! 

Tagging @davidsmejia for his insights!

davidsmejia: I would just stick with convention for now.

If anything I would probably just rename the tagging argument to `tag_set` since that is what it represents, where `tagging` is the wrapping dict.

The other nitpick would be to either raise or return True but this is probably best done in a separate PR that makes sure we are either raising or returning bool across the module, probably more broadly.

PR #1961, opened 2026-05-14
Configure tag-based S3 Lifecycle policy and add a cron job to mark expired datasets

### LINKED ISSUE #1952: Configure tag-based S3 Lifecycle policy and add a cron job to mark expired datasets
### Context

Related Issue: #1910

Feature branch: `feature/user-dataset-expiration` 

In the issue #1951, we populated the `expires_at` timestamp and tagged the corresponding computed files for `UserDatasets` upon successful job completion. 

As the next step, we'll configure the S3 Lifecycle Policy Rule for automated deletion of tagged objects. Additionally, we'll set up a cron job that marks expired `UserDatasets` to enable dataset regeneration via the Portal UI.

### Problem or idea

We can configure the tag-based S3 Lifecycle Policy Rule for automatic object deletion via Terraform.

e.g.) Apply a tag-based rule in Terraform (See [Tag-Based Lifecycle Rules](https://oneuptime.com/blog/post/2026-02-12-set-up-s3-lifecycle-policies-with-terraform/view#tag-based-lifecycle-rules)): 

```py 
resource "aws_s3_bucket_lifecycle_configuration" "tagged_lifecycle" {
  bucket = aws_s3_bucket.data.id

  rule {
    id     = "delete-temporary-data"
    status = "Enabled"

    filter {
      tag {
        key   = "DataType"
        value = "temporary"
      }
    }

    expiration {
      days = 7
    }
  }
}
```

To set up a cron job , we'll add a new management command (e.g., `mark_expired_datasets`) that sets `Dataset.is_expired` to `true` if `Dataset.expires_at` exceeds `datetime.now()`.  Then, we'll register a new cron job entry via the user data script, `api-server-instance-user-data.tpl`, to executes `mark_expired_datasets` hourly.  

> [!note]
> AWS executes Lifecycle rule at 00:00 UTC daily (see [Lifecycle rules: Based on an object's age](https://docs.aws.amazon.com/AmazonS3/latest/userguide/intro-lifecycle-rules.html#intro-lifecycle-rules-number-of-days)). This may result in object deletion not happening exactly 7 days after creation. To address this, we'll set the expatriation to **8 days** on S3, and the cron job checks for expiration hourly after 7 days to provide a buffer and prevent unexpected UX behavior on the portal.

### Solution or next step

- Configure the S3 Lifecycle Policy Rule for tagged object deletion with the 8-day expiration period via Terraform
- Implement a new management command, `mark_expired_datasets`, to set `Dataset.is_expired` to `true` after 7 days
- Register a new cron job  entry that runs `mark_expired_datasets` via `api-server-instance-user-data.tpl`
- Verify the implementation works as expected using the `dev` stack

nozomione:
## Issue Number

Closes #1952

Feature branch: `feature/user-dataset-expiration`

Stacked PR #1960

Changes include:
- Configured the S3 Lifecycle Policy Rule for tagged object deletion with the 8-day expiration period via Terraform
- Added a new management command `expire_user_datasets` and the corresponding test
- Added a new class method `UserDataset.mark_expired_datasets` that marks user datasets as expired
- Added a new hourly cron job for `expire_user_datasets` and configured the corresponding watch log

## Types of changes

## Functional tests

- Added a new test `test_expire_user_datasets`
- Tested and verified the implementation using the `dev` stack ✅ 

## Checklist

## Screenshots

N/A


nozomione:
🗒️ For now, an hourly cron job for the `expire_user_datasets` has been added to `crontab`. The implementation is subject to change once PR #1946 is merged into `feature/user-dataset-expiration`.

davidsmejia (approved):
LGTM

api/scpca_portal/models/datasets/user_dataset.py
```diff
@@ -321,3 +322,34 @@ def get_project_sample_counts(self) -> Dict[str, int]:
...
+
+            if dataset.expires_at < now:
+                dataset.is_expired = True
+                updated_datasets.append(dataset)
+                updated_fields.add("is_expired")
+
+        if updated_fields:
+            cls.objects.bulk_update(updated_datasets, list(updated_fields))
+
+        return len([dataset for dataset in updated_datasets if dataset.is_expired])
```
avrohomgottlieb:
```suggestion
        return sum(1 for dataset in updated_datasets if dataset.is_expired)
```

avrohomgottlieb (approved):
This LGTM. I have a design question that I'll add to the next PR, but this one does exactly what it's supposed to. Good job.

api/scpca_portal/models/datasets/user_dataset.py
```diff
@@ -321,3 +322,34 @@ def get_project_sample_counts(self) -> Dict[str, int]:
...
+        """
+        processed_datasets = cls.objects.filter(is_succeeded=True, is_expired=False)
+        now = make_aware(datetime.now())
+
+        updated_fields = set()
+        updated_datasets = []
+
+        for dataset in processed_datasets:
+            # Ensure the expires_at timestamp is set for all processed datasets
+            if dataset.expires_at is None:
```
davidsmejia: Now that we spoke about this again, I think it makes sense to remove this as it would be misleading to debug. We have a separate script to handle anything that was not configured correctly. This command and function pair should just handle the correctly configured expirable datasets by updating the dataset and purging the computed file. The other command should look for anything that was not expirable (no expires_at set) and then handle those separately to make sure they are configured correctly.

We should either separate the commands or unify them by having one command that handles everything.

Tagging @avrohomgottlieb @nozomione for input

avrohomgottlieb: I agree with this sentiment. As long as there's no edge case that we're concerned about where the `expires_at` attr wouldn't be set and we'd want to clean it up here, it can be removed.

avrohomgottlieb: Honestly, with the management command in https://github.com/AlexsLemonade/scpca-portal/pull/1968/ querying datasets as below, it's **expecting** the `expired_at` attr to **not** be set, so we'd want to avoid setting them here. 
```py
datasets = UserDataset.objects.filter(expires_at=None, is_succeeded=True)
```

Let's keep the concerns separate.

nozomione: Thank you for bringing this up! I also agree with your suggestion! 

Once the cleanup command (in #1968) is run against the old datasets (handling field misconfiguration, i.e.,`expires_at=None && is_succeeded=True`, and deleting old computed files), the check for `if dataset.expires_at is None` will become irrelevant (unless any edge cases as Avrohom suggested!). 

With the introduction of the new workflow implementation (including the merged #1960), the `expires_at` timestamp should populated correctly during the dataset job processing lifecycle, and that allows the `expire_user_datasets` management command to focus solely on marking the datasets. This separation will improve clarity during debugging. 

Let's move forward with this approach unless there are any other points to discuss!

nozomione:
@davidsmejia & @avrohomgottlieb, Thank you for the great feedback! 

I've made the following changes, and this PR is ready for your review. Thank you!
- Replaced `len` with `sum` in the return statement
- Purged computed files (unlinked) for expired datasets
- Removed the misconfiguration check for `expires_at`
- Updated the corresponding tests to verify the implementation

api/scpca_portal/models/datasets/user_dataset.py
```diff
@@ -321,3 +322,28 @@ def get_project_sample_counts(self) -> Dict[str, int]:
...
+            if dataset.expires_at < now:
+                dataset.is_expired = True
+                if dataset.computed_file:
+                    dataset.computed_file.purge()
+                updated_datasets.append(dataset)
+
+        if updated_datasets:
+            cls.objects.bulk_update(updated_datasets, ["is_expired"])
+
+        return sum(1 for dataset in updated_datasets if dataset.is_expired)
```
davidsmejia:
```suggestion
        now = timezone.now()
        expired_datasets = cls.objects.filter(is_succeeded=True, is_expired=False, expires_at__lte=now)
        
        ComputedFile.objects.filter(dataset_id__in=expired_datasets).delete()
        
        return expired_datasets.update(is_expired=True)
```

I think this is the best way to handle it with 2 queries instead of n + 1

davidsmejia: We can delete the ComputedFiles directly because we dont care about purging from s3 for this dataset when expiring. We should probably add some comments here to illuminate that distinction.

nozomione: Thank you for the awesome feedback 💯 🚀 !

avrohomgottlieb (approved):
LGTM

PR #1964, opened 2026-05-19
Add yarn lint to Pre Commit

### LINKED ISSUE #1957: Add yarn lint command to pre-commit
### Context

GHA and Vercel fail if the yarn lint isn't cleared.

### Problem or idea

Instead of waiting for commits to fail after being pushed up, we'd like for this to be addressed locally. This can be accomplished by adding a yarn lint check to pre-commit.

### Solution or next step
- Add yarn lint to pre-commit

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/1957.

This PR includes the following:
  - Adds `yarn lint` (Biome) as a pre-commit hook that runs on changes to `client/` files
  - Adds `supportedArchitectures` to `client/.yarnrc.yml` so yarn install downloads both Linux and macOS Biome binaries (resolving the platform mismatch that caused the hook to fail when node_modules was populated by Docker)
  - Adds `client` and `client:lint` commands to sportal

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,8 +38,8 @@ repos:
   - repo: local
     hooks:
       - id: yarn-lint
-        name: yarn lint (client)
-        entry: bash -c "cd client && yarn lint"
+        name: Yarn Lint (Biome)
         language: system
+        entry: bash -c "cd client && yarn lint"
         pass_filenames: false
         files: ^client/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,6 +40,6 @@ repos:
       - id: yarn-lint
         name: Yarn Lint (Biome)
         language: system
-        entry: bash -c "cd client && yarn lint"
+        entry: bin/sportal client:lint
         pass_filenames: false
         files: ^client/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,4 +41,5 @@ repos:
       - id: biome-check
         name: biome check
         files: ^client/
-        entry: bash -c 'cd client && exec biome check --write'
+        args:
+        - --config-path=client/biome.json
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,3 +34,12 @@ repos:
     hooks:
       - id: black
         language_version: python3
+
+  - repo: local
+    hooks:
+      - id: yarn-lint
+        name: yarn lint (client)
+        entry: bash -c "cd client && yarn lint"
+        language: system
+        pass_filenames: false
+        files: ^client/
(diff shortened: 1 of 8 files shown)


avrohomgottlieb:
Adding `yarn lint` to pre-commit surfaced a platform mismatch: `node_modules/` is populated by the Docker `client-deps-installer` on Linux, so the macOS Biome binary (`@biomejs/cli-darwin-arm64`) is never installed, causing the hook to fail locally.

Two approaches were considered:

**Option A — pre-commit `language: node`:** pre-commit manages its own isolated Biome install, downloading the correct platform binary automatically. The downside of this is that it decouples Biome's version from `package.json` and uses a different invocation (`biome lint`) than the GHA "Run Linters` Action  (`yarn lint`), creating drift risk.

**Option B — `supportedArchitectures` in `.yarnrc.yml`:** tells Yarn to install optional platform binaries for both `linux` and `darwin` regardless of which OS runs yarn install. `yarn lint` remains the single driver across all environments with no syncing necessary. The cost of this is a larger local `node_modules/` folder (including @next/swc-darwin-arm64), as well as longer run times for the pre-commit call in GHA.

Option B was chosen, prioritizing consistency between environments over local disk space and GHA Pre-Commit speed.

.pre-commit-config.yaml
```diff
@@ -34,3 +34,11 @@ repos:
...
       - id: black
         language_version: python3
+
+  - repo: https://github.com/biomejs/pre-commit
+    rev: v2.4.15
+    hooks:
+      - id: biome-check
+        name: biome check
+        files: ^client/
+        entry: bash -c 'cd client && exec biome check --write'
```
avrohomgottlieb, line 44: Ideally we'd want to run it without an entry command, and to point to the config file path explicitly, but the below error continues to creep up:

```suggestion
        args:
        - --config-path=client/biome.json
```

Error:
```
scpca-portal/client/biome.json configuration ━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Found a nested root configuration, but there's already a root configuration.
  
  ℹ The other configuration was found in scpca-portal.
  
  ℹ Use the migration command from the root of the project to update the configuration.
  
  $ biome migrate --write
  

configuration ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Biome exited because the configuration resulted in errors. Please fix them.
  

```

davidsmejia: This is fine until a PR for this lands: https://github.com/biomejs/biome/issues/8672

davidsmejia (approved):
LGTM

nozomione (approved):
LGTM 👍

Issue #1944, opened 2026-04-27
Add Static Analysis With Mypy


avrohomgottlieb:
### Context

With PR https://github.com/AlexsLemonade/scpca-portal/pull/1924, type hints have officially been added to functions throughout the codebase.

### Problem or idea

We'd like to experiment with the static analyzer `mypy`. This Issue should focus on configuring and testing mypy, and deciding on whether or not we'd like to add it to our project. We should also look into where it would be used in the workflow, how it would be added to pre-commit and GitHub actions.

### Solution or next step
- Install and configure mypy
- Run mypy and analyze results
- Investigate and configure integration into pre-commit and GitHub actions

PR #1967, opened 2026-05-19
Fix Exception and Management Command Mypy Errors

avrohomgottlieb:
## Issue Number

https://github.com/AlexsLemonade/scpca-portal/issues/1944

This begin Mypy introduction to the codebase with static analysis code fixes to the `exceptions/` and `management/commands/` dirs

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/exceptions/dataset_data_validation_error.py
+++ b/api/scpca_portal/exceptions/dataset_data_validation_error.py
@@ -1,6 +1,6 @@
 from typing import Iterable
 
-from scpca_portal.enums import Modalities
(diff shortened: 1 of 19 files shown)


davidsmejia (approved):
LGTM

This opens the following questions:
- Do we want to parse arguments that we don't use, I think that *args and *kwargs if there are no consumers, it might make sense to remove them.
- Do we need a policy or direction for when is it ok to use `Any`. This seems like it would negate the benefit of using mypy, for args and kwargs it makes sense but we should try to define the rule for internal purposes.

nozomione (approved):
LGTM 👍

PR #2017, opened 2026-06-26
fix certbot and nginx archival and extraction

davidsmejia:
## Issue Number

N/A

On staging deploy ran into an issue where nginx wasnt loading the correct file. This was because of a double nested /etc/etc/letsencrypt|nginx folders. This PR resolves that as well as changes all tabs to double spaces.

Note: This change will require updating the archived folders in nginx before deploys. Going to do that while this gets reviewed.

This pull request refines the process for backing up and restoring SSL certificate files and the Nginx configuration for the API server. The changes ensure that the archive structure is consistent and that certificate renewals work smoothly after restoration. The main improvements are in how the archive is created and restored, and in the handling of file overwrites.

**Improvements to certificate backup and restore process:**

* The `certbot_s3_sync.sh` script now creates the `letsencrypt.zip` archive with paths relative to `/etc` and uses the `-y` flag to preserve symlinks, which is important for certbot's renewal process. The archive is created in `/home/ubuntu` and then uploaded to S3.
* When restoring the archive, the `unzip` command now uses the `-o` flag to overwrite existing files without prompting, ensuring the restored `nginx.conf` replaces the old one cleanly.

**Minor cleanup:**

* Removed an unnecessary blank line after running the certbot sync script in the user data template.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
+++ b/infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
@@ -40,27 +40,27 @@ apt install docker-ce docker-ce-cli -y
 sudo usermod -a -G docker ubuntu && newgrp docker
 
 if [[ ${stage} == "staging" || ${stage} == "prod" ]]; then
-  # Create and install SSL Certificate for the API.
-  # Only necessary on staging and prod.
-  # We cannot use ACM for this because *.bio is not a Top Level Domain that Route53 supports.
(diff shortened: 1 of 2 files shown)


nozomione (approved):
👍

avrohomgottlieb (approved):

PR #2020, opened 2026-07-06
Implement Loadable Resource Abstract Base Class

### LINKED ISSUE #2005: Create Loadable Resource Abstract Base Class
### Context

As outlined in the [RFC: In-Place Model Updates for Project, Sample, and Library](https://data-lab-knowledge.slab.com/posts/rfc-in-place-model-updates-for-project-sample-and-library-f4851pxy), we're going to be adding in-place model updating for the project, sample and library models. 
 
### Problem or idea

This ticket will focus on adding an abstract base class called `LoadableResourceABC`, which will implement state, hash, and loaded_at attrs, with get_hash and update_state property methods.

### Solution or next step
- Add `LoadableResourceABC` abstract base class implement methods
- Add migration

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/2005.

This PR adds the initial version of the LoadableResourceABC.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/models/loadable_resource_abc.py
+++ b/api/scpca_portal/models/loadable_resource_abc.py
@@ -1,0 +1,14 @@
+from django.db import models
+
+from scpca_portal.config.logging import get_and_configure_logger
+
+logger = get_and_configure_logger(__name__)
+
+
+class LoadableResourceABC(models.Model):
+    class Meta:
+        abstract = True
+
+    state = models.TextField()
+    hash = models.CharField(max_length=32, null=True)
+    loaded_at = models.DateTimeField(null=True)
--- a/api/scpca_portal/models/loadable_resource_abc.py
+++ b/api/scpca_portal/models/loadable_resource_abc.py
@@ -26,16 +26,18 @@ class LoadableResourceABC(TimestampedModel, models.Model):
 
     def update_loadable_state(self, new_state: LoadableResourceStates) -> None:
         """Updates the state and synchronization tracking fields for a single loadable resource."""
-        update_fields = ["state"]
         self.state = new_state
+        self.updated_at = make_aware(datetime.now())
+        # this is necessary so updated_at is not overwritten by auto_new during save op,
+        # allowing for us to set loaded_at after updated_at (if need be)
+        fields_to_update = ["state", "updated_at"]
+
         if new_state == LoadableResourceStates.SYNCED:
             self.hash = self.current_hash
-            self.updated_at = make_aware(datetime.now())
             # loaded_at needs to be set after updated_at to ensure aggregations are re-computed
             self.loaded_at = self.updated_at + timedelta(microseconds=1)
-            # this is necessary so updated_at is not overwritten by auto_new during save op
-            update_fields.extend(["hash", "updated_at", "loaded_at"])
-        self.save(update_fields=update_fields)
+            fields_to_update.extend(["hash", "loaded_at"])
+        self.save(update_fields=fields_to_update)
 
     @classmethod
     def bulk_update_loadable_state(
--- a/api/scpca_portal/models/loadable_resource_abc.py
+++ b/api/scpca_portal/models/loadable_resource_abc.py
@@ -49,14 +49,17 @@ class LoadableResourceABC(TimestampedModel, models.Model):
             # This is an optimized approach to grabbing all hash values
             # in order for the row updating to stay at the DB level
             hash_cases = [
-                models.When(
-                    pk=loadable_resource.pk, then=models.Value(loadable_resource.current_hash)
-                )
-                for loadable_resource in loadable_resources
+                models.When(pk=lr.pk, then=models.Value(lr.current_hash))
+                for lr in loadable_resources
             ]
 
+            loadable_resources_pks = [lr.pk for lr in loadable_resources]
             now_timestamp = make_aware(datetime.now())
-            loadable_resources.update(
+            # filter explicitly by primary keys rather than updating the original queryset directly.
+            # this prevents a race condition where a phantom row matching the original filters
+            # is added to the db after hash_cases evaluates, which would cause an invalid state
+            # inside the SQL CASE statement during the update.
+            cls.objects.filter(pk__in=loadable_resources_pks).update(
                 hash=models.Case(*hash_cases, output_field=cls._meta.get_field("hash")),
                 state=new_state,
                 updated_at=now_timestamp,
--- a/api/scpca_portal/models/loadable_resource_abc.py
+++ b/api/scpca_portal/models/loadable_resource_abc.py
@@ -1,10 +1,9 @@
 from abc import abstractmethod
-from typing import List
+from datetime import datetime, timedelta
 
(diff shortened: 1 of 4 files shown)


davidsmejia (commented):
This looks good to me. Couple of questions:

- Why not `@abstractmethod` for update, bulk_update? Shouldn't we enforce their implementation as well as part of the Loadable contract?
- Shouldnt we enforce ABCModelMeta?

avrohomgottlieb:
> * Why not `@abstractmethod` for update, bulk_update? Shouldn't we enforce their implementation as well as part of the Loadable contract?

Absolutely. I think I was still thinking through how I wanted to implement these methods. I just pushed up some new commits, thinking about it a little more critically I don't see a need to differentiate between updating state from updating the other attrs, so I removed the update/bulk_update functions and cleaned everything up. I think it reads nicer now. 

> * Shouldnt we enforce ABCModelMeta?

In the past what we've done is to assign it as an abstract model (`abstract = True` in `Meta` nested class) instead of as an abstract base class (`ABCModelMeta`) to allow it to play better with its inherited models.

api/scpca_portal/models/loadable_resource_abc.py
```diff
@@ -0,0 +1,85 @@
...
+    def update_loadable_state(self, new_state: LoadableResourceStates) -> None:
+        """Updates the state and synchronization tracking fields for a single loadable resource."""
+        if new_state == LoadableResourceStates.SYNCED:
+            # Executes SQL Update directly on DB, ensuring that loaded_at is set after updated_at,
+            # which is important for determining whether aggregations need to be re-computed
+            type(self).objects.filter(pk=self.pk).update(
+                hash=self.current_hash,
+                state=new_state,
+                updated_at=make_aware(datetime.now()),
+                loaded_at=F("updated_at") + timedelta(microseconds=1),
```
davidsmejia: Why not just have this as a variable then add the delta? I think `F("updated_at")` would refer to the value pre-update no?

api/scpca_portal/models/loadable_resource_abc.py
```diff
@@ -0,0 +1,85 @@
...
+            self.refresh_from_db(fields=["hash", "state", "updated_at", "loaded_at"])
+        else:
+            self.state = new_state
+            self.save()
+
+    @classmethod
+    def bulk_update_loadable_state(
+        cls, loadable_resources: QuerySet[Self], new_state: LoadableResourceStates
+    ) -> None:
+        """Performs a highly optimized batch update on a QuerySet of loadable resources."""
```
davidsmejia, line 52:
```suggestion
        """Performs a batch update on a QuerySet of loadable resources."""
```

This still does N+1 queries so maybe just calling it a batch update is more appropriate.

api/scpca_portal/models/loadable_resource_abc.py
```diff
@@ -0,0 +1,85 @@
...
+                    pk=loadable_resource.pk, then=models.Value(loadable_resource.current_hash)
+                )
+                for loadable_resource in loadable_resources
+            ]
+
+            loadable_resources.update(
+                hash=models.Case(*hash_cases, output_field=cls._meta.get_field("hash")),
+                state=new_state,
+                updated_at=make_aware(datetime.now()),
+                loaded_at=F("updated_at") + timedelta(microseconds=1),
```
davidsmejia: Same as above

api/scpca_portal/models/loadable_resource_abc.py
```diff
@@ -0,0 +1,85 @@
...
+            )
+            # This ensures that the object's fields have their updated values for downstream use
+            self.refresh_from_db(fields=["hash", "state", "updated_at", "loaded_at"])
+        else:
+            self.state = new_state
+            self.save()
+
+    @classmethod
+    def bulk_update_loadable_state(
```
davidsmejia: I think we should re-evaluate the no-return here. Modifying in place and assuming downstream will work off of the same queryselector means that if we modify something that was used to filter that qs then it would not be part of the new result set. I think since we already have a cached version of the initial results we can just return the modified objects.

avrohomgottlieb: This makes sense.

api/scpca_portal/models/loadable_resource_abc.py
avrohomgottlieb: This was a good catch. I updated it.

api/scpca_portal/models/loadable_resource_abc.py
avrohomgottlieb: I removed the N+1 queries under the previous `update`/`refresh_from_db` paradigm and just folded everything into one `bulk_update` action. 

As an FYI - there's a different N+1 query situation where we're hitting the OF table every time we update the hash (every time we call `current_hash` which invokes the `original_files` property). I looked into it, and there is a `prefetch_related` construct in Django which will allow us to grab all OFs in a bulk set at once. This construct is dependent on real relations though, not what we currently have, so when we transition from the OF project/sample/library id relations to real foreign key relations, we'll be able to utilize this.

api/scpca_portal/models/loadable_resource_abc.py
davidsmejia, line 1: This looks pretty good to me. A couple things I just want to comment on but don't necessarily needs to be addressed in this PR, though it would be good to make sure we acknowledge them now.

- Should we namespace the attributes / methods since they currently have a generic names ex `hash`? 
- Should we restructure the other base classes to be in a base folder each getting their own file and then put this new base class in there?
- Since this base class has a direct dependency on the Timestamped Model does it make more sense to break this up so we have just the attributes inherited and the methods inherited separately?

davidsmejia (approved):
Approved as is but I just wanted to raise some concerns that we should think through before finally committing to this approach.

api/scpca_portal/models/loadable_resource_abc.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @property
+    @abstractmethod
+    def original_files(self) -> QuerySet[OriginalFile]:
+        pass
+
+    @property
+    def current_loaded_hash(self) -> str:
+        original_file_hashes = self.original_files.values_list("hash", flat=True)
+        return utils.hash_values(original_file_hashes)
```
nozomione, line 75: **Suggestion:**  `utils.hash_values` currently expects `List[str]`, but we also pass a `QuerySet`. Would it be good to change the type hint to `Iterable[str]` to support both a list and a flat values_list queryset?

nozomione (approved):
I left a small suggestion. Otherwise, it looks great! LGTM 👍

api/scpca_portal/models/loadable_resource_abc.py
avrohomgottlieb: This is a good suggestion. I'm going to address it in the stacked PR https://github.com/AlexsLemonade/scpca-portal/pull/2044.

PR #2038, opened 2026-07-10
Remove Obsolete ComputedFile-related Logic in Models and Relationships

nozomione:
## Issue Number

Closes #2030 
Stacked from PR #2036

## Purpose/Implementation Notes

This PR cleans up the obsolete code blocks for the legacy computed file downloads in models, and removes the project and sample foreign keys from the `ComputedFile` model with migration.

## Types of changes

- Refactor (addresses code organization and design mentioned in corresponding issue)

## Functional tests

Updated: 

- `test/models/test_project.py`

Deleted: 

- `test/models/test_sample.py`

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have updated any associated Storybook stories (if applicable)

## Screenshots

N/A

PR #2041, opened 2026-07-14
Assert that Datasets always have related ComputedFiles

### LINKED ISSUE #2032: Assert that Datasets always have related ComputedFiles and Job
### Context

As legacy `ComputedFile` downloads for projects and individual samples have been replaced by dataset downloads, we should ensure that each each proceed dataset has at most one related  `ComputedFile` and `Job` instance.

### Problem or idea

We should add the following tests to assert that each processed dataset always has at most one related `ComputedFile` and `Job`:
- Has One-To-0ne relationship with `ComputedFile`: One `ComputedFile` instance
- Has One-To-Many relationship with `Job` (defined with the `GenericRelation` field): At least one `Job` instance

### Solution or next step

Add the tests to ensure that every processed dataset instance will always have at most one related computed file and job instance.

nozomione:
## Issue Number

Closes #2032 

Stacked from #2038 

This PR adds assertions to the existing tests to ensure that processed datasets at most one job and one computed file.

## Types of changes

## Functional tests

Updated:
- `test/models/test_computed_file.py`
- `test/models/test_job.py`

## Checklist

## Screenshots
N/A

### DIFF
--- a/api/scpca_portal/models/job.py
+++ b/api/scpca_portal/models/job.py
@@ -158,8 +158,7 @@ class Job(TimestampedModel):
         new_job.apply_state(JobStates.PENDING)
         if save:
             new_job.save()
-            if new_job.dataset:  # TODO: Remove after the dataset release
-                new_job.dataset.save()
+            new_job.dataset.save()
 
         return new_job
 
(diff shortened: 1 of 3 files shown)


davidsmejia (commented):
This should be less about the actual assertions in the test suite but more that do we need to keep checking job.dataset conditionals in the code.

nozomione:
> This should be less about the actual assertions in the test suite but more that do we need to keep checking job.dataset conditionals in the code.

~~I've removed the `job.dataset`check and verified the changes on the `dev` server. A dataset was processed successfully, along with its corresponding computed file and job ✅~~

**UPDATE:** I've updated the job/dataset state mapping logic, and ensure that a corresponding dataset exists for a job at the time of the job instantiation in PR #2053. For this, the cleanup for the `if job.dataset` checks was addressed in the PR. 

I've reverted the `Job` model cleanup made in this PR and it's ready for another look. Thank you, David!

davidsmejia (approved):

avrohomgottlieb (approved):
LGTM

PR #2044, opened 2026-07-17
Implement sync_metadata LoadableResourceABC method

avrohomgottlieb:
## Issue Number

Closes https://github.com/AlexsLemonade/scpca-portal/issues/2045.

- Adds a generic `sync_metadata` to the `LoadableResourceABC` base class
- Adds `update_from_dict` functions to the Project, Sample, and Library models
- Adds `get_loaded_state_metadata_dicts_by_id` functions to the Project, Sample, and Library models
- Adds a `test_sync_metadata` method to the `TestLoadableResourceABC` test class
- Refactors the test setup class `ConcreteLoadableResource` to accommodate `sync_metadata` functionality on `LoadableResourceABC`

## Types of changes

## Functional tests

- `TestLoadableResourceABC::test_sync_metadata`

## Checklist

## Screenshots

N/A


api/scpca_portal/models/library.py
```diff
@@ -68,6 +68,17 @@ def get_from_dict(cls, data: Dict, project: "Project") -> Self:
...
+    def update_from_dict(self, data: Dict) -> Self:
+        # TODO: are format and modality fields immutable like scpca_id and project, or not?
+        self.is_multiplexed = data.get("is_multiplexed", False)
+        self.has_cite_seq_data = OriginalFile.downloadable_objects.filter(
+            library_id=data["scpca_library_id"], is_cite_seq=True
+        ).exists()
+        self.metadata = data
+        self.workflow_version = data["workflow_version"]
+
+        return self
```
avrohomgottlieb, line 100: One current TODO is to figure out how `Library::update_from_dict` should work, specifically which attrs are mutable and which are not.

avrohomgottlieb:
The `metadata_parser` funcs that `Sample` and `Library` call in `get_loaded_state_metadata_dicts_by_id` (and `Project` for that matter as well) could be optimized to better serve them, as `get_loaded_state_metadata_dicts_by_id` on these models is the main entry point to that logic moving forward with the old `load_metadata` workflow being slated for removal.

api/scpca_portal/models/library.py
davidsmejia: everything is mutable according to the source of metadata - no need to cherry pick

api/scpca_portal/models/project.py
```diff
@@ -166,6 +179,41 @@ def get_bulk_rna_seq_sample_ids(self) -> Set[str]:
                 )
         return bulk_rna_seq_sample_ids
 
+    def get_original_files_by_download_config(
```
davidsmejia: This is dead code we removed download configs

api/scpca_portal/models/loadable_resource_abc.py
```diff
@@ -20,12 +21,20 @@ class LoadableResourceABC(TimestampedModel):
     class Meta:
         abstract = True
 
+    # this attribute acts as the defacto primary key on deriving models
+    scpca_id = models.TextField(unique=True)
+
```
davidsmejia: I'm not sold on this change. I don't think the scpca_id should be owned by the ability of the resource. This should be moved to a proper PK on the concrete classes, later on we can consider creating a Resource base class for this.

davidsmejia (commented):
LGTM but we should open an issue to break out the ABC tests and have them be new tests related to the actual concrete classes. The purpose of the abc test suite should be strictly tied to what is in the abc.

davidsmejia (approved):

nozomione (approved):
LGTM 👍

api/scpca_portal/models/project.py
nozomione: @avrohomgottlieb, a quick note 🗒️ : this method was removed in #2038 and merged into `dev`

avrohomgottlieb: Thanks for pointing this out. It must have snuck in there when dev was merged into the feature branch. Commit https://github.com/AlexsLemonade/scpca-portal/pull/2044/commits/03152f45db022645c0a53959681e60f6b2981154 took care of it.

PR #2053, opened 2026-07-22
Update Job and Dataset State Mapping Logic

### LINKED ISSUE #2028: Update Job and Dataset State Mapping Logic
### Context

As part of Dataset and Job reconciliation process, #2019 introduced the `DatasetStates` enum and the `DatasetABC.state` field. This issue incorporates them and refactors the Dataset and Job mapping logic.

### Problem or idea

The `DatasetABC.apply_job_state` method currently maps the dataset state to the latest job state by accepting the most recent job instance as an argument. We should update it to use the `DatasetStates` enum and query the latest job state to sync instead of passing it as an argument, for both single and bulk update.

We'll also replace the `is_*` flag fields in the dataset model (excluding expired-related state) with the `state` field. 

Additionally, we should remove duplicate fields from the dataset model (e.g., state timestamps, failed reason) as they are available via the Job instance.

### Solution or next step

- Incorporate the  `DatasetStates` enum and the `state` field and refactor the  `DatasetABC.apply_job_state`
- Remove the duplicated state fields from the dataset model
- Apply migrations
- Update the existing tests to ensure the states map correctly.

nozomione:
## Issue Number

Closes #2028 

Feature branch: `feature/job-dataset-state-reconciliation`

Changes include:

This PR refactors `DatasetABC.apply_job_state` to use the `DatasetStates` enum and the `state` field, and remove the `job` parameter by query the latest job instead.

**NOTE:** The changed field is 33, but 29 files (mostly tests) relate to reference updates, including the migration file.

Based on the changes above, the following updates were made:
- Refactor the `DatasetABC` model:
  - Added a `latest_job` property
  - Added a `bulk_sync_state` class method, which is called via bulk methods of the `Job` model
  - Removed the `is_*` state field from `DatasetABC` and apply migrations
- Refactored the `Job` model:
  - Cleaned up the `if job.dataset` check (with `TODO`) as no longer required 
  - A single instance `save()` performs job/dataset state mapping as a grouped transaction to enable automatic rollback on exceptions
  - `save()` asserts that the corresponding dataset should exist at the time of job instantiation
  - Methods that mutate state have been adjusted to align with the new state mapping behaviors
- Updated the `clean_up_expired_user_datasets` management command to align with the new state mapping behavior
- Updated the `UserDataset.mark_expired_datasets` to align with the new state mapping behavior
- Updated the`CCDLDataset` and` UserDataset` serializer fields
- Replaced the old dataset state field (`is_*`) references across the codebase

## Types of changes

## Functional tests

- Updated the existing tests to ensure the states map correctly

## Checklist

## Screenshots

N/A


api/scpca_portal/models/datasets/base.py
```diff
@@ -529,58 +518,40 @@ def download_url(self) -> str | None:
 
         return self.computed_file.get_dataset_download_url(self.download_filename)
 
-    def apply_job_state(self, job) -> None:
+    @property
+    def latest_job(self) -> "Job":
+        return self.jobs.order_by("-created_at").first()
+
+    def apply_job_state(self) -> None:
```
davidsmejia: This would throw an error if there were no job yet created, so lets handle that

api/scpca_portal/models/datasets/base.py
```diff
@@ -529,58 +518,40 @@ def download_url(self) -> str | None:
...
-
-        # Populate the expiration date if required
-        if state == JobStates.SUCCEEDED and self.expiration_delta:
+        if self.state == JobStates.SUCCEEDED and self.expiration_delta:
             self.expires_at = self.expiration_delta
 
-        if hasattr(self, f"{state_str}_reason"):
-            setattr(self, f"{state_str}_reason", getattr(job, reason_attr))
+    @classmethod
+    def bulk_sync_state(cls, datasets: List[Self]) -> None:
```
davidsmejia: Should this and the next method be combined?

api/scpca_portal/models/datasets/user_dataset.py
```diff
@@ -75,7 +75,7 @@ def validate_data(data: Dict[str, Any], format: DatasetFormats) -> Dict[str, Any
 
     @property
     def expiration_delta(self) -> datetime | None:
-        return self.succeeded_at + timedelta(days=7)
+        return self.latest_job.succeeded_at + timedelta(days=7)
```
davidsmejia, line 81: we should also have a guard here in the event that there is no job yet

api/scpca_portal/models/job.py
```diff
@@ -127,8 +127,17 @@ def validate_dataset(self) -> None:
...
+        automatically rolling back the transaction if an exception occurs.
+        Calls the dataset's method to sync the job's state.
+        """
         self.validate_dataset()
-        super().save(*args, **kwargs)
+
+        with transaction.atomic():
+            super().save(*args, **kwargs)
+            self.dataset.apply_job_state()
+            self.dataset.get_class().bulk_update_state([self.dataset])
```
davidsmejia: We shouldnt call a bulk version from here

davidsmejia (commented):
let's point this at a feature branch to sync up the client side changes

nozomione:
> let's point this at a feature branch to sync up the client side changes

A feature branch `feature/job-dataset-state-reconciliation` has been created ✅

api/scpca_portal/models/datasets/base.py
nozomione: `bulk_sync_state` and `bulk_update_state` have been combined as `bulk_apply_job_state` (with optional save) ✅

api/scpca_portal/models/job.py
nozomione: `apply_job_state` has been updated to optimally save the instance and is now used in `Job.save()` ✅

nozomione:
My commit message got accidentally cuts off for d51554a: `update DatasetABC.apply_job_state to optionally save the instance, add DatasetABC.bulk_apply_job_state (combined bulk_sync_state & bulk_update_state), add safeguard check for latest_job, and adjust the existing tests (now we enforce that: a dataset must exist at the time of job instantiation)`

nozomione:
I've applied your feedback (including from 1:1), and this PR is now ready for another look. Thank you, David!

nozomione:
@davidsmejia , I've also applied the following feedback:
- Centralized bulk dataset state updates in `DatasetABC.bulk_apply_job_state` (removed the `get_class` logic from bulk-related `Job` methods)
- Removed the `DatasetABC.is_expired`field (the expired dataset `state` will be updated from `SUCCEEDED` to `EXPIRED` via a cron job)

This PR is ready for another look. Thank you, David!

api/scpca_portal/management/commands/clean_up_expired_user_datasets.py
davidsmejia: I think it makes sense here to just add the expired timestamps as a separate loop because in the current implementation we are appending every succeeded dataset to the updated_dataset list.

Also, the expiration_delta can we put those values in the shared config because they done seem to match the comment.

api/scpca_portal/models/datasets/base.py
```diff
@@ -545,15 +544,22 @@ def apply_job_state(self, *, save: bool = True) -> None:
             self.save()
 
     @classmethod
-    def bulk_apply_job_state(cls, datasets: List[Self], *, save: bool = True) -> None:
+    def bulk_apply_job_state(cls, jobs: List["Job"], *, save: bool = True) -> None:
         """
-        Syncs each dataset with its latest job state.
+        Syncs dataset states with the latest state of the given jobs.
```
davidsmejia: You can add a line explaining that since jobs are related to any specific concrete class they need to be broken up by table.

davidsmejia (commented):
Looking good, I had some pending comments and some new ones

api/scpca_portal/management/commands/clean_up_expired_user_datasets.py
davidsmejia: so `backfill_missing_expires_at_timestamps` or something similar, which we would run first.

nozomione: I've made the following changes ✅ :
- Split the dataset update logic into two passes (`backfilled_timestamps`, `marked_expired`)
- Adjusted the `buffer` value (i.e., `timedelta(days=1)`) when marking expired dataset
- Adjusted the help string  and comments to reflect these changes

nozomione:
I've applied your feedback, and this PR is ready for another look. Thank you, David!

api/scpca_portal/management/commands/clean_up_expired_user_datasets.py
davidsmejia: Let's just break this up into two management commands because we dont need to backfil every time so having it in here is noise

backfill_dataset_expires_at

api/scpca_portal/models/datasets/base.py
```diff
@@ -529,58 +515,57 @@ def download_url(self) -> str | None:
...
-        # Sets new state based on the given job
-        state_str = job.state.lower()
-        reason_attr = f"{state_str}_reason"
+    @classmethod
+    def bulk_apply_job_state(cls, jobs: List["Job"], *, save: bool = True) -> None:
+        """
+        Syncs dataset states with the latest state of the given jobs.
+        Optionally saves the datasets.
+        NOTE: Since the given jobs may include both CCDL and user dataset types,
+        they are grouped by class to perform bulk update per table.
```
davidsmejia:
```suggestion
        NOTE: Since jobs have a dynamic Dataset relationship
        they are grouped by class to perform bulk update per table.
```

davidsmejia (approved):
LGTM

api/scpca_portal/management/commands/clean_up_expired_user_datasets.py
nozomione: Done in 64a2e22 ✅

PR #2061, opened 2026-08-04
Load lockfile OriginalFiles

### LINKED ISSUE #2058: Load lockfiles into OF Table
### Context

In PR https://github.com/AlexsLemonade/scpca-portal/pull/2056, while implemented `handle_locked_objects`, we've decided that we want our source of truth for locked files to be lockfile objects in the OriginalFile table.

### Problem or idea

An `is_lockfile` attr should be added to the `OriginalFile` model, and the `OriginalFile` model and `InputBucketS3KeyInfo` class should accommodate setting this.

### Solution or next step
- Release the condition which excludes lockfiles from being loaded into the DB in `sync_original_files`
- Add an `is_lockfile` attr to the OriginalFile model
- Update the InputBucketS3KeyInfo class to handle this case, and set this attr in the `OriginalFile::get_from_dict`.
- Updates tests which confirm this

avrohomgottlieb:
## Issue Number

Closes Issue https://github.com/AlexsLemonade/scpca-portal/issues/2058.

- Add an `is_lockfile` attr to the `OriginalFile` model
- Remove the guard against adding lockfiles as OF objects to the DB in `sync_original_files`
- Add an `is_lockfile` property to the `InputBucketS3KeyInfo` class
- Add 2 new utils functions: `filter_dicts_by_value_substrings` and `exclude_dicts_by_value_substrings`, and add tests.

## Types of changes

## Functional tests

- `TestSyncOriginalFiles::test_sync_original_files_locked_project`
- `utils/test_helpers`
  - `test_filter_dicts_by_value_substrings`
  - `test_exclude_dicts_by_value_substrings`

## Checklist

## Screenshots

N/A


api/scpca_portal/test/utils/test_helpers.py
```diff
@@ -112,6 +112,41 @@ def test_included_keys_not_subset_ignore_value_error(self):
...
+            {"name": "banana bread", "type": "dessert"},
+            {"name": "chicken soup", "type": "meal"},
+        ]
+        key = "name"
+        substrings = ["apple", "bread"]
+
+        expected_result = [{"name": "chicken soup", "type": "meal"}]
+        actual_result = utils.exclude_dicts_by_value_substrings(list_of_dicts, key, substrings)
+
+        self.assertEqual(actual_result, expected_result)
```
davidsmejia, line 161: The line order here makes this a bit convoluted to read

api/scpca_portal/management/commands/sync_original_files.py
```diff
@@ -52,16 +52,24 @@ def sync_original_files(self, bucket: str, allow_bucket_wipe: bool, **kwargs) ->
         locked_project_ids = lockfile.get_locked_project_ids()
         Project.lock_projects(locked_project_ids)
 
-        bucket_objects = s3.list_bucket_objects(bucket, excluded_key_substrings=locked_project_ids)
+        bucket_objects = s3.list_bucket_objects(bucket)
```
davidsmejia, line 57: Instead of going over the wire twice and then looping twice - we should just go over the wire once for all objects, filter out locked project ids, then filter for syncable bucket objects.

davidsmejia: ultimately I would probably rewrite it to do the following:
1. Fetch from s3
2. Get locked and syncable from s3 resp
- 1. Convert to OF Objects
- 2. Loop 1 for locked projects
- 3. Loop 2 to get syncabled objects
3. Lock projects by ids
4. Loop 3 to update
5. Loop 4 to create
6. delete

davidsmejia (approved):
This looks good as is, I think we would want to consolidate the lockfile checking by moving get_from_dict up front and passing those objects around instead. If we do that now, re-request review otherwise we can file an issue to keep moving forward and ultimately apply that change since it is a self-contained refactor.

api/scpca_portal/models/original_file.py
```diff
@@ -114,73 +116,109 @@ def _is_downloadable(s3_key_info: utils.InputBucketS3KeyInfo) -> bool:
...
+            project_key = original_file.project_id if original_file.project_id else "portal-wide"
+            original_files_by_project_id[project_key].append(original_file)
+
+            if original_file.is_lockfile:
+                lockfiles.append(original_file)
+
+        for lockfile in lockfiles:
+            del original_files_by_project_id[lockfile.project_id]
+
+        syncable_original_files = utils.flatten_contents(
```
davidsmejia, line 143: Let's define this at the top and then extend here to remove the magicless magic string.

api/scpca_portal/models/original_file.py
```diff
@@ -114,73 +116,109 @@ def _is_downloadable(s3_key_info: utils.InputBucketS3KeyInfo) -> bool:
...
+            [original_files_by_project_id.values(), lockfiles]
+        )
+
+        return syncable_original_files, lockfiles
+
+    @classmethod
+    def bulk_create(cls, syncable_original_files: List[Self]) -> List[Self]:
```
davidsmejia, line 150: I would add doc strings to most of these. I would also drop syncable_ prefix because the name implies that you want to create them. We can add a guard if we want to avoid that but testing is probably fine

api/scpca_portal/test/management/commands/test_sync_original_files.py
```diff
@@ -136,3 +138,47 @@ def test_sync_original_files(self):
...
+        }
+        locked_project_lockfile = {
+            "LastModified": "2024-04-19T18:44:06.000Z",
+            "StorageClass": "STANDARD",
+            "s3_key": f"{locked_project_id}.lock",
+            "size_in_bytes": 0,
+            "hash": "d41d8cd98f00b204e9800998ecf8427e",
+        }
+
+        # TEST LOCKING - locked project's files are excluded, but its lockfile still syncs
```
davidsmejia, line 161: Lets add a comment to clean up the mock after updating the command to not lead with project locking

api/scpca_portal/models/original_file.py
```diff
@@ -114,73 +116,109 @@ def _is_downloadable(s3_key_info: utils.InputBucketS3KeyInfo) -> bool:
...
         )
-        deletable_file_list = list(deletable_files)
 
-        all_bucket_files = OriginalFile.objects.filter(s3_bucket=bucket)
         # if allow_bucket_wipe flag is not passed, do not allow all bucket files to be wiped
-        if set(all_bucket_files) == set(deletable_files) and not allow_bucket_wipe:
-            return []
+        if not allow_bucket_wipe:
+            if all_bucket_files.exists() and deletable_files.count() == all_bucket_files.count():
+                logger.info(
```
davidsmejia, line 215: we can make this warn

davidsmejia (approved):

nozomione (approved):
Great addition of helpers! LGTM 👍

PR #2062, opened 2026-08-04
Update Failed Job Handling

### LINKED ISSUE #2035: Update Failed Job Handling
### Context

Currently, failed job handling is implemented using hardcoded exception values and the `Job.attempt` field, which stores the count of attempts (the maximum retry attempts (up to 3)  are configured via Terraform). 

We should refactor failed job handling to use configurable retry and notification rules based on the job failure type rather than on the `attempt` count.

This allows predictable and more granular control over  handling of recoverable (retry) vs. unrecoverable (notification) failure.

e.g. ) Certain job failure reasons (e.g.,  `Dataset contains missing libraries`) consistently result in the unrecoverable `FAILED` job. In contrast, `Dataset contains locked project` may be retried later.

### Problem or idea

We should add a new `Job` instance method (e.g., `get_is_retryable`) that returns whether a failed job is retryable based on the failure type. The return value can then be used to decide whether to retry the job or send a notification.

| Failure Type | Next Step |
| --- | --- |
| `Dataset contains locked project` | Retry |
| `Failed to upload the computed file to S3` | Retry |
| `Failed to tag the computed file on S3` | Slack Notification for manual tagging |
| `Dataset contains missing libraries` | Notification |
| `Encountered uncaught exception` | Notification |

### Solution or next step

- Add the `Job` instance method, `get_is_retryable`, which returns a boolean based on the failure reason
- Update `DatasetJobProcessor`'s exception handlers to use `get_is_retryable` to decide how to handle  failed jobs
- Remove the `Job.attempt` field and `Job.increment_attempt_or_fail`
- Add a docstring describing the notification rules
- Apply migration to reflect the schema change
- Add tests to ensure the implementation works as expected

nozomione:
## Issue Number

Closes #2035
Stacked from PR #2053

Feature branch: `feature/job-dataset-state-reconciliation`

I've refactored the failed job handling during dataset job processing. 

Changes include:
- Grouped dataset job processing exceptions into the following categories (the group names indicate the next step for failed jobs):
   - `retryable_exceptions`
   - `slack_notification_exceptions`
   - `email_notification_exceptions`
- Added a `handle_failure` method, which handles failed job state updates and triggers the next step based on the exception types. All exception handlers call this method.
- Made the `notifications.send_slack_notification` function (previously `send_computed_file_tagging_error_email` generic

## Types of changes

## Functional tests

Added the new tests in `test/job_processors/test_dataset_job_processor.py`

## Checklist

## Screenshots

N/A

### DIFF
--- a/api/scpca_portal/job_processors/dataset_job_processor.py
+++ b/api/scpca_portal/job_processors/dataset_job_processor.py
@@ -1,6 +1,6 @@
 from scpca_portal import notifications, s3, utils
 from scpca_portal.config.logging import get_and_configure_logger
-from scpca_portal.enums import JobStates
+from scpca_portal.enums import FailedJobActions, JobStates
 from scpca_portal.exceptions import (
     DatasetLockedProjectError,
     DatasetMissingLibrariesError,
@@ -25,6 +25,7 @@ class DatasetJobProcessor(JobProcessorABC):
         "send_notification",
     ]
 
+    # Handlers to call when steps fail
     exception_handlers = {
         ("create_new_computed_file", DatasetLockedProjectError): "handle_locked_project",
         ("create_new_computed_file", DatasetMissingLibrariesError): "handle_missing_libraries",
@@ -32,11 +33,13 @@ class DatasetJobProcessor(JobProcessorABC):
         ("tag_new_computed_file", S3TaggingError): "handle_tag_failure",
     }
 
-    # Exception types
-    # Determine the next step for failed job handling
-    retryable_exceptions = [DatasetLockedProjectError, S3UploadError]
-    slack_notification_exceptions = [S3TaggingError]
-    email_notification_exceptions = [DatasetMissingLibrariesError]
+    # Actions to take after handling each exception
+    exception_actions = {
+        ("create_new_computed_file", DatasetLockedProjectError): FailedJobActions.RETRY,
+        ("create_new_computed_file", DatasetMissingLibrariesError): FailedJobActions.EMAIL,
+        ("upload_new_computed_file", S3UploadError): FailedJobActions.RETRY,
+        ("tag_new_computed_file", S3TaggingError): FailedJobActions.SLACK,
+    }
 
     # Logging
     def on_run(self) -> None:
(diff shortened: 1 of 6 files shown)


nozomione:
@davidsmejia, regarding the removal of `Job.attempt` and `Job.increment_attempt_or_fail`, I think we should perhaps keep them. 

My initial idea was to remove the field and simply retry job submission in a loop up to the max allowed attempts (`common.MAX_JOB_ATTEMPTS`). but that would immediately resubmit the Batch job within the same `submit_pending` execution block. The current approach persists the submission `attempt` count in DB, which allows the cron job to wait until the next `submit_pending` run before retrying. Since Batch job submission failures are most likely AWS Batch/API issues, waiting between retries seems more appropriate than retrying immediately. 

To make the intent clearer, we could rename: 
- `attempt` -> `submission_attempts`
- `increment_attempt_or_fail` -> `increment_submission_attempt_or_fail`
- `MAX_JOB_ATTEMPTS` -> `MAX_JOB_SUBMISSION_ATTEMPTS`

Renaming them explicitly associate them with Batch job **submission** attempts and distinguish them from Batch job **processing** retries.

I've made the changes locally, but before pushing the changes, I'd like to get your thoughts (or any alternative suggestions) as this would require DB schema changes. Thank you, David!

davidsmejia:
> To make the intent clearer, we could rename:
> 
>     * `attempt` -> `submission_attempts`
> 
>     * `increment_attempt_or_fail` -> `increment_submission_attempt_or_fail`
> 
>     * `MAX_JOB_ATTEMPTS` -> `MAX_JOB_SUBMISSION_ATTEMPTS`

I like this idea but it would probably make more sense to just be clearer in the management command submit_pending where we state that failed jobs actually reached the maximum number of failures and that pending jobs had errored during this submit and will be attempted in the future.

Otherwise I think its probably safe as is to keep, maybe we can add this to the docstrings or something to make sure it is clear that these failures / counts are pre-flight meaning we were unable to have the job represented on AWS infra.

nozomione:
🗒️ ~~Additional changes includes:~~
- ~~Added a new enum `FailedJobActions` and use it in `dataset_job_processor.handle_failure`~~
- ~~Replace the individual lists per exception types with a single dict matching `exception_handlers` data structure~~

**UPDATE:** @davidsmejia Utilizing the existing exception messages, each individual exception handler performs the corresponding action for failed jobs (e.g., job retry, email notification). 

This PR is ready for your review. Thank you, David!

api/scpca_portal/job_processors/dataset_job_processor.py
```diff
@@ -69,28 +70,26 @@ def create_new_computed_file(self) -> None:
...
         self.job.create_retry_job()
 
     def handle_tag_failure(self, step: str, e: Exception) -> None:
-        # Notify about S3 tagging failure via Slack to enable manual tagging
-        self.job.apply_state(JobStates.FAILED, reason="Failed to tag the computed file on S3.")
-        self.job.save()
-        logger.info("Send Slack notification for manual tagging alert.")
-        notifications.send_computed_file_tagging_error_email(self.job)
+        self.job.apply_state(JobStates.FAILED, reason=f"{e}")
+        logger.info("Send Slack notification for manual tagging.")
```
davidsmejia:
```suggestion
        logger.info("Sending Slack notification for manual tagging.")
```

davidsmejia (approved):
LGTM

avrohomgottlieb (approved):
LGTM

